PR #6, opened 2023-01-25
Implement CATKE in the 1/4 degree simulation

glwagner:
This PR adds a script to 1) inpaint land regions and 2) regrid the ECCO 1 degree initial condition to any grid (here, refining in horizontal to 1/4 degree and in the vertical to 87 levels from 48).

It will also remake the datadeps for the initial conditions using the new inpainted 1 degree solution and corresponding 1/4 degree solution (in progress).

Finally, we will run the 1/4 degree, 87 level simulation with CATKE as a turbulence closure.


src/DataWrangling.jl
```diff
@@ -0,0 +1,57 @@
...
+    arch = architecture(field)
+    grid = field.grid
+    loc = instantiated_location(field)
+    Nx, Ny, Nz = size(grid)
+
+    # Could transfer to CPU, then transfer back.
+    arch isa GPU && error("Inpainting on the GPU is not supported yet!")
+
+    for k = 1:Nz
+        mask = [peripheral_node(i, j, k, grid, loc...) for i = 1:Nx, j = 1:Ny]
```
navidcy, line 49:
```suggestion
        mask = [peripheral_node(i, j, k, grid, loc...) for j = 1:Ny, i = 1:Nx]
```

glwagner: huh, it seemed to be working

glwagner: isn't `i` the first index, `j` the second?

navidcy: yes but performance-wise is better to loop over columns first, right?

navidcy: https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-column-major

glwagner: That's up to the generator, not me

navidcy: there is a good chance I got confused here

glwagner: I'm also wasting a lot of memory, it would be more efficient to write

```julia
mask = zeros(Nx, Ny)

for k = 1:Nz
    for i = 1:Nx
        for j = 1:Ny
            mask[i, j] = peripheral_node(i, j, k, grid, loc...)
        end
    end
end
```

Then we have column major ordering for the loop as you suggest and no allocation.

However, note that the generator:

```julia
mask = [peripheral_node(i, j, k, grid, loc...) for j = 1:Ny, i = 1:Nx]
```

may create the array in whatever order is most efficient.

glwagner: I think though that the cost here is completely dominated by the inpainting...

glwagner: I will check to be sure.

navidcy: okie!
either way this step only happens once so it's not at all an issue!

glwagner:
Relevant discussion: https://forum.access-hive.org.au/t/initializing-a-1-4-degree-simulation-given-a-1-degree-temperature-salinity-data/403

glwagner:
This PR has unfortunately gone a little too far. I think we should probably close it, and cherry pick some of its contributions in separate PRs like #18 .

glwagner:
Not really sure the status of this. I think a lot of this PR is already on `main` so I'll close.

PR #7, opened 2023-01-26
Initial conditions and Fluxes

simone-silvestri:
Tools to produce a realistic initial condition and fluxes

(To improve)


glwagner:
Maybe start over without all the binary files?

glwagner:
Just a reminder that this branch needs to be deleted (because it carries a lot of heavy binary files)

simone-silvestri:
Binaries are removed

glwagner:
I believe they are still in history, so you need to delete this branch and open a new PR

navidcy:
Were these .bin ever lived on main? If so we need to perform deeper cleaning. :)

simone-silvestri:
nono just on this branch

PR #9, opened 2023-02-08
Initial conditions and fluxes

simone-silvestri:
A starting point for tools to produce a realistic initial condition and fluxes

(To improve)


glwagner:
We'll have to integrate this stuff with 

https://github.com/CliMA/ClimaOcean.jl/blob/glw/global-catke/src/DataWrangling.jl

glwagner:
Relevant discussion: https://forum.access-hive.org.au/t/initializing-a-1-4-degree-simulation-given-a-1-degree-temperature-salinity-data/403

src/DataWrangling.jl/interpolate_bathymetry.jl
```diff
@@ -24,27 +24,40 @@ SpectralInterpolation(; filter_func = (l) -> exp(-l * (l+1)/ 180 / 240), spectra
 
 
 """
-    interpolate_bathymetry_from_file(resolution, latitude; 
+    interpolate_bathymetry_from_file(resolution, maximum_latitude;
```
glwagner, line 27: Should we use `regrid!` to do this conservatively?

src/DataWrangling.jl/interpolate_bathymetry.jl
```diff
@@ -24,27 +24,40 @@ SpectralInterpolation(; filter_func = (l) -> exp(-l * (l+1)/ 180 / 240), spectra
...
     bathy_old = Float64.(file["bathymetry"])
 
-    Nx  = Int(360 / resolution)
-    Ny  = Int(2latitude / resolution)
+    Nx = Int(360 / resolution)
+    Ny = Int(2maximum_latitude / resolution)
 
-    Nn = (Nx, Ny)  
-    bathy = twodimensional_interpolation(bathy_old, interpolation_method, Nn)
+    bathy = twodimensional_interpolation(bathy_old, interpolation_method, (Nx, Ny))
```
glwagner, line 60:
```suggestion
    bathy = two_dimensional_interpolation(bathy_old, interpolation_method, (Nx, Ny))
```

src/DataWrangling.jl/interpolate_bathymetry.jl
simone-silvestri: we can, but it might not be so important for bathymetry as for IC and BC

glwagner: I think we do typically want to set the bathymetric height to be the "average" height in some region for some dynamics (large scale Rossby waves, tides?) But perhaps even more it might be nice to use fewer functions though to keep the code easier to understand (if the two approaches are equivalent)

glwagner:
Can we have a new PR that just has the stuff we need to build bathymetry, and also put a script in `examples/` that uses the tool, along with downloading ETOPO1?

simone-silvestri:
I'll close this PR as it is superseded by others

PR #17, opened 2023-03-17
ClimaNeverWorld

### LINKED ISSUE #20: 20 year simulation of the "CATKE Neverworld" at 1/4 degree
This issue will track progress with a stable 20 year simulation of the Neverworld2 configuration with CATKE.

A first step is merging #17 as well as https://github.com/CliMA/Oceananigans.jl/pull/3016. Then we are ready to perform a run and close this issue.

glwagner:
This PR implements the [Neverworld](https://gmd.copernicus.org/articles/15/6567/2022/) in ClimaOcean.

with @simone-silvestri 

Closes #20


navidcy:
YEAH!

I was gonna do it tomorrow but GREAT

glwagner:
Ok, also with @navidcy !

examples/run_neverworld_simulation.jl
```diff
@@ -0,0 +1,8 @@
+using Oceananigans
+using ClimaOcean.LimitedAreaSimulations: neverworld_simulation
```
navidcy, line 2: LimitedArea?
Don't like it but can't think of anything much better... Regional is used to mean something else...

glwagner: maybe a better distinction is between "realistic" and "idealized"?

navidcy: Yes!

src/LimitedAreaSimulations/neverworld_simulation.jl
```diff
@@ -29,17 +29,73 @@ using Oceananigans.Operators: Δzᶜᶜᶜ
...
+    y′ = y₁ + ϵ * Δy
+    
+    return sqrt((x′ - x)^2 + (y′ - y)^2)
+end
+
+function distance(point::Point, line::Line)
+    x, y = point.x, point.y
+    x₁, y₁ = line.p₁.x, line.p₁.y
+    x₂, y₂ = line.p₂.x, line.p₂.y
+    num = abs((x₂ - x₁) * (y₁ - y₀) - (y₂ - y₁) * (x₁ - x₀))
```
navidcy: what's `x₀` and `y₀`?

src/LimitedAreaSimulations/neverworld_simulation.jl
```diff
@@ -29,17 +29,73 @@ using Oceananigans.Operators: Δzᶜᶜᶜ
...
+
+distance(p₁::Point, p₂::Point) = sqrt((p₁.x - p₂.x)^2 + (p₁.y - p₂.y)^2)
+
+# https://stackoverflow.com/questions/849211/shortest-distance-between-a-point-and-a-line-segment
+function distance(point::Point, line::LineSegment)
+    x, y = point.x, point.y
+    x₁, y₁ = line.p₁.x, line.p₁.y
+    x₂, y₂ = line.p₂.x, line.p₂.y
+
+    # Line segment lengths
```
navidcy: these are not any lengths of anything... it's just the differences of the line-segment's coordinates, right?

src/LimitedAreaSimulations/neverworld_simulation.jl
glwagner: typo that should be `x, y`

glwagner: but maybe should delete this code cause we dont need it

navidcy: doesn't the bathymetry constructor use it?

src/LimitedAreaSimulations/neverworld_simulation.jl
glwagner: They are components of a vector that points along the segment?

src/LimitedAreaSimulations/neverworld_simulation.jl
glwagner: no

glwagner: this is for `Line`, the constructor uses `LineSegment`

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,298 @@
...
+    
+    return sqrt((x′ - x)^2 + (y′ - y)^2)
+end
+
+function distance(point::Point, line::Line)
+    x, y = point.x, point.y
+    x₁, y₁ = line.p₁.x, line.p₁.y
+    x₂, y₂ = line.p₂.x, line.p₂.y
+    num = abs((x₂ - x₁) * (y₁ - y) - (y₂ - y₁) * (x₁ - x))
+    den = sqrt((x₂ - x₁)^2 + (y₂ - y₁)^2)
```
navidcy, line 88: this is correct, but the way it's written it's very cryptic for me :)
I'd prefer to see it in terms of dot products of vectors :)

```Julia
using LinearAlgebra

 𝐱 = [x, y, 0]
 𝐱₁ = [x₁, y₁, 0]
 𝐱₂ = [x₂, y₂, 0]

return abs(cross(𝐱₂ - 𝐱₁,  𝐱₁ -  𝐱)) / norm(𝐱₂ - 𝐱₁)
```

glwagner: why the third component?

navidcy: because `cross` doesn't work for 2D vectors :)

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,298 @@
...
+function distance(point::Point, line::LineSegment)
+    x, y = point.x, point.y
+    x₁, y₁ = line.p₁.x, line.p₁.y
+    x₂, y₂ = line.p₂.x, line.p₂.y
+
+    # Line segment vector components
+    ℓˣ = x₂ - x₁
+    ℓʸ = y₂ - y₁
+
+    # Fractional increment to closest segment point
```
navidcy, line 72: This comment does not mean anything for me. After spending a few minutes I see what's happening here. It's clever. The clamp is crucial.

Here's a more geometric-inspired code

```Julia
𝐱  = [x,  y]
𝐱₁ = [x₁, y₁]
𝐱₂ = [x₂, y₂]

𝐥 = 𝐱₂ - 𝐱₁ # vector of the line segment

# we compute the ratio of the projection of 𝐱 on the direction 𝐥 over the length of 
ϵ = dot(𝐱, 𝐥) / norm(𝐥)^2

# if 0<ϵ<1 -> the closest point to 𝐱 is within the line segment
# if ϵ<0 -> the closest point to 𝐱 is 𝐱₁
# if ϵ>1 -> the closest point to 𝐱 is 𝐱₂

ϵ = clamp(ϵ, 0, 1)

𝐱′ = 𝐱₁ + ϵ * 𝐥 # this is = 𝐱₁ for ϵ=0 and =𝐱₂ for ϵ=1

return norm(𝐱, 𝐱′)
```

navidcy: or perhaps this is cleaner

```Julia
using LinearAlgebra

𝐱  = [x,  y,  0]
𝐱₁ = [x₁, y₁, 0]
𝐱₂ = [x₂, y₂, 0]

𝐥 = 𝐱₂ - 𝐱₁ # vector of the line segment

distance = dot(𝐱, 𝐥) < 0 ? norm(𝐱₁ - 𝐱) :
           dot(𝐱, 𝐥) > dot(𝐥, 𝐥) ? norm(𝐱₂ - 𝐱) :
           abs(cross(𝐱 - 𝐱₁, 𝐥)) / norm(𝐥)

return distance
```

navidcy: I didn't test it, but this is what it should be based on geometry, right?

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,298 @@
...
+
+    return NeverworldBathymetry(grid,
+                                coastline_spline,
+                                Float64(rim_width),
+                                Float64(slope_width),
+                                Float64(shelf_width),
+                                Float64(shelf_depth),
+                                Float64(abyssal_depth),
+                                Float64(southern_channel_boundary),
+                                Float64(northern_channel_boundary))
```
navidcy, line 146: Should these be same type as `eltype(grid)`? Perhaps if we have:

```julia
struct NeverworldBathymetry{FT, G, B} <: Function
    grid :: G
    coastline_spline :: B
    rim_width :: FT
    slope_width :: FT
    shelf_width :: FT
    shelf_depth :: FT
    abyssal_depth :: FT
    southern_channel_boundary :: FT
    northern_channel_boundary :: FT

    function NeverworldBathymetry(grid, coastline_spline, rim_width,
                                  slope_width, shelf_width, shelf_depth, abyssal_depth,
                                  southern_channel_boundary, northern_channel_boundary)

        FT = eltype(grid)
        B = typeof(coastline_spline)
        G = typeof(grid)

        return new{FT, G, B}(grid, coastline_spline, rim_width,
                            slope_width, shelf_width, shelf_depth, abyssal_depth,
                            southern_channel_boundary, northern_channel_boundary)
    end
end
```

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,298 @@
...
+    ℓˣ = x₂ - x₁
+    ℓʸ = y₂ - y₁
+
+    # Fractional increment to closest segment point
+    ϵ = ((x - x₁) * ℓˣ + (y - y₁) * ℓʸ) / (ℓˣ^2 + ℓʸ^2)
+    ϵ = clamp(ϵ, 0, 1)
+
+    # Closest segment point
+    x′ = x₁ + ϵ * Δx
+    y′ = y₁ + ϵ * Δy
```
navidcy, line 78:
```suggestion
    x′ = x₁ + ϵ * ℓˣ
    y′ = y₁ + ϵ * ℓʸ
```

src/IdealizedSimulations/neverworld_simulation.jl
glwagner: > Basically ϵ is a quantity that determines whether one of the edge points are the closest ones to 𝐱.

That's what I was trying to convey... the closest point on the segment to the target point is parameterized by `ϵ`. It's the relative distance along the segment from x1 to x2 to the closest segment point. So if `ϵ` the relative distance is 0 (ie, the closest point is x1). If `ϵ` is 1 then the relative distance is the full length of the segment is x2. Does that make sense? That's what the comment says. I don't find the linear algebra easier to understand ...

glwagner: "Fractional increment to closest segment point": the location of the closest segment point, parameterized as the relative distance along the segment pointing from x1 to x2.

dot norm and cross are opaque to me :-(

glwagner: I was trying to avoid if statements because it confuses me... I thought this process was clear:

1. Find the point on the segment that's closest to the target point
2. Compute the distance from the closest segment point to the target point

src/IdealizedSimulations/neverworld_simulation.jl
glwagner: The bathymetry is recomputed to be `eltype` grid so it doesn't matter

glwagner: unless i suppose we do higher precision than `Float64`

src/IdealizedSimulations/neverworld_simulation.jl
glwagner: `distance(point, line)` uses the formula from wikipedia:

https://en.wikipedia.org/wiki/Distance_from_a_point_to_a_line#:~:text=In%20Euclidean%20geometry%2C%20the%20distance,nearest%20point%20on%20the%20line.

I do think your code for `distance(point, line_segment)` is reasonable. But I'd vote for retaining the original code for point to line (I don't think invoking the cross product formula is better).

glwagner: I think the original function is more performant though. It doesn't matter, but if we did need to compute on the fly then we would need that original function. So I guess if we want to change to a new version we should at least keep the original as a comment or something.

navidcy: Cool. Keep the code. No worries. :)

But just to be fair you say “you don’t like if statement but you use clamp which is a very sneaky if statement.

src/IdealizedSimulations/neverworld_simulation.jl
navidcy: Ok!

src/IdealizedSimulations/neverworld_simulation.jl
navidcy: > `distance(point, line)` uses the formula from wikipedia:
> 
> https://en.wikipedia.org/wiki/Distance_from_a_point_to_a_line#:~:text=In%20Euclidean%20geometry%2C%20the%20distance,nearest%20point%20on%20the%20line.
> 
> I do think your code for `distance(point, line_segment)` is reasonable. But I'd vote for retaining the original code for point to line (I don't think invoking the cross product formula is better).

This wikipedia article needs work. I’ll fix it. :)

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,298 @@
...
+
+    # Fractional increment to closest segment point
+    ϵ = ((x - x₁) * ℓˣ + (y - y₁) * ℓʸ) / (ℓˣ^2 + ℓʸ^2)
+    ϵ = clamp(ϵ, 0, 1)
+
+    # Closest segment point
+    x′ = x₁ + ϵ * ℓˣ
+    y′ = y₁ + ϵ * ℓʸ
+    
+    return sqrt((x′ - x)^2 + (y′ - y)^2)
```
navidcy, line 80: we can use the method `distance(p₁::Point, p₂::Point)` here? or delete that method?

navidcy, line 80: we can use the method `distance(p₁::Point, p₂::Point)` here? or delete that method?

```suggestion
    return distance(Point(x, y), Point(x′, y′))
```

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,298 @@
...
+end
+
+function distance(point::Point, line::Line)
```
navidcy, line 83:
```suggestion
"""
    distance(point::Point, line::Line)

Return the Euclidean distance of a `point` to a `line`, i.e., the shortest distance from the `point` to a point on the `line`.
"""
function distance(point::Point, line::Line)
```

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,298 @@
...
+    p₂ :: Point{T}
+end
+
+distance(p₁::Point, p₂::Point) = sqrt((p₁.x - p₂.x)^2 + (p₁.y - p₂.y)^2)
+
+# https://stackoverflow.com/questions/849211/shortest-distance-between-a-point-and-a-line-segment
+function distance(point::Point, line::LineSegment)
+    x, y = point.x, point.y
+    x₁, y₁ = line.p₁.x, line.p₁.y
+    x₂, y₂ = line.p₂.x, line.p₂.y
```
navidcy, line 66:
```suggestion
```suggestion
"""
   distance(point::Point, linesegment::LineSegment)

Return the distance between a `point` and a `linesegment`, that is the shortest distance of the `point` with
and of the points within the line segment.
"""
function distance(point::Point, linesegment::LineSegment)
    x, y = point.x, point.y
    x₁, y₁ = linesegment.p₁.x, linesegment.p₁.y
    x₂, y₂ = linesegment.p₂.x, linesegment.p₂.y
```

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,304 @@
...
+end
+
+struct Line{T}
+    p₁ :: Point{T}
+    p₂ :: Point{T}
+end
+
+distance(p₁::Point, p₂::Point) = sqrt((p₁.x - p₂.x)^2 + (p₁.y - p₂.y)^2)
+
+```suggestion
```
navidcy, line 62:
```suggestion
```

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,303 @@
...
+    p₂ :: Point{T}
+end
+
+distance(p₁::Point, p₂::Point) = sqrt((p₁.x - p₂.x)^2 + (p₁.y - p₂.y)^2)
+
+"""
+   distance(point::Point, linesegment::LineSegment)
+
+Return the distance between a `point` and a `linesegment`, that is the shortest distance
+ of the `point` with and of the points within the line segment.
```
navidcy, line 66:
```suggestion
 of the `point` to any of the points within the line segment.
```

glwagner:
I think this is ready for review @navidcy @simone-silvestri

glwagner:
image

glwagner:
Closes #20

examples/analyze_neverworld_simulation.jl
```diff
@@ -0,0 +1,44 @@
+using Oceananigans
+using Oceananigans.ImmersedBoundaries: mask_immersed_field!
+using Oceananigans.Units
+#using GLMakie
```
glwagner, line 4:
```suggestion
```

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,390 @@
...
+
+    if momentum_advection isa Default
+        momentum_advection = VectorInvariant(vorticity_scheme   = WENO(grid.underlying_grid),
+                                             divergence_scheme  = WENO(grid.underlying_grid),
+                                             vertical_scheme    = WENO(grid.underlying_grid))
+    end
+
+    if tracer_advection isa Default
+        # Turn off advection of tke for efficiency
+        tracer_advection = (b=WENO(grid.underlying_grid), e=nothing)
```
glwagner, line 337:
```suggestion
        tracer_advection = Dict()
        tracer_advection = Dict{Symbol, Any}(name => WENO(grid.underlying_grid) for name in tracers)
        tracer_advection[:e] = nothing
        tracer_advection = NamedTuple(name => tracer_advection[name] for name in tracers)
```

glwagner:
Can you look this over @navidcy or @simone-silvestri ?

glwagner:
Animation from the last 4 years of a 200 year "CATKE Neverworld" run with a seasonal cycle.

https://user-images.githubusercontent.com/15271942/230837853-0472c263-f533-424c-9551-d3a8886adab9.mp4

navidcy:
I’ll have a look tomorrow!

examples/gm_one_degree_calibration/distributed_gm_one_degree_model_calibration.jl
```diff
@@ -1,175 +0,0 @@
...
-using Oceananigans.Units
-using Oceananigans.Grids: on_architecture
-using Oceananigans.Utils: WallTimeInterval
-using Oceananigans.BuoyancyModels: buoyancy
-using Oceananigans.Models.HydrostaticFreeSurfaceModels: VerticalVorticityField
-using ClimaOcean.NearGlobalSimulations: one_degree_near_global_simulation
-using ParameterEstimocean
-using ParameterEstimocean.Utils: map_gpus_to_ranks!
-using ParameterEstimocean.Observations: FieldTimeSeriesCollector
-using ParameterEstimocean.Parameters: closure_with_parameters
```
simone-silvestri, line 12: we give up on the 1 degree calibration?

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,393 @@
...
+                               buoyancy_relaxation_time_scale = 14days,
+                               target_buoyancy_distribution = seasonal_cosine_target_buoyancy_distribution,
+                               bottom_drag_coefficient = 2e-3,
+                               equator_pole_buoyancy_difference = 0.06,
+                               seasonal_cycle_relative_amplitude = 0.8,
+                               surface_buoyancy_gradient = 1e-4,
+                               stratification_scale_height = 500, # meters
+                               time_step = 5minutes,
+                               stop_time = 30days,
+                               bathymetry = nothing,
```
simone-silvestri, line 314: this kwarg is probably unused

examples/run_neverworld_simulation.jl
```diff
@@ -0,0 +1,97 @@
...
+using ClimaOcean.IdealizedSimulations: neverworld_simulation
+using Printf
+
+minimum_turbulent_kinetic_energy = 1e-6
+minimum_convective_buoyancy_flux = 1e-11
+closure = CATKEVerticalDiffusivity(; minimum_turbulent_kinetic_energy,
+                                   minimum_convective_buoyancy_flux)
+
+simulation = neverworld_simulation(GPU();
+                                   horizontal_size = (240, 280),
```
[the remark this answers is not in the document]
glwagner: We can do that. I used `size` because the number of grid points is straightforwardly related to performance (resolution has to be combined with the domain size to get performance). But I agree that resolution is probably the thing that you'd vary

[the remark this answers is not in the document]
glwagner: Done.

examples/gm_one_degree_calibration/distributed_gm_one_degree_model_calibration.jl
glwagner: I think so --- do you want to keep maintaining it? I don't think we have plans to return to it so it makes sense to declare now that we are not maintaining it (we can always access this code in git history if we want to). What do you think?

examples/analyze_neverworld_simulation.jl
```diff
@@ -0,0 +1,43 @@
+using Oceananigans
+using Oceananigans.ImmersedBoundaries: mask_immersed_field!
```
navidcy, line 2: is this used?

examples/animate_neverworld_simulation.jl
```diff
@@ -0,0 +1,152 @@
+using Oceananigans
+using Oceananigans.Units
+using GLMakie
+using Printf
+using Statistics
+
+# Some plotting parameters
+ζlim = 6e-5
```
navidcy, line 8:
```suggestion
ζlim = 6e-5     # s⁻¹
```

examples/run_neverworld_simulation.jl
```diff
@@ -0,0 +1,97 @@
+using Oceananigans
+using Oceananigans.Units
+using Oceananigans.TurbulenceClosures: CATKEVerticalDiffusivity
+using ClimaOcean.IdealizedSimulations: neverworld_simulation
+using Printf
+
+minimum_turbulent_kinetic_energy = 1e-6
+minimum_convective_buoyancy_flux = 1e-11
```
navidcy, line 8:
```suggestion
minimum_turbulent_kinetic_energy = 1e-6     # m² s⁻²
minimum_convective_buoyancy_flux = 1e-11    # UNITS?
```

src/ClimaOcean.jl
```diff
@@ -46,10 +48,62 @@ function __init__(; remove_existing_data=false)
...
+        push!(nodes, nodes[end])
+    
+        coordinates = Vector{FT}(coordinates)
+        nodes = Vector{FT}(nodes)
+
+        # Now we can build the spline
+        spline = CubicSpline(coordinates, nodes)
+        S = typeof(spline)
+
+        d == :x || d == :y || d == :z || error("Dimension 'd' must be :x or :y or :z")
```
navidcy, line 81: would that work with a LatitudeLongitude grid with λ/φ coordinates?

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,409 @@
...
+
+instantiate(T::DataType) = T()
+instantiate(t) = t
+
+function minimum_spacing(dir, (LX, LY, LZ), grid)
+    spacing = eval(Symbol(dir, :spacing))
+    loc = map(instantiate, (LX, LY, LZ))
+    Δ = KernelFunctionOperation{LX, LY, LZ}(spacing, grid, loc...)
+    return minimum(Δ)
+end
```
navidcy, line 54: is this different from Oceananigans? Why not use the one from Oceananigans?

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,409 @@
...
+                                Float64(shelf_width),
+                                Float64(shelf_depth),
+                                Float64(abyssal_depth),
+                                Float64(southern_channel_boundary),
+                                Float64(northern_channel_boundary),
+                                Float64(scotia_ridge_height),
+                                Float64(scotia_ridge_radius),
+                                Float64(scotia_ridge_width),
+                                Float64(scotia_ridge_center_longitude),
+                                Float64(scotia_ridge_center_latitude))
```
navidcy, line 198: why all these `Float64` here instead of `eltype(grid)`?

src/IdealizedSimulations/neverworld_simulation.jl
```diff
@@ -0,0 +1,409 @@
...
+#
+#   * Zonally-homogeneous wind stress with mid-latitude jet and trade winds
+#
+#   * Buoyancy 
+#       * restoring hot at the equator and cold at the poles (parabola, cosine, smooth step function)
+#       * equator-pole buoyancy differential: 0.06 (α * g * 30 ≈ 0.06 with α=2e-4, g=9.81)
+#       * exponential initial vertical stratification with N² = 6e-5 and decay scale h = 1000 m
+#           - eg bᵢ = N² * h * exp(z / h)
+#
+#   * Quadratic bottom drag with drag_coefficient = 1e-3
```
navidcy, line 21: should these be part of the docstring of `neverworld_simulation()`?

src/ClimaOcean.jl
glwagner: yeah, here "x" etc only refers to a positional argument. The only point of using "y" is if you have a two-argument function (such as what we load into `GridFittedBottom`) and need to call this with (x, y), where the spline only varies in the second argument.

We could also get rid of this concept since I don't think it's needed (yet)

examples/gm_one_degree_calibration/job_script.sh
navidcy, line 1: this might be useful but perhaps we should add it in the repo's wiki or something -- not in the main repo as it was!!

src/IdealizedSimulations/neverworld_simulation.jl
glwagner: these functions are not called on the fly. They are called just once when `ImmersedBoundaryGrid` is constructed. The bathymetry data is always eltype(grid). I put `Float64` for convenience, and also not to mislead people into thinking that it actually mattered what datatype was used. We could also leave the datatype unspecified, maybe at some cost to performance during setup.

src/IdealizedSimulations/neverworld_simulation.jl
glwagner: We can though I'm not a huge fan of massive docstrings, I think they are not user-friendly...

Better to put them in legit documentation which we don't have yet.

examples/gm_one_degree_calibration/job_script.sh
glwagner: Maybe better to put in Oceananigans wiki since that already has other useful stuff

src/IdealizedSimulations/neverworld_simulation.jl
glwagner: good point, I think this was written before we had it in Oceanagnians

examples/gm_one_degree_calibration/job_script.sh
navidcy: OK.

Docstrings could be either user friendly or not, depending on how they are written.

As a user of any repo I feel it's really user-friendly if I do `help?> a_method` and I get back a useful, meaningful, docstring with some examples that make sense.

src/IdealizedSimulations/neverworld_simulation.jl
navidcy: OK.

Docstrings could be either user friendly or not, depending on how they are written.

As a user of any repo I feel it's really user-friendly if I do `help?> a_method` and I get back a useful, meaningful, docstring with some examples that make sense.

examples/gm_one_degree_calibration/job_script.sh
navidcy: Agree!

examples/gm_one_degree_calibration/distributed_gm_one_degree_model_calibration.jl
simone-silvestri: agreed!

src/IdealizedSimulations/neverworld_simulation.jl
glwagner: Of course I'm in favor of doctstrings, but I don't think we should put anything longer than a page in a docstring...

glwagner: I think the baseline information most people need is: what are the function arguments. If concise examples can be provided then that is excellent. I'm less sure about docstrings that span multiple pages. For example we wouldn't load an entire oceananigans example into a docstring for `Simulation`. There's a limit beyond which the excess information is hurting not helping

PR #39, opened 2023-08-06
Adds biogeochemistry interface + fix example to work out of the box

jagoosw:
I've updated some of the dependencies and modified the one-degree example a bit so it will run as is now.

Also added biogeochemistry to near-global model-building functions.

@glwagner I've commented out the CATKE closure in the one-degree example because I can't get it to run as is, maybe better to remove it from the example?


navidcy:
@glwagner could you/we have a look here? I'd like to merge this in an attempt to deal with https://github.com/OceanBioME/GlobalOceanBioME/issues/1

glwagner (approved):
Not a problem. Note we will eventually deprecate `NearGlobalSimulation` in favor of the new `OceanSeaIceModel` infrastructure.

navidcy:
OK, I'll merge this when tests pass.

PR #41, opened 2023-08-24
Add some CATKE diagnostics

glwagner:
This PR adds some functions for building `KernelFunctionOperation` of CATKE diagnostics.

cc @simone-silvestri


glwagner:
This is stale but we can revisit sometime if we like

Issue #28, opened 2023-06-07
The road to OMIP


glwagner:
We'd like to make a (reproducible, well-documented) OMIP setup. Here's my proposed roadmap. I'm going to start with 1/4 degree resolution on `LatitudeLongitudeGrid`, though the ultimate goal will be a somewhat higher resolution on the cubed sphere. Feedback very welcome on the plan below. @navidcy @simone-silvestri @siddharthabishnu 

1. *Bathymetry generation.*
    * Util for downloading and regridding [ETOPO1](https://www.ncei.noaa.gov/access/metadata/landing-page/bin/iso?id=gov.noaa.ngdc.mgg.dem:316) ([NetCDF with ice surface elevations](https://www.ngdc.noaa.gov/mgg/global/relief/ETOPO1/data/ice_surface/cell_registered/netcdf/ETOPO1_Ice_c_gmt4.grd.gz)), which is 1/60th degree (note that in reality the actual resolution is a lot coarser over the majority of the globe). Regridding is probably expensive so I think we should host the regridded data on engaging. I'm not sure it matters whether we use the bedrock or ice surface elevations.
    * Functions that implement the bathymetry edits we need for high quality solutions at each resolution (starting with 1/4 degree for now). The edits are things like, removing inland seas, widening narrow regions, deepening problematically shallow areas.
    * Declare victory when we can initialize an ocean with constant stratification with realistic bathymetry.

2. *Initializing from WOA.*
    * Util for downloading, regridding, smoothing, NaN in painting. This is less expensive, so in principle could be done on the fly if we like
    * Declare victory when we can run a simple simulation from rest and with WOA temperature / salinity.

4. *Forcing by JRA55-do winds*
    * For now use a constant drag coefficient on the relative atmos-ocean velocity --- start with repeat year forcing rather than the whole 55 year time series?
    * Again host regridded JRA55-do data on engaging. Implement util for regridding.
    * Possibly, implement util for "chunked" loading in FieldTimeSeries. In other words, to force with a very long time series of high resolution 3-hourly winds, we cannot load the whole time-series at the same time. Nor do we want to load data every time step, though. So we need "chunked" loading, where we load some portion of the time-series, say, every 30 days. This kind of chunked loading would also be useful for `FieldTimeSeries`, so we can possible tackle it that way and gain a useful Oceananigans feature.
    * Declare victory when we can run a simulation for some amount of time. Perhaps for a year.

[...]

PR #49, opened 2023-10-05
`OceanSeaIceModel` for coupled ocean-sea-ice simulations with prescribed atmospheric state and radiative fluxes

glwagner:
This PR implements a new model type called `OceanSeaIceModel`, which provides a user interface for coupled ocean-sea-ice simulations with a prescribed atmospheric state and radiative fluxes.

cc @simone-silvestri


glwagner:
I also added an alias `OceanOnlyModel` for cases where `isnothing(IceOceanModel.ice)`. This will facilitate regional simulations with no ice component.

simone-silvestri:
Also, I'll start to implement `compute_air_sea_fluxes!` for `OceanOnly` then we need to enable 3-way interaction. 
(atmosphere should talk to ocean where there is no ice and ice there thickness > 0). 
So another thing to ensure is that the BC of ocean to atmosphere and ice to atmosphere are consistent

simone-silvestri:
I just realized that the feedback of the ice in the ocean is instantaneous, i.e. there is no flux that acts as a boundary condition for the ocean because the change is actuated instantaneously. 

It is kind of an "adjustment". Maybe we should change the name of the functions to reflect it, `compute_flux!` might be misleading.

glwagner:
> I just realized that the feedback of the ice in the ocean is instantaneous, i.e. there is no flux that acts as a boundary condition for the ocean because the change is actuated instantaneously.
> 
> It is kind of an "adjustment". Maybe we should change the name of the functions to reflect it, `compute_flux!` might be misleading.

That is true, but the function does compute a flux which is passed into the ice model. Also, the "adjustment" is not mathematically distinguishable from a flux. In other words from the ocean perspective, there is an "implied flux". From the sea ice perspective, the flux is explicit.

The situation is also a little more nuanced. That function implements a particular parameterization for the "ocean-driven melting" case (eg the "ice bath" model) that is most simply implemented as adjustment and thus implicit flux. However we want generalize it to other parameterizations which do involve explicit fluxes. So it will compute fluxes in the future...

Let's discuss, this is a huge effort and so the code will be in a state of change for some time. We do need to flesh out the scaffolding though.

glwagner:
> I think we need to ensure consistency between models and forcings:
> 
> 1. make sure there is a salinity and a temperature tracer in the ocean model (let's think in another PR about how to extend it for no-salinity simulations)
> 2. top boundary conditions for temperature/salinity are flux boundary conditions and that the forcing is defined on the same grid

Yes I think for the time being we should simply assume that the model has both temperature and salinity. Certainly, there is a place for idealized simulations with only one of the two, but I think it would be better to generalize later rather than build in the complexity now.

src/OceanSeaIceModels/ocean_only_model.jl
```diff
@@ -0,0 +1,48 @@
...
+compute_ice_ocean_salinity_flux!(::OceanOnlyModel) = nothing
+ice_ocean_latent_heat!(::OceanOnlyModel) = nothing
+
+#####
+##### Air-sea fluxes
+#####
+
+function time_step!(coupled_model::OceanOnlyModel, Δt; callbacks=nothing)
+    compute_air_sea_flux!(coupled_model)
+    time_step!(ocean)
```
simone-silvestri, line 18:
```suggestion
    time_step!(coupled_model.ocean)
    time_step!(coupled_model.atmosphere)
```

src/OceanSeaIceModels/ocean_only_model.jl
```diff
@@ -0,0 +1,48 @@
...
+    tick!(coupled_model.clock, Δt)
+    return nothing
+end
+
+function compute_air_sea_fluxes!(coupled_model::OceanOnlyModel) 
+    ocean   = coupled_model.ocean
+    forcing = coupled_model.atmospheric_forcing
```
simone-silvestri, line 25:
```suggestion
    forcing = coupled_model.atmosphere
```

experiments/prototype_omip_simulation/omip_ocean_component.jl
```diff
@@ -0,0 +1,37 @@
...
+
+tracer_advection = WENO()
+momentum_advection = VectorInvariant(vorticity_scheme = WENO(),
+                                     divergence_scheme = WENO(),
+                                     vertical_scheme = WENO())
+
+ocean_model = HydrostaticFreeSurfaceModel(; grid, buoyancy, closure,
+                                          tracer_advection, momentum_advection,
+                                          tracers = (:T, :S, :e),
+                                          free_surface = SplitExplicitFreeSurface(cfl=0.2; grid),
```
simone-silvestri, line 31:
```suggestion
                                          free_surface = SplitExplicitFreeSurface(cfl=0.75; grid),
```

src/DataWrangling/JRA55.jl
```diff
@@ -159,64 +204,266 @@ function jra55_field_time_series(variable_name;
...
+
+## Returns
+- `fts`: `FieldTimeSeries` object containing the retrieved or interpolated data.
+
+If `interpolated_file` is not a `Nothing`:
+(1) If the `interpolated_file` does not exist, the JRA55 data will be written to the file in an Oceananigans compatible format. 
+(2) If the `interpolated_file` exists but is on a different grid, the file will be deleted and rewritten.
+(3) If the `shortname` is already present in the file, the data will not be written again.
+"""
+function retrieve_and_maybe_write_jra55_data(::Nothing, grid, times, loc, boundary_conditions, data, jra55_native_grid; kwargs...)
```
glwagner, line 328: This function isn't specific to JRA55, I think it should be in another file

src/DataWrangling/JRA55.jl
```diff
@@ -159,64 +204,266 @@ function jra55_field_time_series(variable_name;
...
+
     return fts
 end
 
+"""
+    interpolate_and_write_timeseries!(data, loc, grid, times, path, name, bcs, native_grid)
+
+Interpolates and writes a time series of `data` at `times` onto disk in an Oceananigans compatible format.
+"""
+function interpolate_and_write_timeseries!(data, loc, grid, times, path, name, bcs, native_grid)
```
glwagner, line 395: This function is not specific to JRA55 so I don't think it belongs here in JRA55.jl

src/DataWrangling/JRA55.jl
```diff
@@ -159,64 +204,266 @@ function jra55_field_time_series(variable_name;
...
 
-    # Fill the data
-    interior(fts, :, :, 1, :) .= data[:, :, :]
+    boundary_conditions = FieldBoundaryConditions(jra55_native_grid, (Center, Center, Nothing))
+
+    grid = isnothing(grid) ? jra55_native_grid : grid
+    loc  = (LX, LY, Nothing)
+
+    fts = retrieve_and_maybe_write_jra55_data(time_chunks_in_memory, grid, times, loc, boundary_conditions, data, jra55_native_grid; 
+                                              interpolated_file, shortname)
```
glwagner, line 299: Let's design a user interface that generalizes how the user specifies `time_indices` (one of the arguments to `jra55_field_time_series` so that the user can say "load these indices to disk, leave the rest in a file" or something like that.

glwagner:
@simone-silvestri starting a conversation here about how to add features to `jra55_field_time_series`. First, the English version of what we want, I think:

* We want users to be able to load JRA55 data from the native `.nc` files directly to disk (for playing around, inspecting, plotting)
* We want users to be able to "partially" load long JRA55 time-series --- this requires pre-processing the `.nc` files to `.jld2` files, and then building a `FieldTimeSeries` from the `.jld2` files
* I think we want users to be able to directly load pre-processes `.jld2` files, anticipating that we only want to do the pre-processing once

We want the first feature because it's important, for reproducibility and transparency, to be able to replicate the workflow that produces the `.jld2` files. And also new JRA55 data will become available in the future. But I think it'll be most convenient if people mostly work with pre-processed JRA55 data.

On interpolation --- thinking about this more, I wonder if we should interpolate on the fly within the GPU kernels. It seems wasteful to interpolate in advance, I'm not 100% this saves much and has a huge cost upfront. Something to think about and ponder. It might be cool to also have a field that represents interpolation...

simone-silvestri:
It makes sense to interpolate on the fly in the flux calculation. 
I think if that is the goal then it makes sense to organize directly the JRA55 data onto a 1/4 Oceananigans grid in a .jld2 file without interpolation and save those online together with the nc dataset. 
The difference would be that we download the jld2 files directly skipping the extra step of writing down the jld2 files. 
At that point, it would be pretty easy to specify the time indices.
We can have a utility to grab the nc dataset for users that want to look at the reanlysis data without having to use julia.

glwagner:
I was thinking this would be an extra feature, we don't have to eliminate existing features

src/DataWrangling/ECCO2.jl
```diff
@@ -1,111 +1,173 @@
...
-ecco2_file_names = Dict(
-    :temperature             => "ecco2_temperature_19920102.nc",
-    :salinity                => "ecco2_salinity_19920102.nc",
-    :effective_ice_thickness => "ecco2_effective_ice_thickness_19920102.nc",
+urls_19921001 = Dict(
+    :temperature           => "https://www.dropbox.com/scl/fi/169f3981460uhk9h69k0f/THETA.1440x720x50.19921001.nc?rlkey=mgal3xt0qy2c59y395ybio11v",
+    :salinity              => "https://www.dropbox.com/scl/fi/f9zfm34vqz732jrrhjrg3/SALT.1440x720x50.19921001.nc?rlkey=y5dv0s41gb6f9guvu0iorw28p",
+    :sea_ice_thickness     => "https://www.dropbox.com/scl/fi/mtmziurepom8kpjn82d07/SIheff.1440x720.19921001.nc?rlkey=9uhuxg2n9iw6894afj4t53drv",
+    :sea_ice_area_fraction => "https://www.dropbox.com/scl/fi/ntflhyrmsnit9vco402co/SIarea.1440x720.19921001.nc?rlkey=eakzc788btql1q6ndj9l8cr2q",
+    #:u_velocity            => "https://www.dropbox.com/scl/fi/e6s9c013r2ddift4f8ugi/UVEL.1440x720x50.19921001.nc?rlkey=fpd7mv1zv3fkmyg8w11b94sbp&dl=0",
```
navidcy, line 107:
```suggestion
```

glwagner:
This PR now depends on https://github.com/CliMA/Oceananigans.jl/pull/3450

glwagner:
These two lines download data @navidcy:

https://github.com/CliMA/ClimaOcean.jl/blob/f2e76638dfea91a1d14f5b497e2ff5f21d89f4ca/experiments/prototype_omip_simulation/regional_omip_simulation.jl#L25-L26

and

https://github.com/CliMA/ClimaOcean.jl/blob/f2e76638dfea91a1d14f5b497e2ff5f21d89f4ca/experiments/prototype_omip_simulation/regional_omip_simulation.jl#L97

For example, you can try

```julia
atmosphere = JRA55_prescribed_atmosphere(backend=OnDisk())
```

Which (I think) will download and then preprocess the entire repeat year forcing dataset. See here:

https://github.com/CliMA/ClimaOcean.jl/blob/10ece457b3dbf2688e627890d0f128e2e1cf9231/src/DataWrangling/JRA55.jl#L470-L473

experiments/prototype_omip_simulation/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.0"
+julia_version = "1.10.0-rc1"
```
navidcy: @glwagner you should install Julia v1.10.0 :)

experiments/prototype_omip_simulation/Manifest.toml
```diff
@@ -1304,9 +1304,9 @@ version = "0.5.5"
 
 [[deps.Oceananigans]]
 deps = ["Adapt", "CUDA", "Crayons", "CubedSphere", "Dates", "Distances", "DocStringExtensions", "FFTW", "Glob", "IncompleteLU", "InteractiveUtils", "IterativeSolvers", "JLD2", "KernelAbstractions", "LinearAlgebra", "Logging", "MPI", "NCDatasets", "OffsetArrays", "OrderedCollections", "PencilArrays", "PencilFFTs", "Pkg", "Printf", "Random", "Rotations", "SeawaterPolynomials", "SparseArrays", "Statistics", "StructArrays"]
-git-tree-sha1 = "ead1d8ec6911ff84ef52c387c827553bbc859c1b"
+path = "/Users/gregorywagner/Projects/Oceananigans.jl"
```
navidcy: I need to change this to test on Gadi; is there a relevant branch in Oceananigans I should be pointing to?

experiments/prototype_omip_simulation/omip_components.jl
```diff
@@ -0,0 +1,94 @@
...
+                                    advection = nothing,
+                                    ice_consolidation_thickness = 0.05,
+                                    ice_salinity = 4,
+                                    internal_heat_flux = ConductiveFlux(conductivity=2),
+                                    top_heat_flux = ConstantField(0), # W m⁻²
+                                    top_heat_boundary_condition = PrescribedTemperature(0),
+                                    bottom_heat_boundary_condition = bottom_bc,
+                                    bottom_heat_flux = sea_ice_ocean_heat_flux)
+
+    s# et!(sea_ice_model, h=ℋᵢ)
```
navidcy, line 87: is this supposed to be 

```suggestion
    # set!(sea_ice_model, h=ℋᵢ) 
```

src/DataWrangling/ECCO2.jl
```diff
@@ -1,111 +1,173 @@
...
-    Δz = zc[2] - zc[1]
-    pushfirst!(zf, zf[1] - Δz)
 
-    return zf
-end
+surface_variable(variable_name) = variable_name == :sea_ice_thickness
 
-function ecco2_field(variable_name;
+function ecco2_field(variable_name, date=Date(1992, 01, 02);
                      architecture = CPU(),
```
navidcy, line 136: if `ecco2_field` is architecture-depended, then users of HPCs that don't have Internet access on CPU/GPU compute nodes how are they supposed to download it?

for example, on Australia's gadi only the login node has internet access and that's a CPU architecture...

experiments/prototype_omip_simulation/Manifest.toml
glwagner: `glw-ss/ice-ocean-model`

experiments/prototype_omip_simulation/omip_components.jl
glwagner: fixing now

src/DataWrangling/ECCO2.jl
glwagner: the downloading is not architecture dependent

glwagner: you can build a wrapper for this step:

https://github.com/CliMA/ClimaOcean.jl/blob/9125cb5e59334c3705879424952fb0582b98ed3b/src/DataWrangling/ECCO2.jl#L69

just need to design the API

experiments/prototype_omip_simulation/single_column_omip_simulation.jl
```diff
@@ -0,0 +1,406 @@
...
+
+    fluxes = (; τˣ, τʸ, E, Jˢ, Q, Qse, Qla)
+
+    auxiliary_fields = (; N², κᶜ)
+    fields = merge(ocean.model.velocities, ocean.model.tracers, auxiliary_fields)
+
+    # Slice fields at the surface
+    outputs = merge(fields, fluxes)
+
+    output_attributes = Dict{String, Any}(
```
navidcy, line 217: let's open an issue or PR in Oceananigans to make the required changes so that the `{String, Any}` is not required here?

glwagner: do it. NetCDFOutputWriter probably needs a lot of work. I never use it.

navidcy: I just don't know what's needed to be modified. Is it these lines:

https://github.com/CliMA/Oceananigans.jl/blob/3b02321a350003fa9596c4f40839b031a5cc7ad1/src/OutputWriters/netcdf_output_writer.jl#L335-L337

should become

```Julia
                            global_attributes = Dict(String, Any),
                            output_attributes = Dict(String, Any),
                                   dimensions = Dict(String, Any),
```

experiments/prototype_omip_simulation/omip_sea_ice_component.jl
navidcy, line 1: should this empty file be deleted?

src/DataWrangling/JRA55.jl
```diff
@@ -139,83 +225,339 @@ Keyword arguments
...
+
+    isnothing(shortname) && (shortname = jra55_short_names[variable_name])
+    isnothing(filename)  && (filename  = filenames[variable_name])
+    isnothing(url)       && (url       = urls[variable_name])
 
+    # Record some important user decisions
+    totally_in_memory = backend isa TotallyInMemory
+    on_native_grid = isnothing(grid)
+
+    jld2_filename = string("JRA55_repeat_year_", variable_name, ".jld2")
```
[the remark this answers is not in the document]
glwagner: that's right. 

More specifically `FieldTimeSeries` only works with a specific file format.

We could definitely design a JRA55-specific backend for `FieldTimeSeries` though. Might be a nice project and would save a lot setup effort.

experiments/prototype_omip_simulation/single_column_omip_simulation.jl
glwagner: No. Those are the defaults in the constructor. The problem is wherever in the code that we assume `output_attributes` has some type.

Probably the problem is that we need to add elements to `output_attributes`. If the eltype of the dictionary is `Any`, we can add any element. If the eltype is something else, the `NetCDFOutputWriter` may error when trying to add it.

Thus the step that is needed is to convert the user-input into a type that will work with subsequent operations in `NetCDFOutputWriter`.

navidcy:
@glwagner, now that Thermodynamics.jl merged https://github.com/CliMA/Thermodynamics.jl/pull/210 and [tagged v0.12.6](https://github.com/CliMA/Thermodynamics.jl/releases/tag/v0.12.6), I believe you need to bump the compat requirement in SurfaceFluxes to 0.12.6 at

https://github.com/glwagner/SurfaceFluxes.jl/blob/7c3fac89750e5a115356cea4d30c509d2fcaca48/Project.toml#L22

glwagner:
Ah nice ok I will look into that.

@simone-silvestri we need your help with the merge conflicts.

src/DataWrangling/JRA55.jl
```diff
@@ -81,23 +110,152 @@ urls = Dict(
...
+    j₁, j₂ = compute_bounding_indices(φbounds, φc)
+    TX = infer_longitudinal_topology(λbounds)
+
+    return i₁, i₂, j₁, j₂, TX
+end
+
+# Convert dates to range until Oceananigans supports dates natively
+function jra55_times(native_times, start_time=DateTimeNoLeap(1900, 01, 01))
+    Nt = length(native_times)
+    Δt = native_times[2] - native_times[1] # assume all times are equispaced
```
navidcy, line 186: what if `native_times = 1:1`?

navidcy: `test_downloading.jl` errors because of this (and possibly this implicit assumption elsewhere?)

simone-silvestri: In Oceananigans we assume that a `FieldTimeSeries` with values in memory has at least 2 data points

https://github.com/CliMA/Oceananigans.jl/blob/f7042b738da746cbb6ba752e01c3894b09b6f1a4/src/OutputReaders/field_time_series.jl#L45-L48

I think it is a valid assumption given that a field time series with one point is just a field. 
On the other hand, we have to options
(1) fix the tests in `ClimaOcean.jl` keeping the assumption that `Ntimes > 2` (easy option) 
(2) relax the assumption in `Oceananigans.jl` (harder to maintain but more general?)

I vote for (1)

navidcy:
what happens here

https://github.com/CliMA/ClimaOcean.jl/blob/badffb7fe2fc0490e9f54cdffd46e264aa92b30b/src/DataWrangling/JRA55.jl#L467-L476

if 

```Julia
on_native_grid = true
totally_in_memory = false
```

??

navidcy:
When `on_native_grid = true`, `totally_in_memory = false`, and `backend = OnDisk()` nothing is returned

navidcy:
> if the backend is `OnDisk` then `totally_in_memory = false` and the returning atmosphere is still
> 
> ```julia
>     user_fts = FieldTimeSeries(jld2_filename, fts_name; architecture, backend, time_indexing)
>     return user_fts
> ```

OK, then perhaps the logic is OK but there must be a bug because in this case

https://github.com/CliMA/ClimaOcean.jl/blob/badffb7fe2fc0490e9f54cdffd46e264aa92b30b/src/DataWrangling/JRA55.jl#L481

breaks because


```Julia
JRA55 and data wrangling utilities: Error During Test at /Users/navid/Research/ClimaOcean-v3.jl/test/test_jra55.jl:3
  Got exception outside of a @test
  UndefVarError: `fts` not defined
  Stacktrace:
    [1] JRA55_field_time_series(variable_name::Symbol; architecture::CPU, grid::Nothing, location::Nothing, url::Nothing, filename::Nothing, shortname::Nothing, latitude::Nothing, longitude::Nothing, backend::OnDisk, time_indexing::Oceananigans.OutputReaders.Cyclical{Nothing}, preprocess_chunk_size::Int64, preprocess_architecture::CPU, time_indices::UnitRange{Int64})
...
```

src/DataWrangling/JRA55.jl
```diff
@@ -140,83 +298,319 @@ Keyword arguments
...
-    return fts
+    @show totally_in_memory
+    @info "Pre-processing JRA55 $variable_name data into a JLD2 file..."
+
+    on_disk_fts = FieldTimeSeries{LX, LY, Nothing}(fts.grid;
+                                                   boundary_conditions,
+                                                   backend = OnDisk(),
+                                                   path = jld2_filename,
+                                                   name = fts_name)
+
```
simone-silvestri, line 486: I guess something like this would work

```suggestion
    preprocessing_grid = on_native_grid ? JRA55_native_grid : grid

    on_disk_fts = FieldTimeSeries{LX, LY, Nothing}(preprocessing_grid;
                                                                                     boundary_conditions,
                                                                                     backend = OnDisk(),
                                                                                     path = jld2_filename,
                                                                                     name = fts_name)

```

simone-silvestri:
I think this might be ready to merge when tests pass

simone-silvestri (approved):

navidcy:
Great thanks for that!
I'll try to fix things if there are still pending issues

src/OceanSeaIceModels/CrossRealmFluxes/similarity_theory_turbulent_fluxes.jl
```diff
@@ -0,0 +1,528 @@
...
+    Δθ = θ₁ - θ₀
+
+    q₁ = AtmosphericThermodynamics.vapor_specific_humidity(ℂ, 𝒬₁)
+    q₀ = AtmosphericThermodynamics.vapor_specific_humidity(ℂ, 𝒬₀)
+    Δq = q₁ - q₀
+
+    return Δh, Δu, Δv, Δθ, Δq
+end
+
+@inline function compute_similarity_theory_fluxes(roughness_lengths::ConstantRoughnessLength,
```
glwagner, line 385:
```suggestion
@inline function compute_similarity_theory_fluxes(roughness_lengths,
```

glwagner:
This PR makes progress towards #28

PR #54, opened 2023-11-01
Adds utilities for downloading and working with JRA55 data

### LINKED ISSUE #53: Nuke `neverworld_simulation.jl`
@sandreza @simone-silvestri 

Unfortunately nobody started using `neverworld_simulation` in `ClimaOcean.jl`. Since it takes work to maintain, I'm going to nuke it. Raising this issue in case you have changed your mind and would like to conduct your neverworld work here after all.

Here's the code: https://github.com/CliMA/ClimaOcean.jl/blob/main/src/IdealizedSimulations/neverworld_simulation.jl

glwagner:
Also closes #53 .


src/DataWrangling/JRA55.jl
```diff
@@ -0,0 +1,73 @@
...
+
+    isfile(filename) || download(url, filename)
+
+    ds = Dataset(filename)
+    data = ds[variable_name][:, :, time_indices]
+
+    # Make source field
+    λ = ds["lon_bnds"][1, :]
+    φ = ds["lat_bnds"][1, :]
+    times = ds["time"][time_indices]
```
simone-silvestri, line 44: what are the contents of this ds file? Should we write it somewhere?

navidcy (commented):
I think that somewhere around where RYF urls are defined let's add a comment that this is the Repeat-Year-Forcing from JRE55-do dataset and cite paper by Kial et al., doi:[10.1016/j.ocemod.2019.101557](https://doi.org/10.1016/j.ocemod.2019.101557).

src/DataWrangling/JRA55.jl
```diff
@@ -0,0 +1,73 @@
...
+    jra55_field_time_series(name, architecture=CPU();
+                            time_indices = :,    
+                            url = urls[name],
+                            filename = filenames[name],
+                            variable_name = variable_names[name])
+
+Return a FieldTimeSeries representing JRA55 data at the interface between the
+atmosphere and ocean.
+"""
+function jra55_field_time_series(name, arch=CPU();
```
simone-silvestri, line 30: We need to explain what `name` is

src/DataWrangling/JRA55.jl
glwagner: That's not a bad idea, maybe a comment here...

glwagner: done

src/DataWrangling/JRA55.jl
glwagner: done

navidcy (approved):

src/IdealizedSimulations/neverworld_simulation.jl
navidcy, line 1: @glwagner I want to resurrect these and use them
I'll put in a different repo?

PR #59, opened 2023-11-14
Add interface to ECCO2 data

glwagner:
This PR adds a new module called `ECCO2` with a function `ECCO2.ecco2_field`. We just have temperature, salinity, and ice thickness right now.


simone-silvestri:
merge?

PR #60, opened 2023-11-16
Interpolated inital conditions

simone-silvestri:
## Initial conditions (from ECCO fields but also possible to have custom fields)
This PR implements a tool to initialize ClimaOcean's simulations from interpolated tracer fields. 
The initial fields need to be supplemented by a mask that is 1 on well-defined values and 0 on missing values
The fields need to be massaged to _(i)_ remove unstable layers _(ii)_ remove unphysical tracer stemming coming from misaligned bathymetry _(iii)_ diffuse initial conditions to prevent instability? (not sure yet)

## To-do
- [x] assess if it is better to inpaint on the initial field or the final field (still allow both options)
- [x] assess whether we need to diffuse the initial field or not

Edit: there is no need to diffuse the fields initially. The only caveat is that we need to run the initial 1000/2000 time steps with a low time step size (something like 10 seconds does the trick)


examples/freely_decaying_mediterraneum.jl
```diff
@@ -0,0 +1,119 @@
+using GLMakie
+using Oceananigans
+using Oceananigans: architecture
+using Oceananigans.Utils: prettytime
+using ClimaOcean.Bathymetry: regrid_bathymetry
+using ClimaOcean.ECCO2: ecco2_field, ecco2_center_mask
+using ClimaOcean.VerticalGrids: stretched_vertical_faces, PowerLawStretching
+using ClimaOcean.InitialConditions: three_dimensional_regrid!, adjust_tracers!
```
glwagner, line 8: Should we start exporting all the things we need so we don't have to have these big `using` statements at the top?

examples/freely_decaying_mediterraneum.jl
```diff
@@ -0,0 +1,119 @@
+using GLMakie
+using Oceananigans
+using Oceananigans: architecture
+using Oceananigans.Utils: prettytime
```
glwagner, line 4: This is exported I think?

examples/freely_decaying_mediterraneum.jl
```diff
@@ -0,0 +1,119 @@
+using GLMakie
+using Oceananigans
+using Oceananigans: architecture
```
glwagner, line 3: Might be good to figure out how to get rid of this so users don' thave to worry

simone-silvestri:
This PR Should be ready to go. 
I ll post now tomorrow video from the output of `examples/freely_decaying_mediterranean.jl` 
which is a Mediterranean sea initialized from ECCO temperature and salinity at 1/15th degree resolution

simone-silvestri:
This is the first 31 days of evolution


https://github.com/CliMA/ClimaOcean.jl/assets/33547697/f9a288b3-3861-4d08-bd9a-e504129dceb0

src/DataWrangling/ECCO2.jl
```diff
@@ -121,9 +121,10 @@ end
                 filename = ecco2_file_names[variable_name],
                 short_name = ecco2_short_names[variable_name])
 
-Retrieve the ecco2 field corresponding the `variable_name`, stored in 
-`filename` or dowloaded from `url`. If `user_data` is provided, the 
-field is set with `user_data`
+Retrieve the ecco2 field corresponding to `variable_name`. The data is either:
+(1) retrieved from `filename` 
+(2) dowloaded from `url` if `filename` does not exists
+(3) filled from `user_data` if `user_data` is provided
```
glwagner, line 127: why would we need this function if we have the data in scope? (maybe an example would help)

simone-silvestri:
I have removed that pattern. I have added a struct to describe data from ECCO2 and a method for `set!(field, ::ECCO2Field)`.

Now we can do
```julia
set!(model, T = ECCO2Field(:temperature))
```

`ECCO2Field` also has an associated date that for the moment is just 1/2/1992 since that's our only datapoint

examples/freely_decaying_mediterranean.jl
```diff
@@ -65,16 +65,19 @@ model = HydrostaticFreeSurfaceModel(; grid,
                               tracer_advection = WENO(grid; order = 7),
                                   free_surface = SplitExplicitFreeSurface(; cfl = 0.75, grid),
                                       buoyancy = SeawaterBuoyancy(),
-                                      tracers  = (:T, :S),
+                                      tracers  = (:T, :S, :c),
                                       coriolis = HydrostaticSphericalCoriolis(scheme = ActiveCellEnstrophyConserving()))
```
glwagner, line 69:
```suggestion
                                      coriolis = HydrostaticSphericalCoriolis(scheme = ActiveCellEnstrophyConserving()))
```

Why isn't this default?

src/DataWrangling/ECCO2.jl
```diff
@@ -11,10 +11,23 @@ using Oceananigans.Utils
...
 using NCDatasets
 
+import Oceananigans.Fields: set!
+
 temperature_filename = "THETA.1440x720x50.19920102.nc"
 salinity_filename = "SALT.1440x720x50.19920102.nc"
 effective_ice_thickness_filename = "SIheff.1440x720.19920102.nc"
 
+# Ecco field used to set model's initial conditions
+struct ECCO2Field
```
glwagner:
```suggestion
struct ECCO2FieldInformation
```

so it's not confused with `Field`

glwagner: or `ECCO2FieldDescription`

examples/freely_decaying_mediterranean.jl
```diff
@@ -65,16 +65,19 @@ model = HydrostaticFreeSurfaceModel(; grid,
                               tracer_advection = WENO(grid; order = 7),
                                   free_surface = SplitExplicitFreeSurface(; cfl = 0.75, grid),
```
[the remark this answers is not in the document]
simone-silvestri: not yet in Oceananigans

examples/freely_decaying_mediterranean.jl
simone-silvestri: Probably we should make it default

simone-silvestri:
Changed to `ECCO2Metadata`. I think this PR might be ready to merge

simone-silvestri:
I am not sure why the docs test is failing. @navidcy do you maybe know why this is happening?

navidcy:
> I am not sure why the docs test is failing. @navidcy do you maybe know why this is happening?

Yeap, few docstrings were not included in the docs. I fixed it. I added them in the docs but also added an option so that we only get a warning when that happens; see [3234942](https://github.com/CliMA/ClimaOcean.jl/pull/60/commits/32349420ec18a0560d1687a73aec42ccc734edc4)

src/DataWrangling/inpaint_mask.jl
```diff
@@ -47,7 +47,7 @@ end
 propagate_horizontally!(field, ::Nothing, tmp_field=deepcopy(field); kw...) = field
 
 function propagating(field, mask, iter, max_iter)
-    mask_sum = sum(field; condition=mask)
+    mask_sum = sum(field; condition=interior(mask))
```
glwagner, line 50: ?

examples/freely_decaying_mediterraneum.jl
navidcy: agreed

src/DataWrangling/inpaint_mask.jl
simone-silvestri: just checking a thing, apparently, we cannot use `condition::AbstractField` (because there is no `arch_array` method for fields). We could do two things: (1) extend the reductions for field conditions or (2) have an `arch_array` also for fields. I think the nice thing to do come up with a single function `on_architecture` that is valid for arrays, grids and fields alike

glwagner: agree that is needed. I guess there is an invalid assumption somewhere that `arch_array` works for `AbstractArray`.

navidcy: x-ref: https://github.com/CliMA/Oceananigans.jl/pull/3490

glwagner (approved):
I think this is a good start. We have a bit more work to do to make this feature more general and also work for boundary conditions but that can come in the future.

navidcy:
@glwagner, @simone-silvestri : shall I merge if tests pass ?

glwagner:
yes for sure

PR #63, opened 2024-01-24
Adds a utility to wget ecco2 data from PODAAC JPL

glwagner:
This PR adds a utility to download ECCO2 data from [PODAAC](https://podaac.jpl.nasa.gov)

I need to add a docstring on how to get a user name and password, etc.

A test is more annoying because we have to provide a user name and password. I think there is a way to provide something secretly when running CI somehow but we'll have to figure out how to do that. For now, we can test manually I think.

### DIFF
--- a/src/DataWrangling/ECCO2.jl
+++ b/src/DataWrangling/ECCO2.jl
@@ -1,5 +1,6 @@
 module ECCO2
 
+using Dates
 using Oceananigans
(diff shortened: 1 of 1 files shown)


navidcy:
Don’t worry about a test then. But do add the info in the docstring.

simone-silvestri:
This functionality was introduced in #89

PR #68, opened 2024-03-06
Merge 1/12th degree simulation into existing OMIP infrastructure

glwagner:


simone-silvestri:
Look like this is stale, so I ll close

PR #70, opened 2024-03-26
`arch_array` -> `on_architecture`

navidcy:
`arch_array` was deprecated after https://github.com/CliMA/Oceananigans.jl/pull/3490


glwagner (approved):
Nice though I think we should deprecate `NearGlobalSimulations`... they shouldn't be used...

PR #80, opened 2024-05-02
Some revisitation of turbulent fluxes

simone-silvestri:
This PR revisits surface fluxes to have them produce similar results to COARE 3.6. 

In addition there are some OMIP-specific infrastructural changes
- remove lakes and connected regions to bathymetry
- add a transmissivity and latitude dependent albedo from Payne (1972)

an ongoing PR subjected to changes


glwagner:
how is this PR so big?

src/OceanSeaIceModels/PrescribedAtmospheres.jl
```diff
@@ -291,15 +296,16 @@ struct PrescribedAtmosphere{G, U, P, C, F, R, TP, TI, FT}
...
+    boundary_layer_height :: FT
 end
 
 Base.summary(::PrescribedAtmosphere) = "PrescribedAtmosphere"
 Base.show(io::IO, pa::PrescribedAtmosphere) = print(io, summary(pa))
 
 """
     PrescribedAtmosphere(times;
-                         reference_height,
+                         measurement_height,
```
glwagner, line 308: go back to reference height

(this doesn't refer to any measurements, at least not for JRA55 which is our only example)

src/OceanSimulations/OceanSimulations.jl
```diff
@@ -9,59 +16,92 @@ using SeawaterPolynomials.TEOS10: TEOS10EquationOfState
...
     end
 
     tracers = (:T, :S)
     if closure isa CATKEVerticalDiffusivity
         tracers = tuple(tracers..., :e)
+        tracer_advection = (; T = tracer_advection, S = tracer_advection, e = nothing)
     end
 
-    coriolis = HydrostaticSphericalCoriolis(; rotation_rate)
+    coriolis = HydrostaticSphericalCoriolis(; rotation_rate, scheme = ActiveCellEnstrophyConserving())
```
glwagner, line 104: This is default

https://github.com/CliMA/Oceananigans.jl/blob/6730e6f6b2c8f1695e20b95ef467b5b14fdc4c5f/src/Coriolis/hydrostatic_spherical_coriolis.jl#L41
```suggestion
    coriolis = HydrostaticSphericalCoriolis(; rotation_rate)
```

src/OceanSeaIceModels/PrescribedAtmospheres.jl
simone-silvestri: can we call it something different? In COARE 3.6 reference height is something different. Probably `field_height`?

glwagner: What is the reference height in COARE 3.6?

glwagner: I think this like many other parts should have another PR and ideally an issue so we can discuss separately and come up with something well thought out

simone-silvestri:
It is mostly reorganizing code. We just came out of a PR that added new 8800 lines of code (#49), so there is a lot of fine tuning we can do (for example just splitting the gigantic file `similarity_theory_turbulent_fluxes.jl` in more digestable files which is not really changing anything takes a lot of this PR)

The actual code contribution of this PR is not that large if you remove all the restructuring of the code (which is literally just moving code in new files), the new examples and the new Manifests/Project.toml files.

glwagner:
> It is mostly reorganizing code. We just came out of a PR that added new 8800 lines of code (#49), so there is a lot of fine tuning we can do (for example just splitting the gigantic file `similarity_theory_turbulent_fluxes.jl` in more digestable files which is not really changing anything takes a lot of this PR)
> 
> The actual code contribution of this PR is not that large if you remove all the restructuring of the code (which is literally just moving code in new files), the new examples and the new Manifests/Project.toml files.

Then I think the way I can review this PR is not to look at the changes / diff but rather to sit down together and read the code from top to bottom

src/OceanSeaIceModels/CrossRealmFluxes/similarity_theory_turbulent_fluxes.jl
```diff
@@ -151,7 +151,7 @@ end
 
 # The complete bulk coefficient should include also `ψ(ℓ / L)`, but the 
 # JRA55 atmosphere is adjusted to formulae without this last term so we exclude it
-@inline simplified_bulk_coefficient(ψ, h, ℓ, L) = log(h / ℓ) - ψ(h / L) # + ψ(ℓ / L)
+@inline simplified_bulk_coefficients(ψ, h, ℓ, L) = log(h / ℓ) - ψ(h / L) # + ψ(ℓ / L)
```
glwagner, line 154: Not sure I agree with this logic. JRA55 is also nudged strongly to observations. I think we want to treat it like observations --- the objective is not to set up a simulation that is somehow "consistent" with the coupled system that JRA55 was originally run with.

simone-silvestri: It would be interesting to see the difference between the two formulations (maybe it is not that large). These "similiarity functions" are all parameterizations that include empirical elements or relations that are extrapolated from data, so rigorous mathematical consistency does not necessarily equate to results that are closer to observations.

glwagner: That's not a bad idea, in that case something like 

```julia
inline profile(ψ::SimplifiedProfile, h, ℓ, L) = ...
```

might work.

PS this is the profile, not the bulk coeffcient right? It's what we get when we integrate the similarity gradient.

> rigorous mathematical consistency does not necessarily equate to results that are closer to observations.

The purpose of including the term is not to be "rigorous". It's to avoid "random" approximations that rely on hidden assumptions. I think the same argument is actually the real reason to include it. If including the extra term means that this expression generalizes to cases with larger ``ℓ`` or small ``L`` --- but does not affect the results for typical values of these parameters --- that is exactly why we should include it. Approximations need to have a purpose. The purpose can't be to drop a term in a mathematical expression. It would actually have to make a difference in computational performance or somethign like that to justify this.

src/OceanSeaIceModels/CrossRealmFluxes/similarity_theory_turbulent_fluxes.jl
```diff
@@ -295,8 +298,8 @@ end
     Riᶜ = - h / zᵢ / convert(FT, 0.004) / β^3 # - h / zi / 0.004 / β^3
     
     # Calculating the first stability coefficient and the MO length
-    # TODO: explain this formulation of the stability function. Is it empirical?
-    # Found in COARE3.6
+    # TODO: explain this formulation of the stability function. 
+    # Is it empirical? Found in COARE3.6
```
glwagner, line 302: Can you put a link to the COARE 3.6 code here (I'm not sure where to look)

src/OceanSeaIceModels/CrossRealmFluxes/radiation.jl
```diff
@@ -40,6 +67,38 @@ struct LatitudeDependentAlbedo{FT}
...
+                                 diffuse = 0.069, 
+                                 direct = 0.011) 
+    
+    return LatitudeDependentAlbedo(convert(FT, direct),
+                                   convert(FT, diffuse))
+end
+
+Adapt.adapt_structure(to, α::LatitudeDependentAlbedo) = 
+    LatitudeDependentAlbedo(Adapt.adapt(to, α.direct),                       
+                            Adapt.adapt(to, α.diffuse))
```
glwagner, line 100: solid design 💪

docs/make_examples.jl
```diff
@@ -16,6 +16,7 @@ const OUTPUT_DIR   = joinpath(@__DIR__, "src/literated")
 
 to_be_literated = [
     "inspect_ecco2_data.jl",
+    "JRA55forced_single_column_simulation.jl",
```
glwagner, line 19:
```suggestion
    "single_column_simulation.jl",
```

examples/JRA55forced_single_column_simulation.jl
glwagner, line 1: can we name this `single_column_simulation.jl`? That will work provided we don't have other single column examples (right now we just have one)

src/OceanSeaIceModels/CrossRealmFluxes/roughness_lengths.jl
```diff
@@ -0,0 +1,132 @@
+struct MomentumRoughnessLength{FT, V}
+    gravitational_acceleration :: FT
+    air_kinematic_viscosity :: V
+    _wave_parameter :: FT
```
glwagner, line 4: why does this have an underscore?

src/OceanSeaIceModels/CrossRealmFluxes/roughness_lengths.jl
```diff
@@ -0,0 +1,132 @@
...
+- `maximum_roughness_length::Float`: The maximum roughness length. Default is `1.0`.
+- `air_kinematic_viscosity::Float`: The air kinematic viscosity. Default is `temperature_dependent_viscosity`.
+- `_wave_parameter::Float`: The wave parameter. Default is `0.011`.
+- `laminar_parameter::Float`: The laminar parameter. Default is `0.11`.
+"""
+function MomentumRoughnessLength(FT=Float64;
+                                 gravitational_acceleration = default_gravitational_acceleration,
+                                 maximum_roughness_length = 1.0, # An estimate?
+                                 air_kinematic_viscosity = temperature_dependent_viscosity,
+                                 _wave_parameter = 0.011,
```
glwagner, line 63:
```suggestion
                                 gravity_wave_parameter = 0.011,
```

src/OceanSeaIceModels/CrossRealmFluxes/roughness_lengths.jl
```diff
@@ -0,0 +1,132 @@
...
+
+# Empirical fit of the scalar roughness length with roughness Reynolds number `R★ = u★ ℓu / ν`
+# Edson et al. (2013), equation (28)
+@inline empirical_scaling_function(R★ :: FT, args...) where FT = 
+        ifelse(R★ == 0, FT(0), convert(FT, 5.85e-5 / R★ ^ 0.72))
+
+# Temeprature-dependent viscosity law: assumes that θ comes in Kelvin
+@inline function temperature_dependent_viscosity(θ :: FT) where FT 
+    T = convert(FT, θ - celsius_to_kelvin)
+    ν = convert(FT, 1.326e-5 * (1 + 6.542e-3 * T + 8.301e-6 * T^2 - 4.84e-9 * T^3))
```
glwagner, line 88: this is hard coded!

glwagner: Make `TemperatureDependentViscosity` a type and put these parameters in. Convert the parameters when you make the type

src/OceanSeaIceModels/CrossRealmFluxes/roughness_lengths.jl
```diff
@@ -0,0 +1,132 @@
...
+    momentum    = MomentumRoughnessLength(FT)
+    temperature = ScalarRoughnessLength(FT)
+    water_vapor = ScalarRoughnessLength(FT)
+    return SimilarityScales(momentum, temperature, water_vapor)
+end
+
+# Empirical fit of the scalar roughness length with roughness Reynolds number `R★ = u★ ℓu / ν`
+# Edson et al. (2013), equation (28)
+@inline empirical_scaling_function(R★ :: FT, args...) where FT = 
+        ifelse(R★ == 0, FT(0), convert(FT, 5.85e-5 / R★ ^ 0.72))
```
glwagner, line 83: Don't hardcode _any_ parameters at all

src/OceanSeaIceModels/CrossRealmFluxes/stability_functions.jl
```diff
@@ -0,0 +1,127 @@
...
+    p₄ :: FT
+end
+
+struct ScalarStabilityFunction{FT}
+    p₁ :: FT
+    p₂ :: FT
+    p₃ :: FT
+    p₄ :: FT
+    p₅ :: FT
+end
```
glwagner, line 44: Do we need different types for momentum and scalar or can we just have `StabilityFunction`

src/OceanSeaIceModels/CrossRealmFluxes/stability_functions.jl
```diff
@@ -0,0 +1,127 @@
...
+
+function MomentumStabilityFunction(FT = Float64;
+                                   p₁ = convert(FT, 0.35),
+                                   p₂ = convert(FT, 0.7),
+                                   p₃ = convert(FT, 15),
+                                   p₄ = convert(FT, 10.15))
```
glwagner, line 50: I think it might be better to provide constructors with default parameters a different way like a function

```julia
function edson_2013_momentum_stability_function(FT, p....)
    return StabilityFunction(p...)
end
```

sketching it out

src/OceanSeaIceModels/CrossRealmFluxes/stability_functions.jl
```diff
@@ -0,0 +1,127 @@
...
+    p₂ = ψ.p₂
+    p₃ = ψ.p₃
+    p₄ = ψ.p₄
+
+    ζ⁻ = min(zero(ζ), ζ)
+    ζ⁺ = max(zero(ζ), ζ)
+    dζ = min(50, p₁ * ζ⁺)
+
+    # stability function for stable atmospheric conditions 
+    ψ_stable = - p₂ * ζ⁺ - 3 / 4 * (ζ⁺ - 5 / p₁) * exp(-dζ) - 3 / 4 * 5 / p₁
```
glwagner, line 84: I'm confused what the parameters are here. We have `3/4`, `5`, etc. Also `3/4` is not type stable. I think you may want to rewrite these to expose the true free parameters

src/OceanSeaIceModels/CrossRealmFluxes/stability_functions.jl
```diff
@@ -0,0 +1,127 @@
...
+    p₃ = ψ.p₃
+    p₄ = ψ.p₄
+    p₅ = ψ.p₅
+
+    ζ⁻ = min(zero(ζ), ζ)
+    ζ⁺ = max(zero(ζ), ζ)
+    dζ = min(50, p₁ * ζ⁺)
+
+    # stability function for stable atmospheric conditions 
+    ψ_stable = - (4 * ζ⁺ / 3)^(3 / 2) - 2 / 3 * (ζ⁺ - p₂) * exp(-dζ) - p₃
```
glwagner, line 113: Why isn't `3/2` a parameter?

src/OceanSeaIceModels/CrossRealmFluxes/tabulated_albedo.jl
```diff
@@ -0,0 +1,147 @@
...
+                  0.581 0.494 0.393 0.333 0.288 0.237 0.211 0.182 0.158 0.141 0.126 0.110 0.095 0.083 0.074 0.066 0.061 0.057 0.053 0.049 0.045 0.041 0.039 0.034 0.033 0.032 0.031 0.030 0.029 0.028 0.028 0.028 0.028 0.027 0.026 0.026 0.026 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.02
+                  0.453 0.398 0.342 0.301 0.266 0.226 0.205 0.180 0.157 0.140 0.125 0.109 0.095 0.083 0.074 0.065 0.061 0.057 0.052 0.048 0.044 0.040 0.038 0.033 0.032 0.031 0.030 0.029 0.028 0.027 0.027 0.026 0.026 0.026 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.02
+                  0.425 0.370 0.325 0.290 0.255 0.220 0.200 0.178 0.157 0.140 0.122 0.108 0.095 0.083 0.074 0.065 0.061 0.056 0.052 0.048 0.044 0.040 0.038 0.033 0.032 0.031 0.030 0.029 0.028 0.027 0.026 0.026 0.026 0.026 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.025 0.02]
+
+"""
+    TabulatedAlbedo(arch = CPU(), FT = Float64;
+                   S₀ = convert(FT, 1365),
+                   α_table  = α_payne,
+                   φ_values = (0:2:90) ./ 180 * π,
+                   𝓉_values = 0:0.05:1)
```
glwagner, line 51:
```suggestion
                    S₀ = convert(FT, 1365),
                    α_table  = α_payne,
                    φ_values = (0:2:90) ./ 180 * π,
                    𝓉_values = 0:0.05:1)
```

src/OceanSeaIceModels/CrossRealmFluxes/tabulated_albedo.jl
```diff
@@ -0,0 +1,147 @@
...
+
+Note: `TabulatedAlbedo` assumes that the latitude and the transissivity in the table are uniformly spaced. 
+
+The transmissivity of the atmosphere is calculated as the ratio of the downwelling solar radiation to the
+maximum possible downwelling solar radiation for a transparent atmosphere, function of hour of the day, latitude,
+and day in the year.
+
+# Arguments
+============
+
```
glwagner, line 64:
```suggestion
```

src/OceanSeaIceModels/CrossRealmFluxes/tabulated_albedo.jl
```diff
@@ -0,0 +1,147 @@
...
+function TabulatedAlbedo(arch = CPU(), FT = Float64;
+                         S₀ = convert(FT, 1365),
+                         α_table  = α_payne,
+                         φ_values = (0:2:90) ./ 180 * π,
+                         𝓉_values = 0:0.05:1)
+
+    # Make everything GPU - ready
+    α_table  = on_architecture(arch, α_table)
+    φ_values = on_architecture(arch, φ_values) 
+    𝓉_values = on_architecture(arch, 𝓉_values)
```
glwagner, line 83: gotta convert to `FT` too right?

src/OceanSeaIceModels/CrossRealmFluxes/tabulated_albedo.jl
```diff
@@ -0,0 +1,147 @@
...
+
+    φ = deg2rad(φ)
+    λ = deg2rad(λ)
+    time = time.time
+
+    day     = time ÷ 86400
+    day2rad = 2π / 86400
+
+    noon_in_sec = 86400 ÷ 2
+    sec_of_day  = time - day * 86400
```
glwagner, line 106: This introduces a convention for floating point time. Which is ok but I wonder if we don't have to hardcode this, can it be an option?

src/OceanSeaIceModels/PrescribedAtmospheres.jl
glwagner: Still wondering what the reference height is in COARE 3.6

src/OceanSeaIceModels/ocean_only_model.jl
```diff
@@ -1,54 +1,34 @@
 const OceanOnlyModel = OceanSeaIceModel{Nothing}
+const OceanSimplifiedSeaIceModel = OceanSeaIceModel{<:MinimumTemperatureSeaIce}
 
-OceanOnlyModel(ocean; kw...) = OceanSeaIceModel(nothing, ocean; kw...)
+const NoSeaIceModel = Union{OceanOnlyModel, OceanSimplifiedSeaIceModel}
 
 #####
-##### No ice-ocean fluxes in this model!!
+##### No ice-ocean fluxes in this models!!
```
glwagner, line 7:
```suggestion
##### No ice-ocean fluxes in these models!!
```

src/OceanSimulations/OceanSimulations.jl
```diff
@@ -9,60 +17,95 @@ using SeawaterPolynomials.TEOS10: TEOS10EquationOfState
...
 # function that requires latitude and longitude etc for computing coriolis=FPlane...
-function ocean_simulation(grid;
+function ocean_simulation(grid; Δt = 5minutes,
                           closure = default_ocean_closure(),
                           free_surface = default_free_surface(grid),
                           reference_density = 1020,
                           rotation_rate = Ω_Earth,
-                          gravitational_acceleration = g_Earth)
+                          gravitational_acceleration = g_Earth,
+                          drag_coefficient = 0.003,
```
glwagner, line 64:
```suggestion
                          bottom_drag_coefficient = 0.003,
```

?

We may want to implement something neater eventually

glwagner:
This is looking decent. Any results from a simple regional simulation yet?

src/OceanSeaIceModels/PrescribedAtmospheres.jl
simone-silvestri: I couldn't really find it. I think `reference_height` is ok

src/Bathymetry.jl
```diff
@@ -228,7 +229,7 @@ function interpolate_bathymetry_in_passes(native_h, target_grid;
     h_data = Array(interior(target_h, :, :, 1))
 
     if minimum_depth > 0
-        shallow_ocean = h_data .> minimum_depth
+        shallow_ocean = h_data .> - minimum_depth
```
glwagner, line 232:
```suggestion
        shallow_ocean = z_data .> - minimum_depth
```

`h` is depth, `z` is vertical coordinate...

glwagner: and you would have probably wanted `h < minimum_depth` right?

simone-silvestri: In this case `h` is the topography as measured from sea-level, so positive for mountains and negative for the seabed. What about just `topography` as a name?

glwagner: Use the symbol "z" somehow to indicate that it's a z-coordinate

glwagner: `z_topography` if you like

glwagner:
This is passing. Should we merge? We have a bunch of bugs to addresss it seems so better to get this big PR merged sooner rather than later.

glwagner (approved):

simone-silvestri:
Sounds good

simone-silvestri:
Oh shoot, I realized I changed `h_data` to `z_data` in the bathymetry file in another branch, not this one. I ll do a quick PR to change that name

glwagner:
Makes progress towards #28

PR #88, opened 2024-06-20
Some fixes after PR #80

simone-silvestri:
This PR changes all `h_data` to `z_data` in the `Bathymetry` file and removes the `interpolate!` function which has already been changed in Oceananigans

### DIFF
--- a/src/Bathymetry.jl
+++ b/src/Bathymetry.jl
@@ -100,12 +100,12 @@ function regrid_bathymetry(target_grid;
 
     φ_data = dataset["lat"][:]
     λ_data = dataset["lon"][:]
-    h_data = convert.(FT, dataset["z"][:, :])
+    z_data = convert.(FT, dataset["z"][:, :])
 
     # Convert longitude to 0 - 360?
     λ_data .+= 180
-    nhx      = size(h_data, 1)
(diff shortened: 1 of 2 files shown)


glwagner (approved):

src/Bathymetry.jl
```diff
@@ -100,12 +100,12 @@ function regrid_bathymetry(target_grid;
 
     φ_data = dataset["lat"][:]
     λ_data = dataset["lon"][:]
-    h_data = convert.(FT, dataset["z"][:, :])
+    z_data = convert.(FT, dataset["z"][:, :])
```
glwagner, line 103: Is this an array? You can also use

```julia
z_data = convert(Array{FT}, dataset["z"][:, :])
```

which I find a little more explicit.

src/Bathymetry.jl
```diff
@@ -219,59 +224,59 @@ function interpolate_bathymetry_in_passes(native_h, target_grid;
 
         new_h = Field{Center, Center, Nothing}(new_grid)
```
glwagner, line 225:
```suggestion
        new_z = Field{Center, Center, Nothing}(new_grid)
```

PR #89, opened 2024-06-20
ECCO Restoring for OMIP simulations

simone-silvestri:
This PR reworks the whole ECCO2 module, renaming it to ECCO since it adds three different "versions"
- ECCO2 daily data
- ECCO2 monthly averages
- ECCO4 monthly averages

It also introduces an `ECCO_field_time_series` with its associated `ECCONetCDFbackend` and an `ECCO_restoring_forcing` on top of that to restore prognostic variables to ecco data.
The ECCO data is downloaded from the ecco website by providing an `ECCO_USERNAME` and an `ECCO_PASSWORD` as environmental variables

TODOs:
- [x] make sure that tests pass
- [x] speed up the restoring (now it is very slow for some reason)
- [x] put username and password in github secret tokens to allow downloading ecco data for testing / documentation purposes


glwagner:
This PR makes progress towards #28

src/DataWrangling/ecco_restoring.jl
```diff
@@ -87,7 +87,7 @@ function ecco_times(metadata; start_time = first(metadata).dates)
         push!(times, time)
     end
 
-    return times
+    return tuple(times...)
```
glwagner, line 90: the docstring states that this returns an array, but this is a tuple.

Note that using tuples could be problematic if the tuple is really big

simone-silvestri: yep, the slowdown was caused by the tupled times. That should be fixed with https://github.com/CliMA/Oceananigans.jl/pull/3639 so now we can use `CuArray` times

simone-silvestri:
This PR is almost ready to review, everything works quite nicely. I only need to fix the docstrings that lagged behind and make sure we can download ECCO data for the tests. @navidcy I probably need your help for the last point

simone-silvestri:
I was thinking to also make sure the examples are all well massaged but that can happen probably in another PR

simone-silvestri:
> @navidcy I probably need your help for the last point

I think I have found a solution

simone-silvestri:
This PR is ready to review

simone-silvestri:
@glwagner @navidcy ready to review.

examples/ecco_restoring_mediterranean.jl
glwagner, line 1: The filename doesn't give much of a clue as to what this does. Is this a simulation?

glwagner: how about `ecco_restored_mediterranean_simulation` or `mediterranean_simulation_with_ecco_restoring`

examples/generate_atmos_dataset.jl
```diff
@@ -4,9 +4,9 @@ using JLD2
 
 time_indices = 1:1
 
-qt = ClimaOcean.JRA55.jra55_field_time_series(:specific_humidity; time_indices)
-Tt = ClimaOcean.JRA55.jra55_field_time_series(:temperature; time_indices)
-pt = ClimaOcean.JRA55.jra55_field_time_series(:sea_level_pressure; time_indices)
+qt = ClimaOcean.JRA55.JRA55_field_time_series(:specific_humidity; time_indices)
+Tt = ClimaOcean.JRA55.JRA55_field_time_series(:temperature; time_indices)
+pt = ClimaOcean.JRA55.JRA55_field_time_series(:sea_level_pressure; time_indices)
```
glwagner, line 9: Is this related to this PR?

For changes that are unrelated to a PR just put them in a different PR.

src/Bathymetry.jl
```diff
@@ -116,7 +117,7 @@ function regrid_bathymetry(target_grid;
     close(dataset)
 
     # Diagnose target grid information
-    arch = architecture(target_grid)
+    arch = child_architecture(architecture(target_grid))
```
glwagner, line 120:
```suggestion
    arch = child_architecture(target_grid)
```

src/DataWrangling/ecco_metadata.jl
```diff
@@ -0,0 +1,173 @@
...
+import Base
+
+struct ECCO2Monthly end
+struct ECCO2Daily end
+struct ECCO4Monthly end
+
+# Metadata holding the ECCO dataset information:
+# - `name`: The name of the dataset.
+# - `dates`: The dates of the dataset, in a `AbstractCFDateTime` format.
+# - `version`: The version of the dataset, could be ECCO2Monthly, ECCO2Daily, or ECCO4Monthly.
```
glwagner, line 14: why are all the versions "ECCO2" but the type here is ECCO?

src/DataWrangling/ecco_metadata.jl
```diff
@@ -0,0 +1,173 @@
...
+Base.size(data::ECCOMetadata{<:Any, <:ECCO4Monthly}) = (720,  360, 50, length(data.dates))
+
+Base.size(::ECCOMetadata{<:AbstractCFDateTime, <:ECCO2Daily})   = (1440, 720, 50, 1)
+Base.size(::ECCOMetadata{<:AbstractCFDateTime, <:ECCO2Monthly}) = (1440, 720, 50, 1)
+Base.size(::ECCOMetadata{<:AbstractCFDateTime, <:ECCO4Monthly}) = (720,  360, 50, 1)
+
+# The whole range of dates in the different dataset versions
+all_ecco_dates(::ECCO4Monthly) = DateTimeProlepticGregorian(1992, 1, 1) : Month(1) : DateTimeProlepticGregorian(2023, 12, 1)
+all_ecco_dates(::ECCO2Monthly) = DateTimeProlepticGregorian(1992, 1, 1) : Month(1) : DateTimeProlepticGregorian(2023, 12, 1)
+all_ecco_dates(::ECCO2Daily)   = DateTimeProlepticGregorian(1992, 1, 4) : Day(1)   : DateTimeProlepticGregorian(2023, 12, 31)
```
glwagner, line 60: for clarify, you probably want to define things like

```julia
const ECCO2_final_time = DateTimeProlepticGregorian(2023, 12, 1)
```

or something. And put them near the definition of the struct `ECCO2Monthly` for example. Because these things will change in the future presumably.

src/DataWrangling/ecco_metadata.jl
```diff
@@ -0,0 +1,173 @@
...
+
+# File name generation specific to each Dataset version
+function file_name(metadata::ECCOMetadata{<:AbstractCFDateTime, <:ECCO4Monthly})
```
glwagner, line 63: we use `filename` almost exclusively everywhere so I suggest keeping that style here. (technically `filename` is one word, but mainly I think we just want to be consistent)

src/DataWrangling/ecco_metadata.jl
```diff
@@ -0,0 +1,173 @@
...
+or by launching julia with 
+
+ECCO_USERNAME=myuser ECCO_PASSWORD=mypasswrd julia 
+
+# Arguments
+- `metadata::ECCOMetadata`: The metadata specifying the dataset to be downloaded.
+"""
+function download_dataset!(metadata::ECCOMetadata;
+                           username = get(ENV, "ECCO_USERNAME", nothing), 
+                           password = get(ENV, "ECCO_PASSWORD", nothing),
```
glwagner, line 146: is this insecure? This invites people to write their passwords in a script which could get uploaded to git.

You may want to remove the option of providing `username` and `password` via kwarg.

src/DataWrangling/ecco_restoring.jl
```diff
@@ -0,0 +1,284 @@
...
+- `ecco_grid`: The native ECCO grid to interpolate from.
+- `mask`: A mask (could be a number, an array, a function or a field).
+- `varname`: The variable name of the variable that needs restoring.
+- `λ⁻¹`: The reciprocal of the restoring timescale.
+"""
+struct ECCORestoring{FTS, G, M, V, N} <: Function
+    ecco_fts  :: FTS
+    ecco_grid :: G
+    mask      :: M
+    varname   :: V
```
glwagner, line 184: in the JRA55 module we use `variable_name`. Do we want to use `varname` or `variable_name`? We should pick one.

glwagner (approved):

src/DataWrangling/ecco_metadata.jl
simone-silvestri: not sure, `filename` is a variable given as a keyword argument and `file_name` is a function with `metadata` as an input. We can change the name of the function but `filename` should be the variable in my opinion. In JRA55 the pattern is `filename = filenames[variable]` (a dictionary) but I don't think that is a pattern to take example from because it does not generalize to multiple files. 

I wanted to change JRA55 to allow multiple year forcing, in that case we have to come up with a similar pattern of metadata and `function how_we_want_to_call_it(metadata) = the_filename` but we should do that in a new PR.

src/DataWrangling/ecco_metadata.jl
simone-silvestri: removed!

src/DataWrangling/ecco_restoring.jl
simone-silvestri: I prefer `varname` (personal preference) but I am ok with `variable_name`.

src/DataWrangling/ecco_metadata.jl
simone-silvestri: What about `filename = metadata_filename(metadata)`

simone-silvestri:
@navidcy there is some problems with the docs (the operation is cancelled for some kind of reason). Can you take a look to see what is happening?

src/DataWrangling/ecco_metadata.jl
glwagner: `metadata_filename` is good.

src/DataWrangling/ecco_restoring.jl
glwagner: It just departs from the rest of the style of the code which is to use words for names

simone-silvestri: ok, I have changed it to variable_name

src/DataWrangling/ecco_metadata.jl
simone-silvestri: there is one ECCO4Montly version also.

examples/generate_atmos_dataset.jl
simone-silvestri: hmmm that looks like it's a bugfix. The function name is `JRA55_field_time_series`

examples/ecco_restoring_mediterranean.jl
simone-silvestri: changed!

examples/generate_atmos_dataset.jl
glwagner: in the future if you find bugs like this put them in a new PR instead of loading them into an unrelated PR. It will help move things forward faster because it will disentangle PRs.

src/OceanSeaIceModels/minimum_temperature_sea_ice.jl
```diff
@@ -55,7 +55,9 @@ end
         cooling_sea_ice = sea_ice & (Jᵀ[i, j, 1] > 0)
 
         # Don't allow the ocean to cool below the minimum temperature! (make sure it heats up though!)
-        Jᵀ[i, j, 1] = ifelse(cooling_sea_ice, zero(grid), Jᵀ[i, j, 1]) 
+	    # On the other hand, assume all heating fluxes are due to solar radiation and 
+	    # multiply by a fictitious albedo
+	    Jᵀ[i, j, 1] = ifelse(cooling_sea_ice, zero(grid), Jᵀ[i, j, 1])
```
glwagner, line 60:
```suggestion
	Jᵀ[i, j, 1] = ifelse(cooling_sea_ice, zero(grid), Jᵀ[i, j, 1]) 
```

glwagner:
Maybe change `ECCO_USERNAME` and `ECCO_PASSWORD` to `CLIMAOCEAN_ECCO_USERNAME` and `CLIMAOCEAN_ECCO_PASSWORD` to clarify that these are used in clima ocean

simone-silvestri:
Given that the tests pass I will merge this PR. 
Note that the examples are removed from the docs build here as they are tackled in #104

PR #95, opened 2024-06-21
Use Scratch for Bathymetry

vchuravy:
The current directory might not be writeable and this way multiple copies of the code can share the same data.

It might make sense to use something like DataDeps/Artifacts in the long run.


vchuravy:
Probably also need to do something similar for `RYF.*`

glwagner:
cc @simone-silvestri

simone-silvestri:
Looks good! should we include a scratch for the JRA55-downloaded files in this PR?

glwagner:
> Looks good! should we include a scratch for the JRA55-downloaded files in this PR?

yes

glwagner:
@simone-silvestri please review this

simone-silvestri (approved):

PR #99, opened 2024-07-01
Improve notation for computing turbulent fluxes

glwagner:
This PR is a WIP because I think we should have an interface for optionally including gustiness.


glwagner:
Actually I think the gusintess should be part of the `bulk_velocity`

glwagner:
A few more changes --- I think the `fields` should be outside the similarity theory fluxes probably. It'd be nice to have the option to toggle on and off all of the fluxes including the outgoing longwave for example... even though some of these can be recomputed

simone-silvestri:
makes sense to have fields in `OceanSeaIceSurfaceFluxes` instead of `SimilarityTheoryTurbulentFluxes`.
the gustiness is optional for the moment by setting `gustiness_parameter = 0`. I am not really sure about including it in the bulk velocity because it is really a parameter of the similarity theory, it depends on a buoyancy scale.

glwagner:
> makes sense to have fields in `OceanSeaIceSurfaceFluxes` instead of `SimilarityTheoryTurbulentFluxes`. the gustiness is optional for the moment by setting `gustiness_parameter = 0`. I am not really sure about including it in the bulk velocity because it is really a parameter of the similarity theory, it depends on a buoyancy scale.

Why does it matter whether it is a parameter of similarity theory?

The `bulk_velocity` represents how we compute the velocity scale that enters into bulk formula, which have a form `θ★ = C * Δθ * U` where `U` is the bulk velocity scale. The reason for this abstraction is because there are a few different parameterizations for `U`. For example, some models ignore the ocean velocity when computing `U`. Other models include a gustiness as here. One could consider additional parameterizations, perhaps dependent on the surface wave state.

What do you propose as a different intent for the purpose of the `bulk_velocity` property? I don't completely understand what you mean with your argument that the gustiness is not a parameter of the similarity theory. Should parameterizations of `bulk_velocity` only be specific to similarity theory? Why would we want to do that?

simone-silvestri:
I meant that the gustiness gets "solved for" by iterating while the bulk velocity is calculated once at the beginning of the iteration, for that reason I meant that the gustiness is more related to the similarity theory than the bulk velocity, but we could recompute the velocity _if_ using a bulk velocity that includes gustiness

src/OceanSeaIceModels/CrossRealmFluxes/similarity_theory_turbulent_fluxes.jl
```diff
@@ -388,10 +392,11 @@ end
...
     # Buoyancy flux characteristic scale for gustiness (Edson 2013)
     Jᵇ = - u★ * b★
-    uᴳ = β * cbrt(Jᵇ * zᵢ)
+    Uᴳ = β * cbrt(Jᵇ * hᵢ)
 
     # New velocity difference accounting for gustiness
-    ΔUᴳ = sqrt(Δu^2 + Δv^2 + uᴳ^2)
+    ΔU = sqrt(Δu^2 + Δv^2 + Uᴳ^2)
 
     return SimilarityScales(u★, θ★, q★), ΔUᴳ
```
simone-silvestri, line 400:
```suggestion
    return SimilarityScales(u★, θ★, q★), ΔUᴳ
```
```suggestion
    return SimilarityScales(u★, θ★, q★), ΔU
```

glwagner:
needs review @simone-silvestri

simone-silvestri (approved):

PR #104, opened 2024-07-24
Run Examples on GPU for Documentation

simone-silvestri:
This PR is just a test to check that the documentation build triggers correctly on the GPU on the Caltech cluster and the results are literated correctly.

The setup now is that the documentation build without examples runs through github actions every commit while the documentation with examples runs every commit when the label `Generate docs` is attached (and stops running if we remove the label) and (should) run every time the PR is merged. 

thanks @Sbozzolo for helping with the setup

preview here: https://clima.github.io/ClimaOceanDocumentation/previews/PR104/


glwagner:
Nice! Just a small tweak, rather than "climaocean-heavydocs" should we simply call it "climaocean-examples" ? We do expect external contributors I think so clear labels might be useful.

glwagner:
As for merge policy, can we adopt a policy that small PRs may go through without building the examples? I guess we will have to exercise some judgement but if the examples take several hours or a day to run fully, we won't want to have to run them for every PR.

simone-silvestri:
Agreed, I ll take a look into how to make sure that we can trigger the documentation at will. I think we can probably have an option that, if the PR has the `Generate docs` label then, once merged, the build will trigger, otherwise not

glwagner:
> Agreed, I ll take a look into how to make sure that we can trigger the documentation at will. I think we can probably have an option that, if the PR has the `Generate docs` label then, once merged, the build will trigger, otherwise not

Oh sorry I was confused! The build is triggered by the _label_.

Ok, so a subsequent step could involve writing a bot that opens a PR nightly, or every few days (say) with the label attached. And we can open PRs manually more frequently as well.

glwagner:
Will we have doc previews? I think we disabled those for Oceananigans because they started filling up OceananigansDocumentation but we might want to enable them for a little bit here

simone-silvestri:
It should be enabled

glwagner:
Do you have the link? Looks like the build is still pending but will be good to be able to check them out once it runs.

simone-silvestri:
Given the documentation of `deploydocs`,  the preview should be generated in `https://github.com/CliMA/ClimaOceanDocumentation/previews/PR104`, but I don't see the page. Maybe it need to build first and it hasn't been built once yet

navidcy:
Yes, it needs to be built at least once for the page to be created.
Also we need to ensure that Documenter has the proper ssh keys to push. But I can do that!

glwagner:
Nice, the examples CI ran! It has a bug though GPU-related

glwagner:
Do you have a link for the PR preview?

simone-silvestri:
Is not there yet. I think the docs will not build until the pipeline completes without errors. Also I am not sure about this 

>  Also we need to ensure that Documenter has the proper ssh keys to push. But I can do that!

simone-silvestri:
Uff, too much parameter space in these simulations. I ll try simplifying them a tad and see if they go through

simone-silvestri:
The fixes should be working, but we get agent lost...

simone-silvestri:
Ok, fixed the agent lost problem, now we need to deploy the docs to check them.

glwagner:
Can you paste a link to the docs preview so we can check it out?

glwagner:
Why don't we use `execute=true` here:

https://github.com/CliMA/ClimaOcean.jl/blob/f2f9d6b928138986c02fb5e4bc36528cd688e9a0/docs/make.jl#L27

We do this for Oceananiganas:

https://github.com/CliMA/Oceananigans.jl/blob/main/docs/make.jl#L58-L59

We also set the environment variable `JULIA_DEBUG=Documenter` which I believe gives us more info as the examples run:

https://github.com/CliMA/Oceananigans.jl/blob/406eb9c5c7a9fc86947747116128c8c1ba4c93d4/docs/make.jl#L56

We should probably adopt these here.

.buildkite/examples_build.yml
```diff
@@ -42,8 +42,15 @@ steps:
       slurm_ntasks: 1
       slurm_gpus_per_task: 1
       slurm_time: 48:00:00
+
     env:
       JULIA_NUM_PRECOMPILE_TASKS: 8
+      JULIA_DEBUG: "Documenter"
+      # This environment variable is needed to avoid SSL verification errors when Downloads.jl 
+      # tries to download the bathyemtry data. It should not be required so we need to fix our certificates
```
glwagner, line 50:
```suggestion
      # tries to download the bathymetry data. It should not be required so we need to fix our certificates
```

glwagner:
Do you have a link for the PR previews?

simone-silvestri:
When it will be deployed it will be at https://USER.github.io/ClimaOceanDocumentation/previews/PR#104

simone-silvestri:
The global simulation ran on the GPU! The whole docs take about 4 hours with the near global simulation running for 10 days (at 1 minute time step).

glwagner:
I don't see the docs preview yet. I think the link is wrong too, it would be

https://clima.github.io/ClimaOceanDocumentation/previews/PR104

(no hash) --- but also that link does not work.

simone-silvestri:
The documentation did not deploy yet because we need to add deploy keys. I will figure out how to do it today

glwagner:
Would be nice to have it done before tomorrow!

glwagner:
@simone-silvestri examples build failed

navidcy:
Are the keys OK now? I think you need to put the "secret" in this repo and the public in the Documentation repo.

simone-silvestri:
I think they are ok but we will know once they finish building. If it doesn't build now I ll remove the examples and test the build. We have set up the secret key as an environment variable in the buildkite hooks and the public one in the ClimaOceanDocumentation repository.

navidcy:
> I think they are ok but we will know once they finish building. If it doesn't build now I ll remove the examples and test the build. We have set up the secret key as an environment variable in the buildkite hooks and the public one in the ClimaOceanDocumentation repository.

Yes, perhaps just for testing leave an example that takes 1min and when you see things are pushed bring everything back on.

navidcy:
OK, Github pushes now. But there is an issue with file size! Let's fix this because if we start pushing docs of 500MB the repo will be out of size in 30 commits

navidcy:
Are there any other files produced by the docs built that are not `.nc` and `.jld2`? Because those are cleaned out before deployment.

navidcy:
just 2 files alone > 0.5GB!

```
remote: error: File previews/PR104/literated/inspect_ecco_data/index.html is 390.04 MB; this exceeds GitHub's file size limit of 100.00 MB
remote: error: File previews/PR104/search_index.js is 415.77 MB; this exceeds GitHub's file size limit of 100.00 MB
```

simone-silvestri:
I guess that is the curse with a high-resolution video of the ocean surface state. 
Then let's just have the simulation spit out an image. It is a little bit annoying because I hoped to have a video of a quarter degree simulation running in the docs but maybe an image of the final state is enough.

simone-silvestri:
or maybe we can use github LFS to store large files. I ll take a look. In the meantime I am going to kickoff some docs that spit out only a snapshot of the final state so we can at least look at the docs preview

examples/near_global_omip_simulation.jl
```diff
@@ -1,97 +1,133 @@
+# # Near-global Ocean Simulation
+#
+# This Julia script sets up and runs a near-global ocean simulation using the Oceananigans.jl and ClimaOcean.jl packages. 
+# The simulation covers latitudes from 75°S to 75°N with a horizontal resolution of 1/4 degree and 40 vertical levels.
```
navidcy, line 4:
```suggestion
# This example sets up and runs a near-global ocean simulation using the Oceananigans.jl and ClimaOcean.jl packages. 
# The simulation covers latitudes from 75°S to 75°N with a horizontal resolution of 1/4 degree and 40 vertical levels. 
```

navidcy:
can git lfs repos have a gh-page hosted?

navidcy:
> just 2 files alone > 0.5GB!
> 
> ```
> remote: error: File previews/PR104/literated/inspect_ecco_data/index.html is 390.04 MB; this exceeds GitHub's file size limit of 100.00 MB
> remote: error: File previews/PR104/search_index.js is 415.77 MB; this exceeds GitHub's file size limit of 100.00 MB
> ```

from this, it doesn't seem that it's the `mp4` files that are big... `search_indes.jl` 415MB???

examples/near_global_ocean_simulation.jl
```diff
@@ -0,0 +1,268 @@
...
+# including grid setup, physical processes modeling, and data visualization.
+
+using Oceananigans
+using Oceananigans.Units
+using Oceananigans: architecture, on_architecture
+using ClimaOcean
+using ClimaOcean.ECCO
+using ClimaOcean.OceanSimulations
+using ClimaOcean.OceanSeaIceModels
+using OrthogonalSphericalShellGrids
```
navidcy, line 22: Do we need `OrthogonalSphericalShellGrids` here?

simone-silvestri: No here we don't. I was trying to make this a global simulation but it looks like with the `OrthogonalSphericalShellGrid` we incur in parameter space limitations on the calthech cluster GPUs (they are an sm_60), specifically in the flux computation. 

For the moment we can have this example on a lat long grid but my goal is to have the global ocean here.

simone-silvestri:
I think this is ready to merge

examples/near_global_ocean_simulation.jl
```diff
@@ -0,0 +1,299 @@
...
+
+# ### Ocean Model Configuration
+#
+# To configure the ocean simulation, we use the `ocean_simulation` function from ClimaOcean.jl. This function allows us to build
+# an ocean simulation with default parameters and numerics. The defaults include:
+# - CATKE turbulence closure for vertical mixing, see [`CATKEVerticalDiffusivity`](@ref)
+# - WENO-based advection scheme for momentum in the vector invariant form, see [`WENOVectorInvariant`](@ref)
+# - WENO-based advection scheme for tracers, see [`WENO`](@ref)
+# - `SplitExplicitFreeSurfaceSolver` with 75 substeps, see [`SplitExplicitFreeSurface`](@ref)
+# - TEOS-10 equation of state, see [`TEOS10EquationOfState`](@ref)
```
navidcy, line 89: the docs refs here (i.e. `(@ref)`) won't work because they mean to point to the docs of Oceananigans which is a different package.

examples/near_global_ocean_simulation.jl
```diff
@@ -67,10 +67,10 @@ grid = ImmersedBoundaryGrid(grid, GridFittedBottom(bottom_height))
...
 bathymetry[bathymetry .>= 0] .= NaN
 
-fig = Figure(size = (800, 400))
-ax = Axis(fig[1, 1], title = "Bathymetry [m]")
-hm = heatmap!(axis, bathymetry, colormap = :deep, colorrange = (-6000, 0))
-cb = Colorbar(fig[1, 2], hm)
+fig = Figure(size = (1200, 400))
+ax  = Axis(fig[1, 1])
+hm = heatmap!(ax, bathymetry, colormap = :deep, colorrange = (-6000, 0))
+cb = Colorbar(fig[0, 1], hm, label = "Bottom height [m]", vertical = false)
```
glwagner, line 73:
```suggestion
cb = Colorbar(fig[0, 1], hm, label = "Bottom height (m)", vertical = false)
```

We use this style for units throughout the docs here and in Oceananigans. We can change it though. I just think we should use the same style throughout.

glwagner (approved):

Issue #107, opened 2024-07-30
Add biogeochemistry interface to ocean_simulation


vtamsitt:
1. At minimum this requires adding a `biogeochemistry` kwarg to the `ocean_simulation` [function](https://github.com/CliMA/ClimaOcean.jl/blob/514135fc9a1016ec0c32081fe3914e789b3d35ca/src/OceanSimulations/OceanSimulations.jl#L55) (the BGC model can be initialised beforehand with the grid).

2. A way to pass BGC boundary conditions (e.g. surface air-sea CO2 flux) and merge with existing `ocean_boundary_conditions`, and ideally also a way to pass through custom BGC forcings 

3. Also I think the BGC tracers that are defined within the biogeochemistry model need to be added to [tracers](https://github.com/CliMA/ClimaOcean.jl/blob/514135fc9a1016ec0c32081fe3914e789b3d35ca/src/OceanSimulations/OceanSimulations.jl#L99C5-L99C12) for tracer_advection? Could perhaps extract the tracers initialised in the BGC model using `Oceananigans.Biogeochemistry.required_biogeochemical_tracers(biogeochemistry)` and merge with existing `tracers`?

cc @glwagner @jagoosw

glwagner:
Boundary conditions are hard coded here:

https://github.com/CliMA/ClimaOcean.jl/blob/514135fc9a1016ec0c32081fe3914e789b3d35ca/src/OceanSimulations/OceanSimulations.jl#L77-L80

As for tracers, there's a sort-of-ongoing discussion about whether we should add bgc tracers automatically or not. We could definitely do that for `ocean_simulation` but I think it'd be best to take the same approach that we take in Oceananigans (for example, we could also add them automatically in Oceananigans). Just to minimize confusion / dissonance between the user interfaces. Any thoughts about adding them automatically vs not? We had reasoned that requiring the tracers to be explicitly spelled out made scripts easier to understand. And while slightly more inconvenient to write, the fact is that users of bgc need to know what the tracers are called anyways for output, initial conditions, etc.

jagoosw:
> Boundary conditions are hard coded here:
> 
> https://github.com/CliMA/ClimaOcean.jl/blob/514135fc9a1016ec0c32081fe3914e789b3d35ca/src/OceanSimulations/OceanSimulations.jl#L77-L80
Couldn't `ocean_simulation` just have a kwarg e.g. `custom_boundary_conditions` and then a line like:
```julia
[...]

PR #123, opened 2024-08-09
Add biogeochemical interface to ocean_simulation

vtamsitt:
This PR addresses #107 by adding a `biogeochemistry` kwarg to `ocean_simulation`, adding user input `boundary_conditions` as a kwarg and moves `tracer` to be explicitly spelled out be the user as a kwarg.

Right now the user input `boundary_conditions` will be merged with hardcoded boundary conditions for T and S, unless the user supplies BCs for T and S in which case they will override the hardcoded ones. As @glwagner pointed out

> it would be nice if we only replace the boundary conditions we have to, ie we keep the surface fluxes of T, S even if we need to use user-supplied T or S conditions at the bottom or on immersed boundaries

but I haven't implemented that yet.

### DIFF
--- a/src/OceanSimulations/OceanSimulations.jl
+++ b/src/OceanSimulations/OceanSimulations.jl
@@ -106,7 +106,8 @@ function ocean_simulation(grid; Δt = 5minutes,
 
     if closure isa CATKEVerticalDiffusivity
         tracers = tuple(tracers..., :e)
-        tracer_advection = (; zip(tracers,tuple(fill(default_advection, length(tracers))...)))
+        advection_value = tuple(fill(tracer_advection, length(tracers))...)
+        tracer_advection = (; zip(tracers,advection_value)...)
         tracer_advection = merge(tracer_advection, (e = nothing,))
     end
 
--- a/src/OceanSimulations/OceanSimulations.jl
+++ b/src/OceanSimulations/OceanSimulations.jl
@@ -84,7 +84,9 @@ function ocean_simulation(grid; Δt = 5minutes,
                                  T = FieldBoundaryConditions(top = FluxBoundaryCondition(Jᵀ)),
                                  S = FieldBoundaryConditions(top = FluxBoundaryCondition(Jˢ)))
     
-    ocean_boundary_conditions = merge(ocean_boundary_conditions, boundary_conditions)    
+    if !isempty(boundary_conditions)
+        ocean_boundary_conditions = merge(ocean_boundary_conditions, boundary_conditions)    
+    end
 
     if grid isa ImmersedBoundaryGrid
         Fu = Forcing(u_immersed_bottom_drag, discrete_form=true, parameters=bottom_drag_coefficient)
--- a/src/OceanSimulations/OceanSimulations.jl
+++ b/src/OceanSimulations/OceanSimulations.jl
@@ -105,8 +105,8 @@ function ocean_simulation(grid; Δt = 5minutes,
 
 
     if closure isa CATKEVerticalDiffusivity
-        tracers = tuple(tracers..., :e) 
-        tracer_advection = NamedTuple{typeof(tracers)}((get(tracers, i, tracer_advection) for i in 1:length(tracers))...)
+        tracers = tuple(tracers..., :e)
+        tracer_advection = (; zip(tracers,tuple(fill(default_advection, length(tracers))...)))
         tracer_advection = merge(tracer_advection, (e = nothing,))
     end
 
--- a/src/OceanSimulations/OceanSimulations.jl
+++ b/src/OceanSimulations/OceanSimulations.jl
@@ -50,6 +50,8 @@ default_tracer_advection() = TracerAdvection(WENO(; order = 7),
 @inline u_immersed_bottom_drag(i, j, k, grid, clock, fields, μ) = @inbounds - μ * fields.u[i, j, k] * is_immersed_drag_u(i, j, k, grid) * spᶠᶜᶜ(i, j, k, grid, fields) / Δzᶠᶜᶜ(i, j, k, grid)
 @inline v_immersed_bottom_drag(i, j, k, grid, clock, fields, μ) = @inbounds - μ * fields.v[i, j, k] * is_immersed_drag_v(i, j, k, grid) * spᶜᶠᶜ(i, j, k, grid, fields) / Δzᶜᶠᶜ(i, j, k, grid)
 
+NoBiogeochemistry(args...; kwargs...) = nothing
+
 # TODO: Specify the grid to a grid on the sphere; otherwise we can provide a different
 # function that requires latitude and longitude etc for computing coriolis=FPlane...
 function ocean_simulation(grid; Δt = 5minutes,
@@ -63,6 +65,8 @@ function ocean_simulation(grid; Δt = 5minutes,
                           coriolis = HydrostaticSphericalCoriolis(; rotation_rate),
                           momentum_advection = default_momentum_advection(),
                           tracer_advection = default_tracer_advection(),
+                          biogeochemistry = NoBiogeochemistry,
+                          boundary_conditions = NamedTuple(),
                           verbose = false)
 
     # Set up boundary conditions using Field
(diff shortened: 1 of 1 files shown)


src/OceanSimulations/OceanSimulations.jl
```diff
@@ -96,10 +104,12 @@ function ocean_simulation(grid; Δt = 5minutes,
...
     end
 
-    tracers = (:T, :S)
+
     if closure isa CATKEVerticalDiffusivity
         tracers = tuple(tracers..., :e)
-        tracer_advection = (; T = tracer_advection, S = tracer_advection, e = nothing)
+        advection_value = tuple(fill(tracer_advection, length(tracers))...)
+        tracer_advection = (; zip(tracers,advection_value)...)
+        tracer_advection = merge(tracer_advection, (e = nothing,))
```
glwagner, line 112: Something along these lines might work too:

```julia
tracer_advection = NamedTuple(name => tracer_advection for name in tracers)
tracer_advection = merge(tracer_advection, (; e=nothing))
```

glwagner: Also should we have something that checks if the user has already provided a named tuple? something like

```julia
if !(tracer_advection isa NamedTuple)
    # do this
end
```

That way people can experiment with also advecting TKE which may be useful.

cc @simone-silvestri

glwagner: Dictionaries also can be used which may be a little easier to understand:

```julia
tracer_advection = Dict(name => tracer_advection for name in tracers)
tracer_advection[:e] = nothing
tracer_advection = NamedTuple(name => tracer_advection[name] for name in tracers)
```

src/OceanSimulations/OceanSimulations.jl
```diff
@@ -111,6 +121,7 @@ function ocean_simulation(grid; Δt = 5minutes,
                                               free_surface,
                                               coriolis,
                                               forcing,
+                                              biogeochemistry,
```
glwagner, line 124: 🚀

glwagner:
Looking good so far! Dictionaries can be useful for building up keyword arguments / parameters with values that have to be overwritten, since it's easy to overwrite entries in a dictionary.

src/OceanSimulations/OceanSimulations.jl
```diff
@@ -96,10 +104,12 @@ function ocean_simulation(grid; Δt = 5minutes,
         momentum_advection = nothing
     end
```
simone-silvestri, line 106: we probably need to make sure that T and S exist if we are passing tracers as an argument. Something like this might work
```suggestion
tracers = unique(tuple(tracers..., :T, :S))
```

glwagner: Personally I think it is more obvious to simply check whether `:T` and `S` are there. Plus lets all remember not to try to cram lots of code onto one line.

glwagner: The model constructor will show an error, right? Letting the model constructor error is consistent with the approach we are also taking for biogeochemical tracers.

The discussion is whether we should add tracers automatically, or require that users add tracers explicitly.

simone-silvestri: Well, at this moment it's a mix that will not work because in the boundary conditions it's implied that T and S exist. The approach we used is to add T and S (and e if using CATKE) automatically. The question is if we want to change that behavior or not. I would vote for no, but even if we do it, better to do that in another PR.

glwagner: The default is still `tracers = (:T, :S)`.

simone-silvestri: Of course, it is, but there is the possibility to pass tracers that do not include T and S, which is a fundamental change with respect to how we defined this function. 
There are pros and cons to that, but for the moment, `ocean_simulation` defines boundary conditions for T and S whether you pass T and S or not. It does not really make sense to do so if we allow not having T and S in an ocean simulation. So I would say, let's keep the intended behavior for now and we can discuss this in an issue.

src/OceanSimulations/OceanSimulations.jl
vtamsitt: Thanks @glwagner for the suggestions

src/OceanSimulations/OceanSimulations.jl
vtamsitt: Thanks, I wasn't really sure what was the best way to handle the T, S tracers. So for now I can add the additional line @simone-silvestri suggested to ensure that T and S exist.

> we probably need to make sure that T and S exist if we are passing tracers as an argument. Something like this might work

`tracers = unique(tuple(tracers..., :T, :S))`

I'll just add that from my user point of view, with a `tracer` kwarg I would assume that all tracers need to be defined including T and S, it would be a bit odd to not define those two unless the kwarg was `bgc_tracers` or something else.

simone-silvestri: right, we are trying to disentangle two objectives:
- being explicit about the arguments that go into the model
- Avoid awkward situations like having to explicitly declare T and S when adding a biogeochemistry model and not when the latter is not added. 
This is apparently not the correct way and we probably need to figure out a better solution that satisfies both these requirements. Let's discuss solutions to this problem in an issue.

glwagner: I disagree that it's awkward to explicitly declare T, S. What do others think?

vtamsitt:
@glwagner in the most recent [commit](https://github.com/CliMA/ClimaOcean.jl/pull/123/commits/b0591860ca47a1b62fa13276c85c7d27e2d279c5) I added a way to merge the user supplied `boundary_conditions` while keeping the hard coded T and S boundary conditions that aren't supplied by the user (ie we keep the surface fluxes of T, S even if we need to use user-supplied T or S conditions at the bottom or on immersed boundaries). It seems to work, but its ugly... any suggestions to do this more cleanly?

glwagner:
> @glwagner in the most recent [commit](https://github.com/CliMA/ClimaOcean.jl/pull/123/commits/b0591860ca47a1b62fa13276c85c7d27e2d279c5) I added a way to merge the user supplied `boundary_conditions` while keeping the hard coded T and S boundary conditions that aren't supplied by the user (ie we keep the surface fluxes of T, S even if we need to use user-supplied T or S conditions at the bottom or on immersed boundaries). It seems to work, but its ugly... any suggestions to do this more cleanly?

I think what you wrote captures the gist of it! But maybe we can make a few tweaks, I'll suggest something

glwagner:
Here's an idea that may or may not be helpful:

```julia
# create a utility
boundaries = (:west, :east, :south, :north, :bottom, :top, :immersed)
field_bcs_to_dict(field_bcs) = Dict(boundary => getproperty(field_bcs, boundary) for boundary in boundaries)

# inside ocean_simulation...

# make clear these are the defaults
default_boundary_conditions = Dict(
    :u => Dict(:top => FluxBoundaryCondition(Jᵘ), :bottom => u_bot_bc),
    :v => Dict(:top => FluxBoundaryCondition(Jᵛ), :bottom => v_bot_bc),
    :T => Dict(:top => FluxBoundaryCondition(Jᵀ)),
    :S => Dict(:top => FluxBoundaryCondition(Jˢ))
)

# Build a list of the default + user-specified boundary condition names
user_bc_names = keys(boundary_conditions)
bc_names = tuple(:u, :v, :T, :S, user_bc_names...) |> unique

# Merge user and default boundary conditions
merged_boundary_conditions = Dict()

for fieldname in bc_names

    default_bcs = get(default_boundary_conditions, fieldname, Dict())

    if fieldname ∈ user_bc_names
        user_bcs = field_bcs_to_dict(boundary_conditions[fieldname])
    else
        user_bcs = Dict()
    end

    # Generate a dictionary specifying `FieldBoundaryConditions` for each `fieldname`
    merged_boundary_conditions[fieldname] = merge(default_bcs, user_bcs)
end

# Convert to NamedTuple
merged_boundary_conditions = NamedTuple(merged_boundary_conditions)
```

A few things we could do (in a future PR) are:

1. Print some information if the user overwrites the default boundary conditions
2. Make sure `boundary_conditions` is a `NamedTuple` and also populated correctly with `FieldBoundaryConditions`

glwagner:
Not super pretty either...

glwagner:
Realized my code sketch was missing the conversion to `FieldBoundaryConditions`... could be missing some other things too...

vtamsitt:
I dropped the ball on this but picking it back up now and going to test it and hopefully it'll be ready for review this week!

PR #124, opened 2024-08-11
Change to Apache License 2.0

navidcy:
Hi ClimaOcean.jl contributors! We'd like to change the license from MIT to Apache 2.0 and to do that we need approval from all contributors. Could you confirm you are OK with this change?

cc @Sbozzolo, @jagoosw, @glwagner, @simone-silvestri, @vchuravy


navidcy:
Actually we decided to hear back advise on what's the proper procedure to do this change so I'll close this for now. Sorry

PR #130, opened 2024-08-13
Immersed boundary condition instead of forcing

### LINKED ISSUE #128: Why is bottom drag implemented as a forcing rather than `ImmersedBoundaryCondition`?
https://github.com/CliMA/ClimaOcean.jl/blob/9005d74a63668ca85863579091e23eaf956c3f91/src/OceanSimulations/OceanSimulations.jl#L83-L85

We should

1. Use `ImmersedBoundaryCondition`
2. If `ImmersedBoundaryCondition` does not work, fix it in Oceananigans and use it here.
3. If `ImmersedBoundaryCondition` cannot be fixed, discontinue support for it.

simone-silvestri:
To be merged **only** if the near-global example works with reasonable velocities

closes #128


glwagner:
> To be merged **only** if the near-global example works with reasonable velocities
> 
> closes #128

Why would it be different?

If its different there's a bug. We should also add a test to check for this this specifically in Oceananigans, since it has been discovered / is believed to be possible here.

simone-silvestri:
The results here should tell us if this method works. A test in Oceananigans should definitely be added because it looks like this is untested

glwagner:
Why do you think it is possible they are different? Have you seen this before? Is there another example where this occurred?

simone-silvestri:
I remember some time ago there was a problem with this method, but I would have to go back to look at what issue it was.
I guess it has been solved.  I am being cautious because now that we have a global simulation running we don't want to break it so it is always good to look 
1. the wall time step per time step
2. the validity of the of the outputs

glwagner:
> I remember some time ago there was a problem with this method, but I would have to go back to look at what issue it was. I guess it has been solved. I am being cautious because now that we have a global simulation running we don't want to break it so it is always good to look
> 
> 1. the wall time step per time step
> 2. the validity of the of the outputs

We shouldn't have to use such caution, because the code should not have so many bugs that things are so delicate. Can we figure out what the issue is? I don't think `ImmersedBoundaryCondition` has been developed lately. I'm not sure how the problem could have been solved.

simone-silvestri:
Maximum velocity seems to be equal between this PR and the documentation one

https://clima.github.io/ClimaOceanDocumentation/previews/PR104/literated/near_global_ocean_simulation/
vs
https://clima.github.io/ClimaOceanDocumentation/previews/PR130/literated/near_global_ocean_simulation/

glwagner:
This PR lists 29 files changed but the title just states that bottom drag was implemented with `ImmersedBoundaryCondition` vs `Forcing`. Why are so many files changed?

src/OceanSimulations/OceanSimulations.jl
```diff
@@ -74,17 +71,22 @@ function ocean_simulation(grid; Δt = 5minutes,
...
-    if grid isa ImmersedBoundaryGrid
-        Fu = Forcing(u_immersed_bottom_drag, discrete_form=true, parameters=bottom_drag_coefficient)
-        Fv = Forcing(v_immersed_bottom_drag, discrete_form=true, parameters=bottom_drag_coefficient)
-        forcing = merge(forcing, (; u = Fu, v = Fv))
-    end
+    u_immersed_drag = FluxBoundaryCondition(u_immersed_bottom_drag, discrete_form = true, parameters = bottom_drag_coefficient)
+    v_immersed_drag = FluxBoundaryCondition(v_immersed_bottom_drag, discrete_form = true, parameters = bottom_drag_coefficient)
     
+    u_immersed_bc = ImmersedBoundaryCondition(bottom = u_immersed_drag)
+    v_immersed_bc = ImmersedBoundaryCondition(bottom = v_immersed_drag)
```
glwagner, line 83: this is the only change I expected to see

simone-silvestri:
because we have to merge the documentation PR

glwagner:
There are 783 commits on this PR

glwagner:
Then put this PR to merge into that PR, not `main`

simone-silvestri:
oops too late, I ve merged the documentation PR

glwagner:
Still 250 commits here. Maybe open a new PR with just the changes needed here?

glwagner:
@navidcy any idea why the preview for this isn't showing? Or where it might be?

simone-silvestri:
Closed in favor of the new PR

glwagner:
delete the branch?

PR #140, opened 2024-08-14
Refactor bathymetry tools

### LINKED ISSUE #121: New name for `remove_lakes!`
I think `remove_lakes!` could use a better name:

https://github.com/CliMA/ClimaOcean.jl/blob/9005d74a63668ca85863579091e23eaf956c3f91/src/Bathymetry.jl#L279

A "lake" is not a precisely defined thing and is essentially a geographic term. Some inland bodies may be called "seas", and if a body of water is too small, it might be called a "pond". I think using the term `lake` will cause confusion.

Can we find a better term? Think about encountering this term within scripts that span applications from direct numerical simulation on centimeter scales to global scale simulations. What will work in all of those cases?

The kwarg refers to "connected regions". So, to match the kwarg, we may want `remove_unconnected_regions!`. Although that said, I don't completely understand what it means to allow "2 connected regions". Connected to what?

Another term beside "region" (and better than "lake" I think) is "basin" which may be more specific to bathymetry.

### LINKED ISSUE #133: Annoying warnings from `Bathymetry`
I think it's a little unclear what the user is supposed to do about this kind of warning:

```julia
┌ Warning: The westernmost meridian of `target_grid` 0 does not coincide with the closest meridian of the bathymetry grid, -1.4210854715202004e-14.
└ @ ClimaOcean.Bathymetry ~/Projects/ClimaOcean.jl/src/Bathymetry.jl:147
┌ Warning: The southernmost parallel of `target_grid` -81.73682203400921 does not coincide with the closest parallel of the bathymetry grid, -81.73333333333332.
└ @ ClimaOcean.Bathymetry ~/Projects/ClimaOcean.jl/src/Bathymetry.jl:151
```

Should we simply delete the warnings?

glwagner:
This PR cleans up the implementation of bathymetry tools, renaming functions and variables and clarifying whether functions modify in place or not.

Closes #121 
Closes #133

### DIFF
--- a/src/Bathymetry.jl
+++ b/src/Bathymetry.jl
@@ -205,14 +205,15 @@ function interpolate_bathymetry_in_passes(native_z, target_grid;
     Nλt, Nφt = Nt = size(target_grid)
     Nλn, Nφn = Nn = size(native_z)
 
-    # Check whether we are refining the grid in both directions.
-    # If not, skip interpolation passes.
-    if any(Nt[1:2] .> Nn[1:2])
+    # Check whether we are coarsening the grid in any directions.
+    # If so, skip interpolation passes.
+    if Nλt > Nλn || Nφt > Nφn
         target_z = Field{Center, Center, Nothing}(target_grid)
         interpolate!(target_z, native_z)
         @info string("Skipping passes for interplating bathymetry of size $Nn", '\n',
-                     "to target grid of size $Nt. Interpolation passes are effective only", '\n',
-                     "if bathymetry size is larger than target grid size in both horizontal directions.")
+                     "to target grid of size $Nt. Interpolation passes may only", '\n',
+                     "be used to refine bathymetryand requires that the bathymetry", '\n',
+                     "is larger than the target grid in both horizontal directions.")
         return target_z
     end
  
@@ -229,13 +230,14 @@ function interpolate_bathymetry_in_passes(native_z, target_grid;
     Nλ = Int[Nλ..., Nλt]
     Nφ = Int[Nφ..., Nφt]
 
-    old_z     = native_z
-    TX, TY, _ = topology(target_grid)
+    old_z  = native_z
+    TX, TY = topology(target_grid)
 
     for pass = 1:passes - 1
         new_size = (Nλ[pass], Nφ[pass], 1)
 
         @debug "Bathymetry interpolation pass $pass with size $new_size"
+
         new_grid = LatitudeLongitudeGrid(architecture(target_grid),
                                          size = new_size, 
                                          latitude = (latitude[1],  latitude[2]), 
@@ -297,25 +299,25 @@ function remove_minor_basins!(Z, keep_major_basins)
         label_elements[e] = e
     end
         
-    Ib = [] # major basins indexes
-    Iw = findfirst(x -> x == maximum(total_elements), total_elements)
-    push!(Ib, label_elements[Iw])
-    total_elements = filter(x -> x != total_elements[Iw], total_elements)
-    label_elements = filter(x -> x != label_elements[Iw], label_elements)
+    mm_basins = [] # major basins indexes
+    mm_water = findfirst(x -> x == maximum(total_elements), total_elements)
+    push!(mm_basins, label_elements[mm_water])
+    total_elements = filter(x -> x != total_elements[mm_water], total_elements)
+    label_elements = filter(x -> x != label_elements[mm_water], label_elements)
 
-    for _ in 1:keep_major_basins
+    for m = 1:keep_major_basins
         next_maximum = findfirst(x -> x == maximum(total_elements), total_elements)
-        push!(Ib, label_elements[next_maximum])
+        push!(mm_basins, label_elements[next_maximum])
         total_elements = filter(x -> x != total_elements[next_maximum], total_elements)
         label_elements = filter(x -> x != label_elements[next_maximum], label_elements)
     end
         
     labels = map(Float64, labels)
 
-    for i in 1:maximum(labels)
-        remove_basin = (&).(Tuple(i != idx for idx in Ib)...)
+    for ℓ = 1:maximum(labels)
+        remove_basin = all(ℓ == m for m in mm_basins)
         if remove_basin
-            labels[labels .== i] .= 1e10 # Fictitious super large number
+            labels[labels .== ℓ] .= 1e10 # large number
         end
     end
 
--- a/src/Bathymetry.jl
+++ b/src/Bathymetry.jl
@@ -72,10 +72,10 @@ Keyword Arguments:
                         If _refining_ the original grid, additional passes will not help and no intermediate
                         steps will be performed.
 
-- connected_regions_allowed: number of ``connected regions'' allowed in the bathymetry. Connected regions are fluid 
-                             regions that are fully encompassed by land (for example the ocean is one connected region).
-                             Default is `Inf`. If a value < `Inf` is specified, connected regions will be preserved in order
-                             of how many active cells they contain.
+- basins: Number of "independent basins", or fluid regions fully encompassed by land,
+          that are retained by `remove_basins!`. Basins are removed by order of size:
+          the smallest basins are removed first. `basins=1` will retain only the largest basin.
+          Default: `Inf`.
 """
 function regrid_bathymetry(target_grid;
                            height_above_water = nothing,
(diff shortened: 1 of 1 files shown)


src/Bathymetry.jl
```diff
@@ -172,15 +154,23 @@ function regrid_bathymetry(target_grid;
...
     Nxn = length(λ_data)
     Nyn = length(φ_data)
     Nzn = 1
 
     native_grid = LatitudeLongitudeGrid(arch;
                                         size = (Nxn, Nyn, Nzn),
-                                        latitude = (φ₁, φ₂),
-                                        longitude = (λ₁, λ₂),
+                                        latitude = (φ₁_data, φ₂_data),
+                                        longitude = (λ₁_data, λ₂_data),
```
glwagner, line 173: @simone-silvestri please inform whether this is correct.

src/Bathymetry.jl
```diff
@@ -189,26 +179,44 @@ function regrid_bathymetry(target_grid;
...
     return target_z
 end
 
 # Here we can either use `regrid!` (three dimensional version) or `interpolate`
 function interpolate_bathymetry_in_passes(native_z, target_grid; 
                                           passes = 10,
-                                          connected_regions_allowed = 3,
+                                          major_basins = Inf,
                                           minimum_depth = 0)
     Nλt, Nφt = Nt = size(target_grid)
```
glwagner, line 205: I thought that `size` would always return a 3-tuple. How does this work?

simone-silvestri: `Nλt, Nφt` are the first two arguments of `size`. If we want to make it more explicit, `Nλt, Nφt, _` is ok too

src/Bathymetry.jl
```diff
@@ -115,10 +115,10 @@ function regrid_bathymetry(target_grid;
     λ_data = dataset["lon"][:]
     z_data = convert(Array{FT}, dataset["z"][:, :])
 
-    # Convert longitude to 0 - 360?
+    # Convert longitude from (-180, 180) to (0, 360)
     λ_data .+= 180
-    nhx      = size(z_data, 1)
-    z_data   = circshift(z_data, (nhx ÷ 2, 1))
+    Nhx    = size(z_data, 1)
+    z_data = circshift(z_data, (Nhx ÷ 2, 1))
```
simone-silvestri, line 121: I think this should be 
```suggestion
    z_data = circshift(z_data, (Nhx ÷ 2, 0))
```
because the second argument of the circshift are the offsets

src/Bathymetry.jl
simone-silvestri: should be correct

src/Bathymetry.jl
```diff
@@ -244,97 +252,81 @@ function interpolate_bathymetry_in_passes(native_z, target_grid;
...
-    total_elements = filter((x) -> x != total_elements[ocean_idx], total_elements)
-    label_elements = filter((x) -> x != label_elements[ocean_idx], label_elements)
+    Ib = [] # major basins indexes
+    Iw = findfirst(x -> x == maximum(total_elements), total_elements)
+    push!(Ib, label_elements[Iw])
+    total_elements = filter(x -> x != total_elements[Iw], total_elements)
+    label_elements = filter(x -> x != label_elements[Iw], label_elements)
 
-    for _ in 1:connected_regions_allowed
+    for _ in 1:keep_major_basins
```
simone-silvestri, line 306: do we want to have `keep_major_basins` include the largest basin by default or not? That is the difference between having a global ocean using
```
keep_major_basins = 1
```
or 
```
keep_major_basins = 0
```

src/Bathymetry.jl
```diff
@@ -189,26 +179,44 @@ function regrid_bathymetry(target_grid;
...
+    fill_halo_regions!(target_z)
+
     return target_z
 end
 
 # Here we can either use `regrid!` (three dimensional version) or `interpolate`
 function interpolate_bathymetry_in_passes(native_z, target_grid; 
                                           passes = 10,
-                                          connected_regions_allowed = 3,
+                                          major_basins = Inf,
```
simone-silvestri, line 203: we can remove this one right?
```suggestion
```

src/Bathymetry.jl
```diff
@@ -189,26 +179,44 @@ function regrid_bathymetry(target_grid;
...
                                                 passes = interpolation_passes,
-                                                connected_regions_allowed,
+                                                major_basins,
                                                 minimum_depth)
 
+    if minimum_depth > 0
+        zi = interior(target_z, :, :, 1)
+        shallow_ocean = zi .> - minimum_depth
+        zi[shallow_ocean] .= 0
+    end
```
simone-silvestri, line 189: can we do this on the GPU? Or do we need a kernel?

src/Bathymetry.jl
glwagner: I think `1` makes more sense, otherwise its hard to explain the meaning of the variable. And we should check that its not 0 probably.

src/Bathymetry.jl
glwagner: good catch

src/Bathymetry.jl
glwagner: I would have thought it works on the GPU but I will check and write a kernel if not

glwagner: it works

```
julia> a = CuArray(rand(4, 4))
4×4 CuArray{Float64, 2, CUDA.Mem.DeviceBuffer}:
 0.968109  0.435424  0.988271  0.108635
 0.102944  0.956011  0.637797  0.140668
 0.466575  0.800952  0.397086  0.812225
 0.221806  0.308943  0.385949  0.471147

julia> small = a .< 0.5
4×4 CuArray{Bool, 2, CUDA.Mem.DeviceBuffer}:
 0  1  0  1
 1  0  0  1
 1  0  1  0
 1  1  1  1

julia> a[small] .= 0
10-element view(::CuArray{Float64, 2, CUDA.Mem.DeviceBuffer}, CartesianIndex{2}[CartesianIndex(2, 1), CartesianIndex(3, 1), CartesianIndex(4, 1), CartesianIndex(1, 2), CartesianIndex(4, 2), CartesianIndex(3, 3), CartesianIndex(4, 3), CartesianIndex(1, 4), CartesianIndex(2, 4), CartesianIndex(4, 4)]) with eltype Float64:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
```

simone-silvestri: nice

glwagner: this also work son the GPU for in-place

```
julia> map!(x -> x * (x < 0.5), a, a)
4×4 CuArray{Float64, 2, CUDA.Mem.DeviceBuffer}:
 0.0       0.403628  0.0       0.00580952
 0.0       0.393159  0.335011  0.263924
 0.481604  0.0       0.0       0.0
 0.124625  0.0       0.452182  0.399641
```

glwagner:
is any of this tested @simone-silvestri ? Can we add some tests in this PR?

src/Bathymetry.jl
glwagner: actually this is better 

```julia
julia> a .*= a .> b
4×4 CuArray{Float64, 2, CUDA.Mem.DeviceBuffer}:
 0.775472  0.0       0.0  0.0
 0.0       0.838647  0.0  0.0
 0.502095  0.0       0.0  0.0
 0.0       0.74245   0.0  0.639991
```

simone-silvestri:
This is not tested. We can add a test where we verify that interpolating on the same grid does nothing.

glwagner:
Ok, a lot of tests are needed to bring this up to the same standards that we currently use for Oceananigans:

1. Unit tests that ensure the functions run without an error, also testing conditions like:
    a. `remove_basins!(z, Inf)` does nothing
    b. `remove_basins(z, 2.1)` throws an error (for example --- other inputs might be tested)
    c. `interpolate_bathymetry_in_passes` doesn't do anything when we are not refining the bathymetry
    d. The argument `minimum_depth` works
2. Test with synthetic bathymetry to ensure that `remove_basins!` correctly removes the specified number of basins
3. Test that interpolation for multiple passes works correctly, either using a synthetic example or using a regression test

Those come to the top of my head.

glwagner:
Note that in Oceanagnians we are often stingy / conservative about adding tests because we have so many and often testing can be achieved by adding an example or slightly refactoring existing tests. We need a more aggressive approach to testing with ClimaOcean because there are no tests right now.

glwagner:
@simone-silvestri you reviewed the PR but did not approve and I don't see a path towards getting approval.

PR #143, opened 2024-08-16
A couple of bugixes for the bathymetry module

simone-silvestri:
and adapting the example to the new syntax

### DIFF
--- a/src/Bathymetry.jl
+++ b/src/Bathymetry.jl
@@ -158,10 +158,10 @@ function regrid_bathymetry(target_grid;
     Δλ = λ_data[2] - λ_data[1]
     Δφ = φ_data[2] - φ_data[1]
 
-    λ₁_data = λ_data[i₁] - Δλ / 2
-    λ₂_data = λ_data[i₂] + Δλ / 2
-    φ₁_data = φ_data[j₁] - Δφ / 2
-    φ₂_data = φ_data[j₂] + Δφ / 2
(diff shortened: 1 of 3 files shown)


src/Bathymetry.jl
```diff
@@ -287,6 +286,10 @@ function remove_minor_basins!(Z, keep_major_basins)
         throw(ArgumentError("`keep_major_basins` must be a finite number!"))
     end
 
+    if keep_major_basins < 1
+        @warn "keep_major_basins is smaller than 1. This will lead to a completely immersed domain."
```
glwagner, line 290: I think you should throw an error.

glwagner, line 290:
```suggestion
        throw(ArgumentError("keep_major_basins must be greater than 1."))
```

glwagner:
tests?

src/Bathymetry.jl
```diff
@@ -287,7 +287,7 @@ function remove_minor_basins!(Z, keep_major_basins)
     end
 
     if keep_major_basins < 1
-        @warn "keep_major_basins is smaller than 1. This will lead to a completely immersed domain."
+        throw(ArgumentError("keep_major_basins must be greater than 1."))
```
glwagner:
```suggestion
        throw(ArgumentError("keep_major_basins must be greater than 0."))
```

sorry...

glwagner (approved):

glwagner:
We really need tests.

Issue #141, opened 2024-08-14
Tests for bathymetry utilities


glwagner:
There are few tests for bathymetry utilities. Here are a few that we need:

1. Unit tests that ensure the functions run without an error, also testing conditions like:
    a. `remove_basins!(z, Inf)` does nothing
    b. `remove_basins(z, 2.1)` throws an error (for example --- other inputs might be tested)
    c. `interpolate_bathymetry_in_passes` doesn't do anything when we are not refining the bathymetry
    d. The argument `minimum_depth` works
2. Test with synthetic bathymetry to ensure that `remove_basins!` correctly removes the specified number of basins
3. Test that interpolation for multiple passes works correctly, either using a synthetic example or using a regression test

Those come to the top of my head.

_Originally posted by @glwagner in https://github.com/CliMA/ClimaOcean.jl/issues/140#issuecomment-2290024722_

glwagner:
cc @simone-silvestri

francispoulin:
I learned yesterday that in regards to bathymetry, there is something called in-paint.  Does this need a test?

glwagner:
It does seem like we need tests!

glwagner:
> That is the `remove_basins!` feature that removes all connected regions but the largest. BTW there is an issue with the new bathymetry. The `generate_bathymetry` example fails on main with https://buildkite.com/clima/climaocean-examples/builds/240#01915522-a196-4bb4-91ab-66198dbc76ed

#143 claimed to fix this particularly issue and was merged, but we still don't have tests so cannot close this issue.

PR #147, opened 2024-08-19
Add GPU tests, plus tests for bathymetry module

simone-silvestri:
follows the detailes outlined in #141


.buildkite/pipeline.yml
```diff
@@ -54,5 +54,13 @@ steps:
     commands:
       - "julia --project -e 'using Pkg; Pkg.test()'"
 
+  - label: "Run bathyemtry tests"
+    key: "tests_bathymetry"
+    env:
+      CUDA_VISIBLE_DEVICES: "-1"
```
glwagner, line 60: Probably we should test on the GPU right?

glwagner (approved):

test/test_bathymetry.jl
```diff
@@ -0,0 +1,57 @@
...
+                                     latitude = (0, 50),
+                                     z = (-6000, 0))
+
+        # Test that remove_minor_basins!(Z, Inf) does nothing
+        bottom_height = regrid_bathymetry(grid)        
+        control_bottom_height = deepcopy(bottom_height)
+
+        remove_minor_basins!(bottom_height, Inf)
+
+        @test interior(bottom_height) .== interior(control_bottom_height)
```
glwagner, line 20:
```suggestion
        bottom_height = regrid_bathymetry(grid)        
        control_bottom_height = deepcopy(bottom_height)
        remove_minor_basins!(bottom_height, Inf)
        @test interior(bottom_height) .== interior(control_bottom_height)
```

test/test_bathymetry.jl
```diff
@@ -0,0 +1,57 @@
...
+        # A two basins bathymetry
+        bottom(x, y) = - 1000 * Int((x < 10) | (x > 45)) 
+        
+        set!(bottom_height, bottom)
+        set!(control_bottom_height, bottom)
+
+        # This should have not changed anything
+        remove_minor_basins!(bottom_height, 2)
+
+        @test interior(bottom_height) .== interior(control_bottom_height)
```
glwagner, line 35:
```suggestion
        remove_minor_basins!(bottom_height, 2)
        @test interior(bottom_height) .== interior(control_bottom_height)
```

test/test_bathymetry.jl
```diff
@@ -0,0 +1,57 @@
...
+                                     latitude = (-10, 50),
+                                     z = (-6000, 0))
+
+        control_bottom_height = regrid_bathymetry(grid)
+        interpolated_bottom_height = regrid_bathymetry(grid; interpolation_passes = 100)
+
+        # Testing that multiple passes do not change the solution when refining the grid
+        @test interior(control_bottom_height) .== interior(interpolated_bottom_height)
+    end
+end
```
glwagner, line 54:
```suggestion

end 
```

simone-silvestri:
I have added the gpu tests using an environmental variable. I remember we decided to go in that direction, but I forgot which issue we were discussing about GPU tests. I can change it following a different strategy because I cannot find out which was the issue.

.buildkite/pipeline.yml
```diff
@@ -22,45 +22,114 @@ steps:
...
       TEST_GROUP: "fluxes"
     commands:
       - "julia --project -e 'using Pkg; Pkg.test()'"
+    agents:
+      slurm_mem: 32G
+      slurm_cpus_per_task: 8
+      slurm_ntasks: 1
 
-  - label: "Run bathyemtry tests"
+  - label: "CPU bathymetry tests"
```
glwagner, line 65: Is it best to use a separate job for CPU vs GPU tests, or just do both CPU and GPU tests in one job?

.buildkite/pipeline.yml
```diff
@@ -77,57 +77,57 @@ steps:
     key: "gpu_jra55_tests"
     env:
       TEST_GROUP: "jra55"
+      GPU_TEST: "true"
```
glwagner, line 80: A bit more flexible would be something like

```suggestion
      OCEANANIGANS_ARCHITECTURE: "GPU"
```

I suppose "false" is meant to signify CPU. But explicitly denoting the architecture might allow additional possibilities.

test/test_bathymetry.jl
```diff
@@ -18,7 +18,7 @@ using ClimaOcean.Bathymetry: remove_minor_basins!
 
         # A fictitiously large number which should presumably keep all the basins
         remove_minor_basins!(bottom_height, 10000000) 
-        @test interior(bottom_height) .== interior(control_bottom_height)
+        @test all(interior(bottom_height) .== interior(control_bottom_height))
```
glwagner, line 21:
```suggestion
        @test parent(bottom_height) == parent(control_bottom_height)
```

does this work?

glwagner:
CUDA initialization doesn't seem to be working, since https://buildkite.com/clima/climaocean-ci/builds/1215#01916d39-cf52-4943-a59e-d63e89f92b49/206-472

PR #151, opened 2024-08-19
Generalize "runoff" to any "auxiliary" freshwater flux for clarity

glwagner:
This PR attempts to clarify how the code reads by exchanging the term "runoff" for "auxiliary_freshwater_flux". The point is that what we call "runoff" is not really specific to runoff at all (and besides, it also applies to icebergs which is not "runoff"). 

Rather the fluxes that are now called "runoff" actually allows users to implement prescribed atmospheres that have two kinds of freshwater flux --- one on the "main" grid common to other atmospheric variables, and a second one on a "different" grid (also with different times).


simone-silvestri (approved):
adding a test would be nice

glwagner:
We can but note that this PR does not add functionality

glwagner:
@simone-silvestri in painting seems to take a long time:

https://buildkite.com/clima/climaocean-ci/builds/1237#01917130-a246-49fd-8850-ac2df60453f1/202-597

any idea why? Can we use a smaller test for this?

simone-silvestri:
We can also inpaint the field partially. 
There is this keyword argument `maxiter` for `inpaint_mask!` that we can propagate from outside. This should make the tests lighter.

glwagner:
> We can also inpaint the field partially. There is this keyword argument `maxiter` for `inpaint_mask!` that we can propagate from outside. This should make the tests lighter.

Why don't we inpaint something that is, say, 10x10x10?

PR #152, opened 2024-08-19
Remove CUDA as dependency

glwagner:
Adds `GPUArrays` as a test dependency so we can use `@allowscalar` in the tests.


simone-silvestri (approved):

glwagner:
Meh, let's close this for now

PR #153, opened 2024-08-20
Try a new way to initialize the CUDA runtime

glwagner:


glwagner:
This looks like its going to work @simone-silvestri

simone-silvestri:
Nice. Is this a problem with the new CUDA?

glwagner:
> Nice. Is this a problem with the new CUDA?

You mean, is the problem that this PR fixes a problem with the new CUDA? I don't think so. If you look at your PR, you'll also find that the CUDA runtime is not initialized when it is supposed to, and instead gets initialized in a later build.

I also noticed that the tests were failing intermittently, which I can't explain but at least these pass.

Note that we shouldn't have to depend on CUDA in ClimaOcean, but I wasn't able to make things work after getting rid of it.

glwagner:
Can I get a review?

PR #162, opened 2024-08-27
Refactor turbulent flux solver to have a `DiagnosticSurfaceTemperature` mode

glwagner:
This PR refactors the turbulent flux solver to have a diagnostic surface temperature mode.

Also I noticed that some ECCO files were floating outside of their submodule (looking at you @simone-silvestri ...)


glwagner:
Superceded by #278

PR #171, opened 2024-08-28
Just one `timeout_in_minutes`

### LINKED ISSUE #144: Why is `timeout_in_minutes` written twice?
It appears twice in `examples_build.yml`:

https://github.com/CliMA/ClimaOcean.jl/blob/ad28ba8acca2ecff7ab94f25ffb0a72e0ae3429a/.buildkite/examples_build.yml#L7

https://github.com/CliMA/ClimaOcean.jl/blob/ad28ba8acca2ecff7ab94f25ffb0a72e0ae3429a/.buildkite/examples_build.yml#L54

@simone-silvestri

glwagner:
Closes #144


navidcy (approved):
lgtm

but perhaps @simone-silvestri could confirm if that was there for a reason? I'm not that confident in yml file structure..

simone-silvestri:
I think one is enough. Let's see if the examples build.

simone-silvestri (approved):

PR #172, opened 2024-08-29
Putting together a near-global prototype OMIP setup

glwagner:
And misc changes that go along with it.


src/OceanSeaIceModels/time_step_ocean_sea_ice_model.jl
```diff
@@ -28,28 +28,26 @@ function time_step!(coupled_model::OceanSeaIceModel, Δt; callbacks=[], compute_
...
     #   accurate flux computation?
     tick!(coupled_model.clock, Δt)
-    
     update_state!(coupled_model, callbacks; compute_tendencies)
     
     return nothing
 end
 
-function update_state!(coupled_model::OceanSeaIceModel, callbacks=[]; compute_tendencies=false)
+function update_state!(coupled_model::OceanSeaIceModel, callbacks=[]; compute_tendencies=true)
```
simone-silvestri, line 46: do we need compute_tendencies here?

simone-silvestri (approved):

src/OceanSeaIceModels/time_step_ocean_sea_ice_model.jl
glwagner: I'll test that, we need if it is used in `Simulations` in `Oceananigans`

glwagner:
I'll merge this cause there's some useful stuff.

PR #180, opened 2024-09-17
Add different download directories for JRA55 and Bathymetry

simone-silvestri:
I have noticed that JRA55 and Bathymetry tests might struggle because they are split into CPU and GPU tests. Sometimes, one of the tests downloads the files while the other wants to read them before the downloading process is complete, and the tests fail with a NetCDF (101) error. By splitting the download directory into two different paths for CPU and GPU, this problem can be solved.

This PR also introduces a download directory for the ECCO module directly in the `ECCOMetadata` type. This is a first step to generalize the `Metadata` type for all the data we need to download


src/DataWrangling/ecco_metadata.jl
```diff
@@ -16,6 +22,7 @@ struct ECCOMetadata{D, V}
     name  :: Symbol
     dates :: D
     version :: V
+    dir :: String
```
glwagner, line 25: what? why?

glwagner:
The PR isn't well motivated.

The problem described has to do with improper initialization of the tests. During test initialization we apparently should download all relevant data. This can be done by adding to the `init` test group here;

https://github.com/CliMA/ClimaOcean.jl/blob/ddea9398cbcb729b194dec775b0329f7dca2102a/test/runtests.jl#L7-L10

I'm not sure about the other changes regarding download directories. Maybe are positive. But making these changes for _tests_ is clearly wrong.

glwagner:
This PR won't fix the issue that will arise eventually (hopefully) when more than one test group has to use the downloaded bathymetry data.

simone-silvestri:
Ok i wll try downloading data in the init step

src/DataWrangling/ecco_metadata.jl
simone-silvestri: I could also revert it back, but it is nice to know the path of the directory where the data resides, compactly in the metadata information

simone-silvestri:
This PR should be ready to go. We have to remember to add the data we download in the tests in the init when adding new tests.

src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -1,11 +1,17 @@
...
 
 using Base: @propagate_inbounds
 
 import Oceananigans.Fields: set!
 import Base
 
+download_ECCO_cache::String = ""
+function __init__()
+    global download_ECCO_cache = @get_scratch!("ECCO")
+end
```
glwagner, line 13: Should these lines be in the top level file `ECCO.jl`?

glwagner (approved):
I think __init__ for any module should always go in the top-level file. It's going to be hard to find otherwise. Plus we want the top-level file to give an overview of the module.

gaelforget:
Was trying to run `examples/near_global_ocean_simulation.jl` but am getting the error below. Would this branch help?

ps. NOAA servers struggled during past weeks after Hurricane Helene, which may explain the 503 error

```
[ Info:  ... downloaded 412.0 B (100% complete, 33.845 seconds)
--2024-10-12 13:54:38--  https://www.ngdc.noaa.gov/thredds/fileServer/global/ETOPO2022/60s/60s_surface_elev_netcdf/ETOPO_2022_v1_60s_N90W180_surface.nc
Resolving www.ngdc.noaa.gov (www.ngdc.noaa.gov)... 140.172.190.1
Connecting to www.ngdc.noaa.gov (www.ngdc.noaa.gov)|140.172.190.1|:443... connected.
HTTP request sent, awaiting response... 503 Service Unavailable
```

glwagner:
> Was trying to run `examples/near_global_ocean_simulation.jl` but am getting the error below. Would this branch help?
> 
> ps. NOAA servers struggled during past weeks after Hurricane Helene, which may explain the 503 error
> 
> ```
> [ Info:  ... downloaded 412.0 B (100% complete, 33.845 seconds)
> --2024-10-12 13:54:38--  https://www.ngdc.noaa.gov/thredds/fileServer/global/ETOPO2022/60s/60s_surface_elev_netcdf/ETOPO_2022_v1_60s_N90W180_surface.nc
> Resolving www.ngdc.noaa.gov (www.ngdc.noaa.gov)... 140.172.190.1
> Connecting to www.ngdc.noaa.gov (www.ngdc.noaa.gov)|140.172.190.1|:443... connected.
> HTTP request sent, awaiting response... 503 Service Unavailable
> ```

Huh, this looks like an issue downloading the etopo data (which happens in `regrid_bathymetry`:

https://github.com/CliMA/ClimaOcean.jl/blob/fc175f6c6b3e095bd460d10d323e4361006a5008/examples/near_global_ocean_simulation.jl#L55-L58

@gaelforget can you check to see if it is still broken?

Another solution is you can provide a `filename` and `dir` to `regrid_bathymetry` which should avoid the download. I am not sure this is tested yet.

PR #183, opened 2024-09-18
Add a download test on windows

### LINKED ISSUE #179: Difficuty accessing ECCO file
On trying the example generate_surface_fluxes.jl at the line

mask=ecco_mask()
I get the error (in this minimal example)
run(`wget --http-user=kevin_broughan --http-passwd=Nasa44earth-data 'https://ecco.jpl.nasa.gov/drive/files/ECCO2/cube92_latlon_quart_90S90N/daily/THETA\THETA.1440x720x50.19930101.nc'`)
ERROR: IOError: could not spawn `wget --http-user=kevin_broughan --http-passwd=Nasa44earth-data 'https://ecco.jpl.nasa.gov/drive/files/ECCO2/cube92_latlon_quart_90S90N/daily/THETA\THETA.1440x720x50.19930101.nc'`: no such file or directory (ENOENT)
However, when I login to Nasa earth data and navigate to the file the file its in place at 74.5Mb.  

I tried running Julia as su with no change in other problematic examples.

Thanks for looking at this -- Kevin

simone-silvestri:
should close #179. I have to find a way to use Downloads.jl with username and password. Probably HTTP.jl works better?

### DIFF
--- a/src/DataWrangling/ecco_metadata.jl
+++ b/src/DataWrangling/ecco_metadata.jl
@@ -164,6 +164,8 @@ function download_dataset!(metadata::ECCOMetadata;
                 fileurl = joinpath(url, shortname, year, filename)
             end
 
+            # Here add a method to download the file on windows where we can specify a password and a username
+
             cmd = `wget --http-user=$(username) --http-passwd=$(password) $(fileurl)`
         
             run(cmd)
--- a/test/test_downloading.jl
+++ b/test/test_downloading.jl
@@ -6,3 +6,11 @@ include("runtests_setup.jl")
         fts = ClimaOcean.JRA55.JRA55_field_time_series(name; time_indices=2:3)
     end
 end
+
+@testset "Availability of ECCO data" begin
+    @info "Testing that we can download ECCO data..."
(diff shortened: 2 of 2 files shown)


simone-silvestri:
There is probably a way by using a file that stores login info https://discourse.julialang.org/t/download-a-url-with-a-password/93669/10

glwagner:
Hmm true, that is interesting. The downside of the `.netrc` file is that it is less flexible than environment variables. 

If `.netrc` is the only method that works, we can start with that. We might be able to make things easier in the future if we need to (eg auto-writing `.netrc` files or something).

glwagner:
Description of `.netrc` files: https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html

glwagner:
Thinking about it more I feel the `.netrc` file makes a lot of sense. It's also nice to be able to copy a dedicated file easily between machines (imagine a file with tons of users/pwds) rather than copy/pasting bits in a `.bashrc` file or something. So `.netrc` seems better.

We can write a utility that writes to `.netrc` files if we need some programmatic convenience at some point.

PR #185, opened 2024-09-20
New interface for ECCORestoring

### LINKED ISSUE #177: `ECCORestoring` can be used directly as a forcing
Right now, `ECCORestoring` is abstract as the forcing "function", which must then be further wrapped in `Forcing` by `ECCO_restoring_forcing`:

https://github.com/CliMA/ClimaOcean.jl/blob/b3b44b709c70b296adb63cf7b2f4552334dcfd1b/src/DataWrangling/ecco_restoring.jl#L276-L279

There's no barrier to using `ECCORestoring` directly as a forcing though, provided the appropriate functionality within Oceananigans is extended correctly. And this would be much better, not just for syntax but also because the implementation will be less complicated and easier to understand.

simone-silvestri:
should close #177

This PR also includes a `maxiter` field to the ECCONetCDFbackend that controls the maximum number of iterations for the inpainting algorithm


src/DataWrangling/ECCO/ECCO_mask.jl
```diff
@@ -40,3 +41,48 @@ end
...
+struct LatitudinallyTaperedPolarMask{N, S, Z} 
+    northern_edges :: N
+    southern_edges :: S
+    z_edges :: Z
+end
+
+"""
+    LatitudinallyTaperedPolarMask(; northern_edges = (70,   75),
+                                    southern_edges = (-75, -70),
+                                    z_edges = (-20, 0))
```
glwagner, line 53: Nice! Is the proposed syntax change intentional? It wasn't mentioned in the description...

src/DataWrangling/ECCO/ECCO_restoring.jl
```diff
@@ -20,43 +21,41 @@ import Oceananigans.OutputReaders: new_backend, update_field_time_series!
 struct ECCONetCDFBackend{N} <: AbstractInMemoryBackend{Int}
     start :: Int
     length :: Int
+    maxiter :: Int
```
glwagner, line 24: should we make it more clear this refers to inpainting?

src/DataWrangling/ECCO/ECCO_restoring.jl
```diff
@@ -173,22 +176,22 @@ A struct representing ECCO restoring.
 - `ECCO_grid`: The native ECCO grid to interpolate from.
 - `mask`: A mask (could be a number, an array, a function or a field).
 - `variable_name`: The variable name of the variable that needs restoring.
-- `λ⁻¹`: The reciprocal of the restoring timescale.
+- `rate`: The reciprocal of the restoring timescale.
 """
-struct ECCORestoring{FTS, G, M, V, N} <: Function
+struct ECCORestoring{FTS, G, M, V, N} 
     ECCO_fts      :: FTS
     ECCO_grid     :: G
```
glwagner, line 183:
```suggestion
    field_time_series      :: FTS
    grid     :: G
```

src/DataWrangling/ECCO/ECCO_restoring.jl
```diff
@@ -231,52 +234,65 @@ end
...
-function ECCO_restoring_forcing(metadata::ECCOMetadata;
-                                architecture = CPU(), 
-                                time_indices_in_memory = 2, # Not more than this if we want to use GPU!
-                                time_indexing = Cyclical(),
-                                mask = 1,
-                                timescale = 20days,
-                                grid = nothing)
+function ECCORestoring(metadata::ECCOMetadata;
+                       architecture = CPU(), 
+                       time_indices_in_memory = 2, # Not more than this if we want to use GPU!
```
glwagner, line 281: should this be `backend`?

src/DataWrangling/ECCO/ECCO_mask.jl
simone-silvestri: I am actually ok also with `LinearlyTaperedPolarMask`. I just thought `Latitudinally` would express the fact that longitude doesn't play a role with the mask. However, I am happy to switch to `LinearlyTaperedPolarMask`

glwagner: yes and also adding the clarification `edges` to every argument. Somehow I think we want to make it so we don't need that clarification, perhaps by making the name of the type better. Notice that every property says "edges", suggesting that there should be a better way

test/test_ecco.jl
```diff
@@ -38,14 +38,67 @@ using Dates
...
+
+    mask = LatitudinallyTaperedPolarMask(northern_edges = (φ₃, φ₄), 
+                                         southern_edges = (φ₁, φ₂), 
+                                         z_edges = (z₁, 0))
+
+    t_restoring = ECCORestoring(:temperature, CPU(); 
+                                dates, 
+                                mask, 
+                                rate = 1 / 1000.0,
+                                maxiter = 1)
```
glwagner, line 65: should the max iter be set to the order of the advection scheme? or something like that?

src/DataWrangling/ECCO/ECCO_restoring.jl
simone-silvestri: At the moment, the only supported backend is the `ECCOBackendNetCDF`. If we want to be able to preprocess data we can change this to accept a backend and support different backends.

src/DataWrangling/ECCO/ECCO_restoring.jl
simone-silvestri: I have changed it to `inpainting_maxiter`. I am not super happy about it though. Do you have a suggestion for a name for this variable?

test/test_ecco.jl
simone-silvestri: I am not sure. I don't think it has to do with the model detail, it has more to do with the discrepancy of the bathymetry we use with the ECCO bathymetry

src/DataWrangling/ECCO/ECCO_restoring.jl
glwagner: How about making a new struct, `Inpainting`, which has a property `maxiter` (and which may be generalized to include more settings in the future)

simone-silvestri:
should be mergeable

glwagner (approved):

test/runtests.jl
```diff
@@ -46,7 +49,7 @@ if test_group == :turbulent_fluxes || test_group == :all
 end
 
 if test_group == :bathymetry || test_group == :all
-    include("test_bathymetry.jl")
+    CUDA.set_runtime_version!(v"12.6"; local_toolkit = true)
```
glwagner: you may want to keep the test file though

PR #191, opened 2024-10-09
Don't specify arch in github `ci.yml`

glwagner:
I'm not sure I completely understand the logic of the tests (do we really have to test two Windows architectures? Do we have a good reason to think the code we write might work on one architecture but not the other?)

Hopefully this is an ok change and also addresses the @Sbozzolo found in specifying `x64` for macOS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,9 +20,11 @@ jobs:
           - os: ubuntu-latest
             arch: x64
             version: '1.10'
+            run-name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
           - os: ubuntu-latest
             arch: x86
             version: '1.10'
+            run-name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
     steps:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -15,26 +15,13 @@ jobs:
           - '1.10'
         os:
           - ubuntu-latest
-        arch:
-          - x64
-          - x86
-        include:
-          # test macOS and Windows with latest Julia only
-          - os: macOS-latest
-            arch: x64
-            version: '1.10'
-          - os: windows-latest
-            arch: x64
-            version: '1.10'
-          - os: windows-latest
-            arch: x86
-            version: '1.10'
+          - macOS-latest
+          - windows-latest
     steps:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@latest
         with:
           version: ${{ matrix.version }}
-          arch: ${{ matrix.arch }}
       - uses: actions/cache@v4
         env:
           cache-name: cache-artifacts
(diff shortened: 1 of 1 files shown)


Sbozzolo:
Testing somewhere on x86 might be worth keeping if you are about weirder hardware/setups. One of the things that changes between x86 and x64 is the default size of integers (32 bit instead of 64), so you might catch some bugs/edge cases by keeping it around.

Sbozzolo (approved):

glwagner:
> Testing somewhere on x86 might be worth keeping if you care about weirder hardware/setups. One of the things that changes between x86 and x64 is the default size of integers (32 bit instead of 64), so you might catch some bugs/edge cases by keeping it around.

Word.

I think I set it up to run on Mac, Windows, and then Ubuntu with both x86 and x64.

I've never run into a situation where something failed on x86 but ran on x64. Probably, lower-level folks have to solve those problems first...

It is important for us to test on windows though.

The actual meaning of x64 is pretty confusing, but this helped me:

https://www.reddit.com/r/linuxquestions/comments/ra4vjj/difference_between_x86_x8664_amd64_and_x64/

The reason we want to make this change is because "Apple Silicon" (M-series processors) are _not_ x64. But they are 64 bit, duh! 😂 (x64 is a little too short of a designation, maybe). I guess github actions uses the M-series, such that forcing x64 yields an "emulation mode" that runs slow.

A bit more on x64 vs M-series:

https://www.reddit.com/r/mac/comments/u2k1t2/what_is_the_difference_between_mac_m1_vs_apple/

Sbozzolo:
Here's a test that would pass in x64 but fail in x86 (I think):
```julia
myvar = 1e11
@test myvar > 10^10
```
(Because of integer overflow)

https://docs.julialang.org/en/v1/manual/integers-and-floating-point-numbers/#Overflow-behavior

glwagner:
> Here's a test that would pass in x64 but fail in x86 (I think):
> 
> ```julia
> myvar = 1e11
> @test myvar > 10^10
> ```
> 
> (Because of integer overflow)
> 
> https://docs.julialang.org/en/v1/manual/integers-and-floating-point-numbers/#Overflow-behavior

Hmm, so maybe with big `active_cell_map` we might get failure? @simone-silvestri

glwagner:
Finally figured out a pattern that allows us to show the architecture and also test x86... learning github actions esotera

Sbozzolo:
(Just realized that example should have been
```julia
myvar = 1e9
@test myvar < 10^10
```
if I did the math right, 10^10 would be -1,474,836,47 on a 32 bit machine [-2,147,483,648 + (10,000,000,000 - 2,147,483,647) + 1 ], so the test would still pass)

.github/workflows/ci.yml
```diff
@@ -15,26 +15,21 @@ jobs:
...
           - ubuntu-latest
+          - windows-latest
         arch:
           - x64
-          - x86
         include:
-          # test macOS and Windows with latest Julia only
           - os: macOS-latest
-            arch: x64
+            arch: arm64
```
Sbozzolo, line 23: It's `aarch64` ;)

simone-silvestri:
> > Here's a test that would pass in x64 but fail in x86 (I think):
> > 
> > ```julia
> > myvar = 1e11
> > @test myvar > 10^10
> > ```
> > 
> > (Because of integer overflow)
> > 
> > https://docs.julialang.org/en/v1/manual/integers-and-floating-point-numbers/#Overflow-behavior
> 
> Hmm, so maybe with big `active_cell_map` we might get failure? @simone-silvestri 

Hmmm, for active cells map the type is enforced to be the smallest unsigned integer that can represent the largest integer so in theory if the indices are large we should have `UInt64`. However, I am not sure about how `UInt64` work on 32-bit architectures. Are they converted to 32 bit integers automatically? Then it might be a problem for indexing.

glwagner:
> Hmmm, for active cells map the type is enforced to be the smallest unsigned integer that can represent the largest integer so in theory if the indices are large we should have `UInt64`. However, I am not sure about how `UInt64` work on 32-bit architectures. Are they converted to 32 bit integers automatically? Then it might be a problem for indexing.

Interesting, so it seems maybe most stuff will work on x86 but there may be a few things that don't. Hard to test huge active cells map via github actions though.

PR #195, opened 2024-10-21
Remove concurrency on `init`

simone-silvestri:
We have a concurrency flag on the `init` step in CI that disallows multiple agents. We had problems of the whole CI pipeline stalling if one job stalls so maybe it is better to remove this flag. @Sbozzolo is this flag intentional? Will we incur problems on the Caltech cluster if we remove it?

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -13,8 +13,6 @@ env:
 steps:
   - label: "initialize"
     key: "init"
-    concurrency: 1
-    concurrency_group: 'depot/climaocean-ci'
     env:
       TEST_GROUP: "init"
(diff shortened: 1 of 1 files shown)


Sbozzolo:
The flag is prevents two init jobs from running at the same time. We added this because init jobs often modify some shared spaces, so they could corrupt each other. 

These flags should not prevent you from running multiple pipelines at the same time: only the init steps. If you have seen pipelines stalling recently, it might be due to the fact that we are also upgrading our software infrastructure, and sometimes the version that is running is the new one (that is not fully tested yet).

simone-silvestri:
I guess we can close this

Issue #206, opened 2024-10-31
Avoiding redundant data download in parallel plus a nice way to download data prior to launching an MPI job


glwagner:
If we try to run a simulation for the first time with `mpiexec`, I think we can end up redundantly downloading the same data from different MPI processes.

I'm not sure the best way to handle this --- we could only download from rank 0, or we can somehow distribute the downloading amongst processes (even cooler). 

A related issue is that we don't have a simple way to download data in isolation before running a script (the data is all downloaded automatically by trying to run a simulation, which is very nice of course). But probably there are situations where a user may want to get the downloading out of the way in advance before launching a job and it would be nice to make that simple.

simone-silvestri:
should be closed

PR #207, opened 2024-10-31
Distributed DataWrangling

simone-silvestri:
This PR is a proposal to solve #206 

It introduces a couple of distributed utilities:
```julia
blocking_download(url, filepath; kw...) # only rank 0 downloads
blocking_run(cmd) # only rank 0 executes the command
distributed_run(cmd) # if `cmd isa Union{AbstractVector, Tuple}` split the commands throughout the ranks
```

We would have to find a way to test this since distributed tests are not in place yet


simone-silvestri:
Closed in favor of #208

Issue #205, opened 2024-10-31
`retrieve_bathymetry` fails on a 1 degree grid


glwagner:
MWE:

```julia
julia> using Oceananigans, OrthogonalSphericalShellGrids, ClimaOcean
Precompiling ClimaOcean
  3 dependencies successfully precompiled in 9 seconds. 216 already precompiled.

julia> grid = TripolarGrid(size = (360, 170, 1),
                           z = (0, 1),
                           north_poles_latitude = 55,
                           first_pole_longitude = 75)
360×170×1 OrthogonalSphericalShellGrid{Float64, Periodic, RightConnected, Bounded} on CPU with 4×4×4 halo and with precomputed metrics
├── centered at (λ, φ) = (75.0, 1.8005)
├── longitude: Periodic  extent 360.156 degrees     variably spaced with min(Δλ)=0.00798438, max(Δλ)=1.05373
├── latitude:  RightConnected  extent 171.0 degrees variably spaced with min(Δφ)=0.0127345, max(Δφ)=1.00592
└── z:         Bounded  z ∈ [0.0, 1.0]              regularly spaced with Δz=1.0

julia> bathymetry = retrieve_bathymetry(grid)
[ Info: Regridding bathymetry from existing file /Users/gregorywagner/.julia/scratchspaces/0376089a-ecfe-4b0e-a64f-9c555d74d754/Bathymetry/ETOPO_2022_v1_60s_N90W180_surface.nc.
ERROR: ArgumentError: The northern latitude cannot be less than -90 degrees.
Stacktrace:
 [1] validate_lat_lon_grid_args
   @ ~/.julia/packages/Oceananigans/IsF1o/src/Grids/latitude_longitude_grid.jl:281 [inlined]
 [2] LatitudeLongitudeGrid(architecture::CPU, FT::DataType; size::Tuple{…}, longitude::Tuple{…}, latitude::Tuple{…}, z::Tuple{…}, radius::Float64, topology::Nothing, precompute_metrics::Bool, halo::Tuple{…})
   @ Oceananigans.Grids ~/.julia/packages/Oceananigans/IsF1o/src/Grids/latitude_longitude_grid.jl:189
 [3] LatitudeLongitudeGrid
   @ ~/.julia/packages/Oceananigans/IsF1o/src/Grids/latitude_longitude_grid.jl:174 [inlined]
[...]

PR #209, opened 2024-10-31
Fix a precision problem in the ETOPO1 dataset

simone-silvestri:
This PR addresses the precision problem that we incur if we want to regrid a bathymetry up to 90 degrees north.
We probably also want to allow some tolerance in `validate_lat_lon_grid_args` in Oceananigans.jl

https://github.com/CliMA/Oceananigans.jl/blob/f2a8fb32251135f9cd9b230e0873f7bc1936f762/src/Grids/latitude_longitude_grid.jl#L279-L285

see issue #205

### DIFF
--- a/src/Bathymetry.jl
+++ b/src/Bathymetry.jl
@@ -111,8 +111,8 @@ function regrid_bathymetry(target_grid;
 
     FT = eltype(target_grid)
 
-    φ_data = dataset["lat"][:]
-    λ_data = dataset["lon"][:]
+    φ_data = dataset["lat"][:] 
+    λ_data = dataset["lon"][:] 
     z_data = convert(Array{FT}, dataset["z"][:, :])
 
     # Convert longitude from (-180, 180) to (0, 360)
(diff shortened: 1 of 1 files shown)


src/Bathymetry.jl
```diff
@@ -111,8 +111,8 @@ function regrid_bathymetry(target_grid;
 
     FT = eltype(target_grid)
 
-    φ_data = dataset["lat"][:]
-    λ_data = dataset["lon"][:]
+    φ_data = dataset["lat"][:] 
+    λ_data = dataset["lon"][:]
```
navidcy, line 115:
```suggestion
    φ_data = dataset["lat"][:]
    λ_data = dataset["lon"][:]
```

src/Bathymetry.jl
```diff
@@ -142,9 +142,9 @@ function regrid_bathymetry(target_grid;
     jj = j₁:j₂
 
     # Restrict bathymetry _data to region of interest
```
navidcy, line 144:
```suggestion
    # Restrict bathymetry coordinate_data to region of interest
```

glwagner (approved):
Happy to try this for now though I think a more general fix may be needed eventually

PR #220, opened 2024-11-05
Reorganize files so we can collaborate better

glwagner:
We have an `experiments` directory where we are going to collaborate on a coarse one degree setup as well as a mesoscale resolving setup.

@simone-silvestri @rafferrari @seamanticscience @xkykai


rafferrari:
My first attempt at raising a comment to this line: https://github.com/CliMA/ClimaOcean.jl/blob/c0ddc83a4f372e86241562b6c47b982da5044420/experiments/one_degree_simulation/one_degree_simulation.jl#L30C1-L30C13

navidcy (approved):

PR #228, opened 2024-11-10
Remove literated examples; a try

navidcy:
I'd like to see the docs built once and then add literate examples one by one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,31 +1,10 @@
-<!-- Title -->
-<h1 align="center">
-  ClimaOcean.jl
-</h1>
+# ClimaOcean.jl
 
-<!-- description -->
-<p align="center">
-  <strong>🌎 A framework for realistic ocean-only and coupled ocean + sea-ice simulations driven by prescribed atmospheres and based on
-          <a href="https://github.com/CliMA/Oceananigans.jl">Oceananigans</a>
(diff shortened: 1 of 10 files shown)


glwagner (approved):

navidcy:
Note: I changed the README style a bit to consist _only_ markdown. This way we can use the README as the Docs landing page and we only have to maintain one thing.

glwagner:
> Note: I changed the README style a bit to consist _only_ markdown. This way we can use the README as the Docs landing page and we only have to maintain one thing.

Ultimately I think we will want different pages but we can start with that. I think we will want to devote quite a bit to these docs eventually, when it becomes the main source of info about realistic ocean and sea ice simulations. So it will be worth it to tailor the README and docs.

navidcy:
Yes, agreed that this is just a start.
Better having the readme there than nothing. We can alter later!

glwagner:
Probably you're right. But a counter argument is that it gives us the excuse to put off writing legitimate docs! :-D I kinda think it looks bad to copy and paste that though. I will try to update it soon.

navidcy:
Oh feel free to remove then, no problem!

glwagner:
I'll write something. I don't see it up yet though.

navidcy:
Oh... it'll take few hours (if the deploy succeeds!)

navidcy:
https://buildkite.com/clima/climaocean-examples/builds/879#01931931-405f-4220-8726-e0aacde7f428

PR #230, opened 2024-11-10
Caching inpainting data + some updates to the one degree simulation

glwagner:
The main thing this PR does is refactor `ECCO_field` so that inpainted data is cached and reloaded. The inpainted data is saved to a `jld2` file. This really speeds up prototyping especially with small models (such as the three degree model in `one_degree_simulation.jl`) where inpainting dominates the cost of initializing a simulation.

PR #232, opened 2024-11-10
free_surface default for a DistributedGrid

simone-silvestri:


src/OceanSimulations/OceanSimulations.jl
```diff
@@ -44,7 +45,7 @@ default_free_surface(grid) = SplitExplicitFreeSurface(grid; cfl=0.7)
 
 # 70 substeps is a safe rule of thumb for an ocean at 1/4 - 1/10th of a degree
 # TODO: pass the cfl and a given Δt to calculate the number of substeps?
-default_free_surface(grid::TripolarGrid) = SplitExplicitFreeSurface(grid; substeps = 70)
+default_free_surface(grid::Union{TripolarGrid, DistributedGrid}) = SplitExplicitFreeSurface(grid; substeps = 70)
```
glwagner, line 48: Why not just make this always the default?

Note that this doesn't cover ImmersedBoundaryGrid but it needs to (which is done on #230)

All these changes need tests I think.

simone-silvestri:
This is not required anymore.

PR #236, opened 2024-11-11
Add the near-global example in the Docs

navidcy:


simone-silvestri:
nice, shall we also add the other ones since the docs are built now?

navidcy:
I wanted to add them one by one. I think the surface fluxes and the Mediterranean were broken (last week that I was trying them out).

simone-silvestri:
I think the surface flux test case is working. The Mediterranean test case was not ready yet.

navidcy:
OK, I'll add them all but the Mediterranean in a next PR

navidcy:
Actually I just run the examples locally (except the Mediterranean) and they all run until the end no errors.

PR #242, opened 2024-11-12
Add empty PR template

navidcy:
I don't like the prefilled PR template...


glwagner (approved):
Thanks for this, I agree that the template is counter productive

Issue #237, opened 2024-11-11
Some bugs in the near global example


simone-silvestri:
I think there is something wrong in the fluxes with the near-global example. I have ran it locally after it was not running in PR #137 and the results were a little odd.

The simulation is running but the extrema of SST drift extremely quickly, reaching -35 and +47 degree celsius in a month.
This is the video that is produced by the example that shows that temperature increases quite steeply in the tropics.

https://github.com/user-attachments/assets/94017de1-9751-42a1-ad9b-f71080c6431d

The output is the following:
```julia
[ Info: In-painting ECCO temperature
[ Info: In-painting ECCO salinity
[ Info: Initializing simulation...
[ Info: Iter: 0, time: 0 seconds, Δt: 1.500 minutes, max|u|: (0.00e+00, 0.00e+00, 0.00e+00) m s⁻¹, extrema(T): (31.25, -1.96) ᵒC, wall time: 2.681 minutes
[ Info:     ... simulation initialization complete (6.395 seconds)
[ Info: Executing initial time step...
┌ Warning: Simulation stopped during initialization.
└ @ Oceananigans.Simulations ~/.julia/packages/Oceananigans/s1DfC/src/Simulations/run.jl:129
[ Info:     ... initial time step complete (3.408 minutes).
[ Info: Iter: 4800, time: 5 days, Δt: 1.500 minutes, max|u|: (1.73e+00, 2.26e+00, 1.87e-02) m s⁻¹, extrema(T): (34.14, -10.23) ᵒC, wall time: 40.442 minutes
[ Info: Simulation is stopping after running for 1.289 hours.
[ Info: Simulation time 10 days equals or exceeds stop time 10 days.
[ Info: Iter: 9600, time: 10 days, Δt: 1.500 minutes, max|u|: (2.07e+00, 2.05e+00, 9.71e-03) m s⁻¹, extrema(T): (37.57, -15.24) ᵒC, wall time: 36.879 minutes
[ Info: Initializing simulation...
[ Info:     ... simulation initialization complete (1.970 ms)
[ Info: Executing initial time step...
[ Info:     ... initial time step complete (3.960 seconds).
[ Info: Iter: 10320, time: 15 days, Δt: 10 minutes, max|u|: (1.94e+00, 1.97e+00, 1.61e-02) m s⁻¹, extrema(T): (41.27, -17.67) ᵒC, wall time: 5.780 minutes
[...]

PR #246, opened 2024-11-13
Clamp `LinearlyTaperedRestoringMask` to lie within (0, 1)

simone-silvestri:


navidcy:
What does this do? Is it related to #237?

simone-silvestri:
> What does this do? Is it related to #237?

I don't think so. This is because the mask might be larger than one for particular input choices. 
For example, if the domain spans (-80, 80) in latitude and someone would have specified a linearly tapered polar mask with 
```julia
mask = LinearlyTaperedPolarMake(southern = (-70, -60), northern = (60, 70))
```
between (-80, -70) and (70, 80) the mask would have been larger than one.

glwagner:
Heh, I did not understand that from the PR title. I changed it. Hopefully it's right.

glwagner:
Also this description should be put into the source code.

src/DataWrangling/ECCO/ECCO_mask.jl
```diff
@@ -88,17 +88,18 @@ end
                                southern = (-75, -70),
                                z = (-20, 0))
 
-Build a mask that is linearly tapered in latitude inbetween the northern and southern edges.
-The mask is constant in depth between the z and is equal to zero everywhere else.
+Build a mask that is linearly tapered in latitude between the northern and southern edges.
+The mask is constant in depth between the z and equals zero everywhere else.
+The mask is capped to one if the latitude is beyond the outer northern or southern edges.
```
glwagner, line 93: Usually we call this "clipping", not "capping".

glwagner:
```suggestion
The mask is limited to lie between (0, 1).
```

glwagner (approved):

PR #247, opened 2024-11-14
Put nice formatting in the README + new docs landing page

glwagner:
I can't remember if this was never added in the first place or got eliminated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,14 @@
-# ClimaOcean.jl
+<!-- Title -->
+<h1 align="center">
+  ClimaOcean.jl
+</h1>
 
-### 🌎 A framework for realistic ocean-only and coupled ocean + sea-ice simulations driven by prescribed atmospheres and based on [Oceananigans](https://github.com/CliMA/Oceananigans.jl) and [ClimaSeaIce](https://github.com/CliMA/ClimaSeaIce.jl).
+<!-- description -->
+<p align="center">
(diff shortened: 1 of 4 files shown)


navidcy:
I converted the readme in all markdown so it can be the landing page in the docs

glwagner:
Ah ok but I think the README is the most read thing so it has higher priority right?

navidcy:
This PR will break the docs. Happy with it though if you like the style better. 

Perhaps let’s just revert the landing docs page to an empty placeholder one like before? Empty is better than broken.

navidcy:
I can do it.

glwagner:
What do we want to put on the landing page?

glwagner:
The actual process of copy/pasting or writing code to change the landing page is probably more effort than actually just writing something simple!

navidcy:
Yeah, I'll write 1-2 sentences and urge people to look at the examples.

glwagner:
I wrote a few paragraphs let me know what you think.

glwagner:
I wrote a few paragraphs let me know what you think.

navidcy (approved):

glwagner:
Shouldn't the label be called "build docs"? Why say "built" in past tense?

navidcy:
typo ;)

glwagner:
> typo ;)

touche

PR #254, opened 2024-11-15
Fix a bug in `MinimumTemperatureSeaIce` and limit temperature in the water column

simone-silvestri:
This was for sure the problem with the cooling case in #251 and I believe is also what is causing #237, 
because we cap cooling fluxes on the base of the temperature of the bottom cell.
This is almost always 0 in the arctic because it is immersed, causing cooling fluxes to always be active and reducing the temperature much below the freezing point. I am not sure about the temperature increasing too much in the tropics though.

I ll also build the examples to see if this is the problem

PR #255, opened 2024-11-15
Merge `MinimumTemperatureSeaIce` and `FreezingLimitedOceanTemperature`

simone-silvestri:
When we don't have a sea ice model, we probably want to 
- limit cooling fluxes over sea ice (done by `MinimumTemperatureSeaIce`)
- limit temperature in the water column (done by `FreezingLimitedOceanTemperature`)

so it makes sense to merge the two types into one

### DIFF
--- a/src/OceanSeaIceModels/freezing_limited_ocean_temperature.jl
+++ b/src/OceanSeaIceModels/freezing_limited_ocean_temperature.jl
@@ -18,7 +18,7 @@ All fluxes are shut down when the surface is below the `T < Tₘ` except for hea
 
 the melting temperature is a function of salinity and is controlled by the `liquidus`.
 """
-FreezingLimitedOceanTemperature(; liquidus = LinearLiquidus()) = FreezingLimitedOceanTemperature(liquidus) 
+FreezingLimitedOceanTemperature(FT::DataType) = FreezingLimitedOceanTemperature(LinearLiquidus(FT))
 
 const FreezingLimitedCoupledModel = OceanSeaIceModel{<:FreezingLimitedOceanTemperature}
 
--- a/src/OceanSeaIceModels/freezing_limited_ocean_temperature.jl
+++ b/src/OceanSeaIceModels/freezing_limited_ocean_temperature.jl
@@ -82,7 +82,7 @@ end
         Sₒ = ocean_salinity[i, j, 1]
 
         Tₘ = melting_temperature(liquidus, Sₒ)
-    
+
         τx = centered_velocity_fluxes.u
         τy = centered_velocity_fluxes.v
         Jᵀ = net_tracer_fluxes.T
--- a/src/OceanSeaIceModels/freezing_limited_ocean_temperature.jl
+++ b/src/OceanSeaIceModels/freezing_limited_ocean_temperature.jl
@@ -8,6 +8,16 @@ struct FreezingLimitedOceanTemperature{L}
     liquidus :: L
 end
 
+"""
+    FreezingLimitedOceanTemperature(FT=Float64; liquidus = LinearLiquidus(FT)) 
+
+The minimal possible sea ice representation, providing an "Insulating layer" on the surface and clipping the 
+temperature above the freezing point. Not really a ``model'' per se, however, 
+it is the most simple way to make sure that temperature does not dip below freezing temperature. 
+All fluxes are shut down when the surface is below the `T < Tₘ` except for heating.
+
+the melting temperature is a function of salinity and is controlled by the `liquidus`.
+"""
 FreezingLimitedOceanTemperature(FT=Float64; liquidus = LinearLiquidus(FT)) = FreezingLimitedOceanTemperature(liquidus) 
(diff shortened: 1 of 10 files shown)


src/OceanSeaIceModels/ocean_sea_ice_model.jl
```diff
@@ -69,7 +69,7 @@ function heat_capacity(eos::TEOS10EquationOfState{FT}) where FT
     return convert(FT, cₚ⁰)
 end
 
-function OceanSeaIceModel(ocean, sea_ice=MinimumTemperatureSeaIce();
+function OceanSeaIceModel(ocean, sea_ice=freezing_limited_ocean_temperature();
```
glwagner, line 72:
```suggestion
function OceanSeaIceModel(ocean, sea_ice=FreezingLimitedOceanTemperature();
```

?

glwagner (approved):

test/test_surface_fluxes.jl
```diff
@@ -137,6 +137,53 @@ end
     @test turbulent_fluxes.sensible_heat[1, 1, 1] ≈ Qs
     @test turbulent_fluxes.latent_heat[1, 1, 1]   ≈ Ql
     @test turbulent_fluxes.water_vapor[1, 1, 1]   ≈ Mv
+
+    @info " Testing MinimumTemperatureSeaIce..."
```
glwagner:
```suggestion
    @info " Testing FreezingLimitedOceanTemperature..." 
```

src/OceanSeaIceModels/freezing_limited_ocean_temperature.jl
```diff
@@ -18,7 +18,7 @@ All fluxes are shut down when the surface is below the `T < Tₘ` except for hea
 
 the melting temperature is a function of salinity and is controlled by the `liquidus`.
 """
-FreezingLimitedOceanTemperature(FT=Float64; liquidus = LinearLiquidus(FT)) = FreezingLimitedOceanTemperature(liquidus) 
+FreezingLimitedOceanTemperature(; liquidus = LinearLiquidus()) = FreezingLimitedOceanTemperature(liquidus)
```
glwagner, line 21:
```suggestion
FreezingLimitedOceanTemperature() = FreezingLimitedOceanTemperature(LinearLiquidus())
```

just my opinion that we don't need a kwarg when its just one property

glwagner: but i guess this is a new opinion

glwagner: its a new anti-kwarg Greg

simone-silvestri: ahah okok, agreed

navidcy:
`freezing_limited_ocean_temperature` is used before included?

https://github.com/CliMA/ClimaOcean.jl/blob/d6ba22f935793a908f55f24d8e8f85832ae037ce/src/OceanSeaIceModels/ocean_sea_ice_model.jl#L72

I'm getting errors.

src/OceanSeaIceModels/freezing_limited_ocean_temperature.jl
```diff
@@ -0,0 +1,102 @@
...
+            net_tracer_fluxes,
+            grid,
+            sea_ice.liquidus,
+            ocean_temperature,
+            ocean_salinity)
+
+    return nothing
+end
+
+@kernel function _cap_fluxes_on_sea_ice!(centered_velocity_fluxes,
```
glwagner, line 71:
```suggestion
@kernel function _limit_fluxes_over_sea_ice!(centered_velocity_fluxes,
```

navidcy:
Noting that the preview of this PR still shows temperatures at -40C

https://clima.github.io/ClimaOceanDocumentation/previews/PR255/literated/near_global_ocean_simulation/#Running-the-simulation-for-real

navidcy:
#254 showed improved temperatures

Does this PR incorporate #254?

glwagner:
It supercedes #254. This PR is supposed to fix temperatures throughout the water column. I don't think the global solution is a good way to test it. We should use a single column simulation. It would be nice to add an explicit test. We need to keep merging these, because we are visiting caltech very soon and have to make sure we are all synced. It would be great to add an explicit test that is cheaper than the global simulation in the examples.

glwagner:
Thinking bout this a bit more --- this PR changes something that does affect the examples. However, it does not change the examples directly.

I think that we should not use the examples as a "test". The examples are extremely expensive and this is a slow strategy.

If we find that we are checking the examples as a way to test something it means that we are missing a cheaper and more explicit test. So to take that lesson here, we want to build a test for this utility that can be iterated and merged quickly, with the theoretical expectation that the example will work. If the example subsequently fails, we need to add _another_ test that ensures the example works. We should keep going until we have all of the tests we need to be confident that the example will work.

PR #262, opened 2024-11-19
Speed up near global example

simone-silvestri:


glwagner (approved):
Let's consider making this default

navidcy:
We could merge this PR since it's done it's purpose and look into the NaNs in a new PR?

glwagner:
We have to find the NaNs ASAP but unfortunately we are a little tied up this week

simone-silvestri:
https://github.com/CliMA/Oceananigans.jl/pull/3940 was causing the crashing. I ran it on main and it is not crashing anymore. I will bump the minor version in oceananigans and we can get the example back and running.

glwagner:
great

PR #264, opened 2024-11-19
Avoid extra step in computing turbulent fluxes

simone-silvestri:
Instead of converting twice (first ocean velocities then fluxes) between native and latlon grid, we can just convert once the atmosphere velocities to the native grid

@jm-c

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,7 @@
 
 The MIT License (MIT)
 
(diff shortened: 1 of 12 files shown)


jm-c (commented):
This looks good (but did not try to run this on Curvilinear grid).

glwagner (approved):
Merge!

PR #271, opened 2024-11-21
Fixes for flux computation + add fluxes regression tests

simone-silvestri:
The problem with NaNs in the example should be cleared. Let's see about the fluxes and the one degree


simone-silvestri:
The manifest here will be removed before merging

glwagner (approved):

simone-silvestri:
Ok, the fluxes have been corrected and I have added a regression test. The examples still crash until we tag a new version of Oceananigans, but let's merge this one now and open yet another one to work on the one degree simulation

src/OceanSeaIceModels/CrossRealmFluxes/atmosphere_ocean_fluxes.jl
```diff
@@ -365,9 +365,9 @@ end
     time = Time(clock.time)
 
     @inbounds begin
-        Tₒ = ocean_temperature[i, j, 1]
+        Tₒ = ocean_temperature[i, j, kᴺ]
```
navidcy, line 368: this was definitely one issue for getting warm SSTs!

this PR is from long ago though... not part of our current problems

but there was indeed a bug because the bulk formulae was seeing the temperature at the ocean floor instead of the one at the ocean surface and that was resulting in much much more heat fluxes

cc @taimoorsohail

PR #278, opened 2024-11-26
A first stab at a diagnostic surface temperature implementation

simone-silvestri:
This PR implements two types: 
- `BulkTemperature`
- `SkinTemperature`

The first type is supposed to represent the default option we have know: the surface temperature is provided and equal to  the temperature of "surface" model (either sea ice or ocean). This temperature is fixed and not iterated over.

The second type represents a diagnostic temperature that is diagnosed by a flux balance between the 
- top turbulent surfaces (dependent on the surface temperature and iterated by the solver)
- top radiative fluxes (prescribed)
- outgoing longwave (calculated from the surface temperature in the previous iteration)
- internal fluxes: provided from the `SkinTemperature`, for the moment only a simple diffusive flux with prescribed diffusivity leading to an explicit expression for temperature, but they can become more complicated.

In this second case the temperature is solved for in the `flux_balance_temperature(::SkinTemperature, T_ocean, net_external_flux)` function

These types are provided to the `similarity_theory` as a new field named `surface_temperature_type`. The surface temperature is then stored in a `T_surface` field which is added to `similarity_theory.fields`

There is a bit of a pitfall in this implementation: it is difficult to implement temperature-dependent radiative properties, because radiative properties (that depend on i and j and the grid) are passed already computed to the solver. 
I would be against passing i, j and the grid to the surface flux solver (it would entangle the solver with oceananigans so it would be difficult to swap it out with a `SurfaceFluxes.jl` solver), so we could rethink albedo and emissivity computation.


simone-silvestri:
A very crude first calculation in `experiments/single_column_experiments/os_papa_surface_temperature.jl` shows the difference between prescribed and diagnostic surface temperature. In the latter case, the internal fluxes are just a very simple diffusive flux with 
```math
\kappa \frac{T_s - T_o}{\Delta z}
```
where $\kappa = 10^{-2}$, $T_s$ is the incognita, $T_o$ is the temperature of the first ocean cell, and $\Delta z$ is half the upper cell spacing.
The solid lines are the results of `DiagnosticSurfaceTemperature` and the dashed ones are the respective variables calculated using `PrescribedSurfaceTemperature`

Uploading single_column_profiles.mp4…

There is a large difference in the ocean surface temperature (about 1 degree K), but very little in the fluxes.

simone-silvestri:
However, the surface diffusivity predicted by CATKE in the above video is $O(1)$, much larger than the $\kappa = 10^{-2}  m^2s^{-1}$ that I prescribed. If I use $\kappa = 0.5 m^2s^{-1}$ there is even less difference in the surface temperature and barely any in the heat fluxes.

https://github.com/user-attachments/assets/3cd2818b-321f-4ba2-bb02-f3708aa4af9c

I need to test this on the whole surface to see if there are maybe hot or cold spots that show more difference. As it stands, the net difference in heat flux is less than 1 $Wm^{-2}K^{-1}$. Difficult to assess the impact on a 10-day timescale.

simone-silvestri:
There is actually quite a large difference in net heat flux: This is the same computation as in the `generate_surface_fluxes.jl` but done also with a `DiagnosticSurfaceTemperature` with a diffusivity of 0.1.
Here I am showing the difference of net heat flux (`Q_prescribed - Q_diagnostic`) on the left and surface temperature (`Ts_prescribed - Ts_diagnostic`) on the right. 
The surface temperature difference is upward of 2 degrees and the net heat flux has differences upward of 40 Wm2K-1 (which is quite large). In practice, a diagnostic temperature seems to have much more extreme fluxes, i.e heating regions (where solar radiation peaks) are heating much more and cooling regions are cooling much more. There might be a bug in the solver though, because some regions NaN.

Screenshot 2024-11-26 at 4 46 38 PM

simone-silvestri:
The external fluxes in the surface temperature computation were oriented in the wrong direction 😅 . The new computation looks much better (same plot as the above). The computation of the surface temperature is actually mitigating the surface fluxes (as it should), where heating regions heat less and cooling regions cool less

Screenshot 2024-11-26 at 5 00 12 PM

simone-silvestri:
There were some gross bugs in the implementation (including the sign of the fluxes). Correcting the implementation the results look like this (same plot as before)
Screenshot 2024-11-26 at 5 19 18 PM

So in general a diagnostic temperature seems to be _mitigating_ (slightly) the surface fluxes.

src/OceanSeaIceModels/CrossRealmFluxes/surface_temperature.jl
```diff
@@ -0,0 +1,96 @@
...
+#       δ
+#
+# Where the LHS is the internal diffusive flux inside the ocean (within the boundary layer of thickness δ) 
+# plus the (semi-implicit) outgoing longwave flux and the RHS are the remaining atmospheric and radiative fluxes
+# provided explicitly.
+@inline flux_balance_temperature(F::DiffusiveFlux, θo, Qn) = (θo - Qn / F.κ * F.δ)
+
+# he flaw here is that the ocean emissivity and albedo are fixed, but they might be a function of the 
+# surface temperature, so we might need to pass the radiation and the albedo and emissivity as arguments.
+@inline function retrieve_temperature(st::DiagnosticSurfaceTemperature, θ₀, ℂ, 𝒬₀,
```
glwagner, line 72: What does "retrieve" mean?

src/OceanSimulations/OceanSimulations.jl
```diff
@@ -42,10 +43,30 @@ default_or_override(override, alternative_default=nothing) = override
...
+    # - 2.5 minutes for a 1/32 degree ocean
+
+    return 40minutes / Δθ
+end
+
 const TripolarOfSomeKind = Union{TripolarGrid, ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:TripolarGrid}}
-default_free_surface(grid::TripolarOfSomeKind) = SplitExplicitFreeSurface(grid; substeps=70)
+
+function default_free_surface(grid::TripolarOfSomeKind; 
+                              fixed_Δt = infer_maximum_Δt(grid),
```
glwagner, line 64:
```suggestion
                              fixed_Δt = compute_maximum_Δt(grid),
```

"inference" implies a process of deduction based on indirect information which is not what is happening here

plus its better to re-use terminology

src/OceanSeaIceModels/CrossRealmFluxes/surface_temperature.jl
```diff
@@ -0,0 +1,96 @@
...
+    # upwelling radiation is calculated explicitly 
+    Ru = upwelling_radiation(θ₀, radiation_properties) 
+    Rn = Rd + Ru
+    
+    u★ = Σ★.momentum
+    θ★ = Σ★.temperature
+    q★ = Σ★.water_vapor
+ 
+    # sensible heat flux + latent heat flux (positive out of the ocean)
+    Qs = - ρₐ * u★ * (cₚ * θ★ + q★ * ℰv)
```
glwagner, line 88: Elsewhere in the code "Qs" means sensible ("s" for "sensible") so maybe another symbol

src/OceanSeaIceModels/CrossRealmFluxes/similarity_theory_turbulent_fluxes.jl
```diff
@@ -381,24 +407,42 @@ end
     q₀ = AtmosphericThermodynamics.vapor_specific_humidity(ℂ, 𝒬₀)
```
glwagner, line 407: This isn't taking into account `θ₀` is it? This is the main effect of the prognostic temperature --- it changes the saturation vapor pressure thereby impacting latent heat fluxes

glwagner: You have to rebuild the thermodynamic state with the new temperature

glwagner: or actually just recompute it, I think there is a function for this

simone-silvestri: oh right! good find

src/OceanSeaIceModels/CrossRealmFluxes/surface_temperature.jl
simone-silvestri: I guess compute is a better nomenclature

src/OceanSeaIceModels/CrossRealmFluxes/surface_temperature.jl
```diff
@@ -0,0 +1,106 @@
...
+@inline upwelling_radiation(θ₀, r) = r.σ * r.ϵ * θ₀^4
+
+####
+#### Bulk surface temperature (the easiest case)
+####
+
+""" 
+    struct BulkTemperature end
+
+A type to represent the surface temperature used in the flux calculation.
```
glwagner, line 18:
```suggestion
Represents the surface temperature used in fixed-point iteration for surface fluxes following similarity theory.
```

simone-silvestri:
Shall we merge?

glwagner:
There was a question about recomputing the saturation specific humidity. Was this changed and did it change the posted results?

glwagner:
It's a very nice result that it converges faster by the way.

simone-silvestri:
> There was a question about recomputing the saturation specific humidity. Was this changed and did it change the posted results?

Yep I changed it and indeed, the difference between the different solvers is much larger now in the order of 0.2 degrees instead of 0.02. I will post here another test showing the differences

experiments/single_column_experiments/os_papa_surface_temperature.jl
glwagner, line 1: I'd be ok to just have one Papa simulation, which uses the diagnostic surface temperautre (no need to keep the other one)

src/OceanSeaIceModels/CrossRealmFluxes/surface_temperature.jl
```diff
@@ -0,0 +1,106 @@
...
+# Where the LHS is the internal diffusive flux inside the ocean (within the boundary layer of thickness δ) 
+# and the RHS are the atmospheric and radiative fluxes are provided explicitly, or
+# 
+#   Θₒ - Θₛⁿ⁺¹    σϵ θₛⁿ⁺¹θₛⁿ³
+# κ ---------- - ------------ = Jᵀ (all fluxes positive upwards)
+#       δ           ρₒ cpₒ
+#
+# Where the LHS is the internal diffusive flux inside the ocean (within the boundary layer of thickness δ) 
+# plus the (semi-implicit) outgoing longwave radiation and the RHS are the remaining atmospheric and radiative fluxes
+# provided explicitly.
```
glwagner, line 76: I appreciate the comment, but which method was chosen?

simone-silvestri: right, the fully explicit one. Should I remove the comment? Or maybe just update it

src/OceanSeaIceModels/CrossRealmFluxes/similarity_theory_turbulent_fluxes.jl
```diff
@@ -360,45 +374,88 @@ end
...
-    q₀ = AtmosphericThermodynamics.vapor_specific_humidity(ℂ, 𝒬₀)
+
+    # Recomputing the saturation specific humidity at the surface based on the new temperature
+    q₀ = seawater_saturation_specific_humidity(ℂ, θ₀, S₀, 𝒬₁,
+                                               water_mole_fraction,
+                                               water_vapor_saturation,
+                                               AtmosphericThermodynamics.Liquid())
+    
+    𝒬ₛ = AtmosphericThermodynamics.PhaseEquil_pTq(ℂ, 𝒬₀.p, θ₀, q₀)
+    q₀ = AtmosphericThermodynamics.vapor_specific_humidity(ℂ, 𝒬ₛ)
```
glwagner, line 415: You don't want to call this again, because the "seawater" part is important --- we account for the h2O mole fraction there. Also, we may be using a different saturation parameterization than `AtmosphericThermodynamics` (this is to comply with OMIP protocol)

simone-silvestri: hmm ok, if I remove that line the results are different (regression tests do not pass). Isn't ok that the seawater part called above?

glwagner: weird, not sure I understand what is going on then

src/OceanSeaIceModels/CrossRealmFluxes/surface_temperature.jl
glwagner: I think just add a remark about what the code does now; the linearization strategy is a potential optimization that isn't taken yet

glwagner: or you can remove it :-D

src/OceanSeaIceModels/CrossRealmFluxes/similarity_theory_turbulent_fluxes.jl
glwagner: I mean, were the regression tests also incorrect?

glwagner: Or maybe I am confused. Does `vapor_specific_humidity(ℂ, 𝒬ₛ)` compute a new specific humidity or merely extract it?

glwagner: I think I'm confused sorry, I guess `vapor_specific_humidity(ℂ, 𝒬ₛ)` just extracts it. It says nothing about saturation there.

glwagner: But I am weirded out that it changes the result. What is the difference between the two q0?

simone-silvestri: then probably yes, because in the main code we do:
https://github.com/CliMA/ClimaOcean.jl/blob/0fff4fa3564d50a5f44ba7fe45b6e647242387b8/src/OceanSeaIceModels/CrossRealmFluxes/similarity_theory_turbulent_fluxes.jl#L380-L382 
where `𝒬₀` holds the previously calculated `q₀`

glwagner: Ok, on `main` what we do is compute q0, and then we assemble it into a thermodynamic / dynamic state. Then we pass the whole state to the solver. Within the solver, I guess it uses `vapor_specific_humidity` to extract q0 --- or at least, that's what I thought. This result suggests though that somehow q0 is changed?

glwagner:
> > There was a question about recomputing the saturation specific humidity. Was this changed and did it change the posted results?
> 
> Yep I changed it and indeed, the difference between the different solvers is much larger now in the order of 0.2 degrees instead of 0.02. I will post here another test showing the differences

Can you also show that the difference depends on the grid spacing at the top? It should right?

simone-silvestri:
> > > There was a question about recomputing the saturation specific humidity. Was this changed and did it change the posted results?
> > 
> > 
> > Yep I changed it and indeed, the difference between the different solvers is much larger now in the order of 0.2 degrees instead of 0.02. I will post here another test showing the differences
>  
> Can you also show that the difference depends on the grid spacing at the top? It should right?

It does not for the moment because the boundary layer thickness `δ` is a parameter passed to the `SkinTemperature` type. 
In practice, the first cell is assumed to be within the well-mixed region, so T represents a cell average and `δ` is the boundary layer at the top. These are the differences with the default parameters for `SkinTemperature` (`κ = 0.01` and `δ = 1.0`):

Screenshot 2024-12-09 at 1 02 16 PM

These parameters probably require calibration.

glwagner (approved):

src/OceanSeaIceModels/CrossRealmFluxes/similarity_theory_turbulent_fluxes.jl
simone-silvestri: I think it is changed in the `vapor_specific_humidity` function. It is kind of tough to follow the thread, but it looks like the `vapor_specific_humidity` builds a `PhasePartition` type to calculate the humidity vapor fraction associated with the thermodynamic state

glwagner:
> > Can you also show that the difference depends on the grid spacing at the top? It should right?
> 
> It does not for the moment because the boundary layer thickness `δ` is a parameter passed to the `SkinTemperature` type. In practice, the first cell is assumed to be within the well-mixed region, so T represents the (mixed) cell average, and `δ` is the boundary layer at the top of the mixed region. These are the differences with the default parameters for `SkinTemperature` (`κ = 0.01` and `δ = 1.0`) (BulkTemperature - SkinTemperature)

Hmm right, maybe it shouldn't depend on the grid spacing. That's helpful actually.

PR #284, opened 2024-12-02
(0.3.0) Cleaner syntax for ECCOFieldTimeSeries and ECCORestoring

glwagner:
Closes #231

PR #251, opened 2024-11-15
Refactor PrescribedAtmosphere and add an idealized atmosphere simulation

glwagner:
This PR refactors `PrescribedAtmosphere` so that we can set up ocean simulations beneath idealized atmospheres.

I did the minimum possible in this PR, which basically preserves the kind of atmosphere that JRA55 uses --- everything as a FieldTimeSeries. Even to get this to work I needed to make some changes, including to `Oceananigans`. So it depends on https://github.com/CliMA/Oceananigans.jl/pull/3928

I also had to use a hacked version of `OrthogonalSphericalShellGrids` (#glw/fix) since it's broken for Oceananigans 0.94 currently.

I added an example called `idealized_single_column_simulation.jl` which evolves a linearly stratified ocean beneath constant 10 m/s winds at 20 deg C and with 400 W / m^2 solar insolation. It's interesting to see the difference between this and the typical idealized forcing that we use --- with constant winds, there is significant cooling and evaporation. Also, the fluxes decrease in time because 1) the ocean cools and 2) the ocean speeds up, decreasing the momentum flux in a relative winds bulk formulation, which in turn decreases the latent heat flux because the latent heat flux is proportional to the friction velocity.

Here's a movie of the results:

https://github.com/user-attachments/assets/1837bf42-9fda-42e7-b562-77ba0a7167cb

Here's the single column setup:

```julia
using ClimaOcean
using Oceananigans
using Oceananigans.Units
using SeawaterPolynomials

# Ocean state parameters
T₀ = 20   # Surface temperature, ᵒC
S₀ = 35   # Surface salinity
N² = 1e-4 # Buoyancy gradient due to temperature stratification
f = 0     # Coriolis parameter

# Atmospheric state parameters
Tₐ = 273.15 + 20 # Kelvin
u₁₀ = 10 # wind at 10 m, m/s
qₐ = 0.01 # specific humidity
Qℓ = 400 # shortwave radiation (W m⁻², positive means heating right now)

# Build the atmosphere
atmosphere_grid = RectilinearGrid(size=(), topology=(Flat, Flat, Flat))
atmosphere_times = range(0, 1day, length=3)
atmosphere = PrescribedAtmosphere(atmosphere_grid, atmosphere_times)

parent(atmosphere.tracers.T) .= Tₐ     # K
parent(atmosphere.velocities.u) .= u₁₀ # m/s
parent(atmosphere.tracers.q) .= qₐ     # mass ratio
[...]

PR #286, opened 2024-12-05
(0.5.1) `Metadata` for JRA55

### LINKED ISSUE #182: Refactoring JRA55
The JRA55 requires a refactor because it assumes that all the data for a variable is contained in one file. 
This is true for only for repeat year forcing; in reality, multi-year forcing is contained in different files.
This requires  us to restructure 

1) the downloading procedure
2) the JRA55NetCDFbackend 
3) the setting of the data in memory

We also need a design that allows us to possibly add different versions efficiently. I propose to generalize `ECCOMetadata` to a general `Metadata` type and structure the JRA55 data handling in the same way we do ECCO at the moment and provide a `JRA55RepeatYear` and a `JRA55MultiYear` version.

@glwagner

simone-silvestri:
This PR is an initial proposal to generalize `ECCOMetadata` to `Metadata` and rework the JRA55 module to use `Metadata`. In this way, we can have different JRA55 versions (repeat year and multiple year) and we can define a `download_dataset` function to download the dataset independently of using JRA55 as we can do for ECCO

This PR also removes the ability to generate a `JRA55FieldTimeSeries` directly interpolated on the ocean grid, since we need to interpolate anyways when we compute fluxes

closes #182


glwagner:
I think we should get #251 and #284 merged first, do you want to help with those? Otherwise we will have conflicts.

simone-silvestri:
sounds good

simone-silvestri:
Another change I made is that `JRA55FieldTimeSeries` does not accept `time_indices` anymore, but `dates` (as ECCO does). I was thinking this is more in line with the metadata implementation and uniformizes the interface between the different datasets. I can revert it to how it was previously though

simone-silvestri:
Tests should pass so this is ready to merge / review

simone-silvestri:
This PR should be ready to merge

src/DataWrangling/ECCO/ECCO_restoring.jl
```diff
@@ -156,7 +129,7 @@ function ECCOFieldTimeSeries(metadata::ECCOMetadata, grid::AbstractGrid;
     inpainting isa Int && (inpainting = NearestNeighborInpainting(inpainting))
     backend = ECCONetCDFBackend(time_indices_in_memory, metadata; on_native_grid, inpainting, cache_inpainted_data)
 
-    times = ECCO_times(metadata)
+    times = native_times(metadata)
```
glwagner, line 132: this used to be called `dates`, is it `times` now?

src/DataWrangling/metadata.jl
```diff
@@ -0,0 +1,113 @@
+using CFTime
+using Dates
+using Base: @propagate_inbounds
+
+struct Metadata{D, V} 
+    name  :: Symbol
+    dates :: D
```
glwagner, line 7: I see `dates` here but elsewhere `times`. Pick one...

src/DataWrangling/metadata.jl
```diff
@@ -0,0 +1,113 @@
...
+Arguments
+=========
+- `metadata`: The metadata containing the date information.
+- `start_time`: The start time for calculating the time difference. Defaults to the first date in the metadata.
+
+Returns
+=======
+An array of time differences in seconds.
+"""
+function native_times(metadata; start_time=first(metadata).dates)
```
glwagner, line 92:
```suggestion
function native_times(metadata; start_date=first(metadata).dates)
```

its mixing up the meaning of "time" and "date"

src/DataWrangling/metadata.jl
```diff
@@ -0,0 +1,113 @@
...
+    times = zeros(length(metadata))
+    for (t, data) in enumerate(metadata)
+        date = data.dates
+        time = date - start_time
```
glwagner, line 96:
```suggestion
        delta = date - start_time
```

src/DataWrangling/metadata.jl
simone-silvestri: dates it better, times has no reference

src/DataWrangling/metadata.jl
simone-silvestri: this was just moved from a file to another. We can open an issue to discuss this fact maybe

glwagner: the sneaky introduction occurred in a prior PR

glwagner (approved):

src/DataWrangling/metadata.jl
glwagner: @simone-silvestri fix this

navidcy:
merge?

navidcy:
Is this a breaking change?

simone-silvestri:
Yeah, it is breaking. 
I think there is the last problem remaining that is how to deal with `rivers_and_icebergs` because they have different dates than the other fields. 
I will try to make them work today (however I think they will work nicely only when we have a `DateTime` clock in Oceananigans) add a simulation test that includes them then we can merge.

glwagner:
We somehow have to add information about what "t = 0" means ... somewhere ... if we don't use `DateTime` in the clock directly

simone-silvestri:
I have fixed the rivers and icebergs. Since this PR is an intense refactor and breaking the whole JRA55/ECCO modules, I ll run the examples on top of it before merging.

navidcy:
From the Docs build:

```Julia
ERROR: LoadError: LoadError: UndefVarError: `Metadata` not defined
in expression starting at /central/scratch/esm/slurm-buildkite/climaocean-examples/2189/climaocean-examples/docs/src/literated/generate_surface_fluxes.md:1
when executing the following code block from inputfile `/central/scratch/esm/slurm-buildkite/climaocean-examples/2189/climaocean-examples/examples/generate_surface_fluxes.jl`
```julia
T_metadata = Metadata(:temperature; version=ECCO4Monthly())
S_metadata = Metadata(:salinity; version=ECCO4Monthly())
```

navidcy:
New error...

https://buildkite.com/clima/climaocean-examples/builds/2198#019577ee-1d73-4c61-bf16-aac55a52b241/190-1078

navidcy:
Seems like somehow we are trying to construct an InterfaceState with `u★`, `θ★`, and `q★` Float64 and the rest variables Float32?

navidcy:
I added `Oceananigans.defaults.FloatType = Float32` in the example; that did the job.

navidcy:
> I added `Oceananigans.defaults.FloatType = Float32` in the example; that did the job.

Do we need this in other/all examples?

navidcy (approved):

navidcy:
Seems like the `JRA55NetCDFBackend` has no effect?
Look below. Whether I call with `JRA55NetCDFBackend(2)` or `JRA55NetCDFBackend(248)` I get an atmosphere with 2920 time slices (that is a full year).

```Julia
julia> simulation_days = 31
31

julia> snapshots_per_day = 8 # corresponding to JRA55's 3-hour frequency
8

julia> time_indices_in_memory = simulation_days * snapshots_per_day
248

julia> atmosphere = JRA55PrescribedAtmosphere(longitude = λ★,
                                              latitude = φ★,
                                              backend = JRA55NetCDFBackend(time_indices_in_memory))

2×2×1×2920 PrescribedAtmosphere{Float32} on LatitudeLongitudeGrid:
├── times: 2920-element StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}
├── surface_layer_height: 10.0
└── boundary_layer_height: 600.0

julia> atmosphere = JRA55PrescribedAtmosphere(longitude = λ★,
                                              latitude = φ★,
                                              backend = JRA55NetCDFBackend(2))
2×2×1×2920 PrescribedAtmosphere{Float32} on LatitudeLongitudeGrid:
├── times: 2920-element StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}
├── surface_layer_height: 10.0
└── boundary_layer_height: 600.0
```

glwagner:
> > I added `Oceananigans.defaults.FloatType = Float32` in the example; that did the job.
> 
> Do we need this in other/all examples?

We don't want to switch all examples to using Float32, do we?

glwagner:
> Seems like somehow we are trying to construct an InterfaceState with `u★`, `θ★`, and `q★` Float64 and the rest variables Float32?

Then you should convert all the variables to the appropriate type which is probably Float64 in this case (but needs to be inferred from something --- the exchange grid makes sense to me for this).

glwagner:
> > > I added `Oceananigans.defaults.FloatType = Float32` in the example; that did the job.
> > 
> > 
> > Do we need this in other/all examples?
> 
> We don't want to switch all examples to using Float32, do we?

Given what I read, I think you should remove that because it is not the right solution here.

navidcy:
Gotcha! I tried converting and then that changing the default was the way to go. I’m actually bit confused regarding when to change the default. But I’ll drop it from the example and infer the conversion from the exchanger grid!

navidcy:
> Seems like the `JRA55NetCDFBackend` has no effect? Look below. Whether I call with `JRA55NetCDFBackend(2)` or `JRA55NetCDFBackend(248)` I get an atmosphere with 2920 time slices (that is a full year).
> 
> ```julia
> julia> simulation_days = 31
> 31
> 
> julia> snapshots_per_day = 8 # corresponding to JRA55's 3-hour frequency
> 8
> 
> julia> time_indices_in_memory = simulation_days * snapshots_per_day
> 248
> 
> julia> atmosphere = JRA55PrescribedAtmosphere(longitude = λ★,
>                                               latitude = φ★,
>                                               backend = JRA55NetCDFBackend(time_indices_in_memory))
> 
> 2×2×1×2920 PrescribedAtmosphere{Float32} on LatitudeLongitudeGrid:
> ├── times: 2920-element StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}
> ├── surface_layer_height: 10.0
> └── boundary_layer_height: 600.0
> 
> julia> atmosphere = JRA55PrescribedAtmosphere(longitude = λ★,
>                                               latitude = φ★,
>                                               backend = JRA55NetCDFBackend(2))
> 2×2×1×2920 PrescribedAtmosphere{Float32} on LatitudeLongitudeGrid:
> ├── times: 2920-element StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}
> ├── surface_layer_height: 10.0
> └── boundary_layer_height: 600.0
> ```

This comes from 

https://github.com/CliMA/ClimaOcean.jl/blob/04200c5516787724ead40f5c8a3bb22fe994d6ea/src/DataWrangling/JRA55/JRA55_field_time_series.jl#L321

`native_times()` replaced `jra55_times()` method, but the former ignores the backend so it always returns the times for the whole JRA dataset (in which case above a whole year). @simone-silvestri is there something missing or is this intentional?

simone-silvestri:
> Seems like the `JRA55NetCDFBackend` has no effect? Look below. Whether I call with `JRA55NetCDFBackend(2)` or `JRA55NetCDFBackend(248)` I get an atmosphere with 2920 time slices (that is a full year).

right, the backend has no effect on the data in the timeseries, but it indicates how this is organized in memory. 

In main, a way to limit the time series is to pass `time_indices` as positional arguments like so
```julia
time_indices = 1:10
JRA55PrescribedAtmosphere(time_indices; kw...)
```
for 10 elements, while in this PR we switch to `dates` as a keyword argument, to anticipate adding different JRA55 metadata versions that have different dates:
```julia
start_date = DateTime(1990, 1, 1)
end_date  = DateTime(1990, 2, 1)
dates = range(start_date, end_date, step = Hour(3))
JRA55PrescribedAtmosphere(; dates, kw...)
```
The source of truth for the dates associated with a particular dataset `version` can be retrieved with 
```julia
all_dates(version, name)
```
for example:
```julia
julia> all_dates(JRA55RepeatYear(), :temperature)
Dates.DateTime("1990-01-01T00:00:00"):Dates.Hour(3):Dates.DateTime("1990-12-31T21:00:00")

julia> all_dates(JRA55RepeatYear(), :river_freshwater_flux)
Dates.DateTime("1990-01-01T00:00:00"):Dates.Day(1):Dates.DateTime("1990-12-31T00:00:00")

julia> all_dates(JRA55MultipleYears(), :pressure)
Dates.DateTime("1958-01-01T00:00:00"):Dates.Hour(3):Dates.DateTime("2021-01-01T00:00:00")
```
This applies also to `ECCO` datasets:
```julia
julia> all_dates(ECCO4Monthly(), :salinity)
Dates.DateTime("1992-01-01T00:00:00"):Dates.Month(1):Dates.DateTime("2023-12-01T00:00:00")
```
that are also constructed with a `dates` keyword argument.
```julia
julia> ECCOFieldTimeSeries(:temperature; dates = DateTime(1993, 1, 1):Month(1):DateTime(1993, 2, 1))
[ Info: Note: ECCO temperature data is in /Users/simonesilvestri/.julia/scratchspaces/0376089a-ecfe-4b0e-a64f-9c555d74d754/ECCO.
[ Info: Note: ECCO temperature data is in /Users/simonesilvestri/.julia/scratchspaces/0376089a-ecfe-4b0e-a64f-9c555d74d754/ECCO.
[ Info: Note: ECCO temperature data is in /Users/simonesilvestri/.julia/scratchspaces/0376089a-ecfe-4b0e-a64f-9c555d74d754/ECCO.
[ Info: Note: ECCO temperature data is in /Users/simonesilvestri/.julia/scratchspaces/0376089a-ecfe-4b0e-a64f-9c555d74d754/ECCO.
[ Info: Note: ECCO temperature data is in /Users/simonesilvestri/.julia/scratchspaces/0376089a-ecfe-4b0e-a64f-9c555d74d754/ECCO.
[ Info: Note: ECCO temperature data is in /Users/simonesilvestri/.julia/scratchspaces/0376089a-ecfe-4b0e-a64f-9c555d74d754/ECCO.
720×360×50×2 FieldTimeSeries{ClimaOcean.DataWrangling.ECCO.ECCONetCDFBackend} located at (Center, Center, Center) on Oceananigans.Architectures.CPU
├── grid: 720×360×50 LatitudeLongitudeGrid{Float32, Oceananigans.Grids.Periodic, Oceananigans.Grids.Bounded, Oceananigans.Grids.Bounded} on Oceananigans.Architectures.CPU with 7×7×3 halo and with precomputed metrics
├── indices: (:, :, :)
├── time_indexing: Cyclical(period=5.3568e6)
├── backend: ECCONetCDFBackend(1, 2)
└── data: 734×374×56×2 OffsetArray(::Array{Float32, 4}, -6:727, -6:367, -2:53, 1:2) with eltype Float32 with indices -6:727×-6:367×-2:53×1:2
    └── max=31.2508, min=-1.98588, mean=3.33469
```

Before merging I can write more details in the description of the PR and I am for changes if suggestes

navidcy:
Gotcha.

So at the papa example, I did this:

https://github.com/CliMA/ClimaOcean.jl/blob/94d0cae9e392ebb87f7372fd3d68252df7d505e2/examples/single_column_os_papa_simulation.jl#L78

What would be a better way to do it in 

https://github.com/CliMA/ClimaOcean.jl/blob/94d0cae9e392ebb87f7372fd3d68252df7d505e2/examples/single_column_os_papa_simulation.jl#L64

?

simone-silvestri:
We could also think about passing `start_date`, `end_date` and frequency (which should be a multiple of the native frequency)

glwagner:
Why aren't we using `backend` anymore?

@navidcy the size of the data in memory is displayed at the top of the file:

```julia
julia> atmosphere = JRA55PrescribedAtmosphere(longitude = λ★,
                                              latitude = φ★,
                                              backend = JRA55NetCDFBackend(2))
2×2×1×2920 PrescribedAtmosphere{Float32} on LatitudeLongitudeGrid:
├── times: 2920-element StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}
├── surface_layer_height: 10.0
└── boundary_layer_height: 600.0
```

the "backend" and "times" give different information. "backend" refers to how the data is stored, not to the number of time points. The objective is to have an object that can represent a long time-series, but which may have some data stored on disk rather than in memory ("partially in memory").

glwagner:
@simone-silvestri I think that using `start_date` and `end_date` in the constructor would be better than having to provide a vector of dates. This alleviates the user from needing to understand the frequency of the information, and from having to learn the name of an additional function like `all_dates`.

By the way, I am still worried that the design of `Metadata` is confusing, whereby the difference between "many dates" and a "single date" is expressed by the type of `dates`. In addition to being more intuitive and easier to understand, `Vector{Metadata}` is also more general. For example you can string two different filenames / types together with this method, by putting two different Metadata into a single vector.

simone-silvestri:
ok, I ll put here the `start_date`, `end_date` for the moment.

Then when we merge it I can open a new PR that changes `metadata` to only one date and uses `Vector{Metadata}` instead of a metadata of dates. However, as a con, I think we lose the option to represent a full dataset with one `metadata` object.
Another option behind the confusion of dates vs date is to have `dates` being always an iterable

glwagner:
>  However, as a con, I think we lose the option to represent a full dataset with one `metadata` object.

I see how this is a consequence. But what specifically is the downside?

simone-silvestri:
I think the difference is whether we want `Metadata` to be part of the user interface or just an internal convenience type that helps us wrangle data. 

In the first case, it is nice to be able to represent a dataset composed of a version, a name, and a set of dates in a type (here we can probably go the `Metadata` - `Metadatum` route), while a `Vector{Metadata}` might not be a consistent dataset because there can possibly be differences in versions or variable names.

In the latter case, there is no problem with the user interface, `Metadata` can be hidden in the internals with the user interface exposing only the methods that explicitly accept `variable_name`, `dates`, and `version` for things like `ECCOFieldTimeSeries` and `JRA55FieldTimeSeries`. We would have to think about how to change the `set!` function to remove the `Metadata` option. In the internals, when we pass a `Vector{Metadata}` to functions like `FieldTimeSeries` we need to ensure consistency of `variable_name` and `version` of all the elements of the vector.

glwagner:
> I think the difference is whether we want `Metadata` to be part of the user interface or just an internal convenience type that helps us wrangle data.
> 
> In the first case, it is nice to be able to represent a dataset composed of a version, a name, and a set of dates in a type (here we can probably go the `Metadata` - `Metadatum` route), while a `Vector{Metadata}` might not be a consistent dataset because there can possibly be differences in versions or variable names.
> 
> In the latter case, there is no problem with the user interface, `Metadata` can be hidden in the internals with the user interface exposing only the methods that explicitly accept `variable_name`, `dates`, and `version` for things like `ECCOFieldTimeSeries` and `JRA55FieldTimeSeries`. We would have to think about how to change the `set!` function to remove the `Metadata` option. In the internals, when we pass a `Vector{Metadata}` to functions like `FieldTimeSeries` we need to ensure consistency of `variable_name` and `version` of all the elements of the vector.

Ok, please clarify what you see as the trade-offs for user interface. I think you are assuming a design, but it is not being explicitly state. I can't respond or judge what user interface you are referring to,  unless you state it explicitly.

Part of the problem is the proposal to define someting like

```julia
struct Metadatum # represents a single file
    # properties
end
```

const Metadata = Vector{Metadatum}
```

has no specific implications for how `Metadata` is constructed. We can keep the same constructor.

The difference is mainly that we could define a _new_ constructor for a single `Metadatum`. This avoids the confusion that we pass `dates = a_single_date` to `Metadata`.

So in summary i don't see what changes would be required of the user interface. The difference is that we can expand the user interface to make more sense while leaving existing components unchanged.

**Comment by glwagner** [comment]
Here's a simple example, just one of many possibilities

```julia
# source
Metadata(name; version, dates) = [Metadatum(name; version, date for date in dates]
```

```julia
datum = Metadata(name; version, date)
data = Metadata(name; version, dates)
```

the idea of a "version" for general Metadata is weird to me by the way. What concept are we expressing here. `data_source` or just `source`? Version implies that a data is specific but has multiple "versions" or "updates" which doesn't seem right.

simone-silvestri:
I was referring to achieving something similar avoiding a `Vector` that might mix different versions, I guess the end goal is the same.
An example is
```julia
Metadatum
   name
   version 
   date
   dir
end

@propagate_inbounds Base.getindex(m::Metadata, i::Int) = Metadatum(m.name, m.dates[i],   m.version, m.dir)
@propagate_inbounds Base.first(m::Metadata) = Metadatum(m.name, m.dates[1],   m.version, m.dir)
@propagate_inbounds Base.last(m::Metadata) = Metadatum(m.name, m.dates[end], m.version, m.dir)

@inline function Base.iterate(m::Metadata, i=1)
    if (i % UInt) - 1 < length(m)
        return Metadatum(m.name, m.dates[i], m.version, m.dir), i + 1
    else
        return nothing
    end
end
```
also in this way it would be possible to do
```julia
datum = Metadatum(name; version, date)
data = Metadata(name; version, dates)
```

glwagner:
I guess I was seeing the ability to mix "versions" as a perk. For example JRA55 was originally generated up to 2018, but there are updates which continue the dataset past that. You may not be able to form a single consistent dataset (eg a single "version") that encompasses all dates, but it still could be valid to write something like

```julia
up_to_2018 = Metadata(name; dataset=original_dataset, dates=dates_till_2018)
past_2018 = Metadata(name; dataset=continuation, dates=dates_past_2018)
data = vcat(up_to_2018, past_2018)
```

by the way what does "version" mean in the context of a general `Metadatum`? It seems we need a different word like "dataset" or "label" which is more general than just "version"

simone-silvestri:
I guess version is more suited in main where Metadata is `ECCOMetatada` and the dataset can only be part of the ECCO dataset with a specific version (2, 4, etc).

You are right that probably that field name has to change in this PR, I like `dataset` because it points to the dataset the metadata points to

simone-silvestri:
yep, in general the full file path is determined by the whole `metadatum` (name, date, dataset / version, and dir)

glwagner:
Ruminating on "dataset" --- I think it could be a good term, because it expresses the concept of a "category of data". Which is what we mean here, a single `metadatum` refers to one file; a whole dataset will have many files, each of which has a metadatum.

Also just to offer an alternative --- rather than metadatum / metadata, we could have 

```julia
struct Metadata
    name
    dataset
end

const MetadataSeries = Vector{Metadata}
```

semantically, MetadataSeries is easier to distinguish from `Metadata` than `Metadata` is to distinguish from `Metadatum`. It might help too, to emphasize that a vector of `Metadata` is a series of snapshots.

simone-silvestri:
I think I can merge here even if the examples are not passing. We are hitting a scalar indexing problem, which I believe is connected to https://github.com/CliMA/Oceananigans.jl/discussions/4087 but is not part of this PR.

Then we can think about an interface for metadata that allows distinguishing between dates and date

PR #290, opened 2024-12-10
Bugfix `set!`ting ECCO FTS with cycling boundaries

simone-silvestri:
which should happen as soon as `model.clock.time > last(ECCO_times(metadata))`


glwagner:
don't you wnt to add this to the PR with the test? This is test driven dev: open a PR with the test, then commit to the fix, then merge. Boom done.

PR #294, opened 2024-12-11
(0.4.0) Integrate sea ice in the flux computation

simone-silvestri:
This PR is a tentative implementation for atmosphere-sea-ice computations to be able to integrate sea-ice in the `OceanSeaIceModel`. 

The objective is trying to reuse as much as possible of the infrastructure in place for atmosphere - ocean fluxes

I think everything is roughly in place, we just need to change the surface properties and implement a way to limit the surface temperature to the melting temperature in the surface temperature computation


src/OceanSeaIceModels/CrossRealmFluxes/similarity_theory_turbulent_fluxes.jl
```diff
@@ -143,7 +143,7 @@ Keyword Arguments
 - `similarity_profile_type`: The type of similarity profile used to relate the atmospheric state to the 
                              surface fluxes / characteristic scales.
 - `bulk_velocity`: The velocity used to calculate the characteristic scales. Default: `RelativeVelocity()` (difference between
-                   atmospheric and oceanic speed).
+                   atmospheric and surfaceic speed).
```
glwagner, line 146: typo

test/test_surface_fluxes.jl
```diff
@@ -86,8 +86,8 @@ _fractional_indices(at_node, grid, ::Nothing, ::Nothing, ::Nothing) = (nothing,
                 set!(ocean.model, u = uₐ, v = vₐ, T = Tₒ)
 
                 # Compute the turbulent fluxes (neglecting radiation)
-                coupled_model    = OceanSeaIceModel(ocean; atmosphere, similarity_theory)
-                turbulent_fluxes = coupled_model.fluxes.turbulent.fields
+                coupled_model    = OceanSeaIceModel(ocean; atmosphere, ocean_similarity_theory=similarity_theory)
+                turbulent_fluxes = coupled_model.fluxes.turbulent.ocean.fields
```
glwagner, line 90: Just thinking about this... would we rather write `fluxes.turbulent.ocean` or `fluxes.ocean.turbulent`?

I'm ok to keep this pattern but wanted to have the discussion

test/test_surface_fluxes.jl
```diff
@@ -120,7 +120,7 @@ _fractional_indices(at_node, grid, ::Nothing, ::Nothing, ::Nothing) = (nothing,
             # mid-latitude ocean conditions
             set!(ocean.model, u = 0, v = 0, T = 15, S = 30)
             
-            coupled_model = OceanSeaIceModel(ocean; atmosphere, similarity_theory)
+            coupled_model = OceanSeaIceModel(ocean; atmosphere, ocean_similarity_theory=similarity_theory)
```
glwagner, line 123: This doesn't make sense to me. It's a similarity theory for the atmosphere, not for the ocean.

simone-silvestri:
The sea ice model is not quite ready in this PR, but the fluxes have changed to allow the integration of a sea ice model. I would probably merge this and integrate the sea ice in another PR. 

As an example for the integration we can use the one degree simulation.

glwagner:
https://github.com/user-attachments/assets/e39ef880-81ea-4d3c-ba2f-00b14b486e3b

glwagner:
https://github.com/user-attachments/assets/8dbc5b2e-cc65-43d5-8ea6-57d00ee34851

glwagner:
@simone-silvestri can you hold commits on this branch for a bit? I'm working on it. Just got a bunch of conflicts.

glwagner:
As part of this PR we have some important design decisions to make regarding how fluxes are computed and stored. These decisions are not _numerical_ in nature (we know what to do for the physics) but more questions of the user API and internal code design.

I'll start by noting that the way fluxes are currently computed is not correct. This PR will require 3 types of fluxes: atmosphere-ocean, atmosphere-sea-ice, and sea-ice-ocean. (A fourth flux, land-ocean, is implicitly included in atmosphere-ocean for now but we may need to break this out eventually too.)

Right now we do not explicitly store these three types of fluxes. Instead we have an object that represents the _total_ flux into the ocean, which combines the atmosphere and sea-ice fluxes. For example, when we compute atmospheric fluxes, we store the result in the total ocean surface flux fields, then multiply those by $(1 - \aleph)$ where $\aleph$ is the sea ice concentration. (The sea ice concentration $\aleph$ will play a prominent role in this discussion.) See:

https://github.com/CliMA/ClimaOcean.jl/blob/e1d0066b34ff8eeaf848d639b0784f8a9663b0e6/src/OceanSeaIceModels/CrossRealmFluxes/atmosphere_sea_ice_fluxes.jl#L23-L27

The next step is to add the sea-ice contribution to the total ocean flux, multiplied by $\aleph$. This actually isn't implemented yet (though we do compute sea-ice-ocean fluxes, we end up overwriting the salinity flux which is wrong of course):

https://github.com/CliMA/ClimaOcean.jl/blob/e1d0066b34ff8eeaf848d639b0784f8a9663b0e6/src/OceanSeaIceModels/CrossRealmFluxes/sea_ice_ocean_fluxes.jl#L54

In proceeding we have a choice to make. We can proceed along the current path, and fix the sea-ice-ocean fluxes to incorporate concentration correctly. This will necessarily assume that the atmosphere-ocean fluxes have been previously computed, and therefore _add_ the sea-ice contribution to them (rather than overwriting the values as we do now, which assumes that the previous values are from a prior time step for example). Note that currently, the sea ice does not require explicit fluxes at the ice-ocean interface, because the ice-ocean interface is assumed to be in thermal equilibrium at the melting temperature. Therefore, we do not functionally require storing the ice-ocean fluxes; merely adding them to the total ocean flux is sufficient. (This could change in the future though.) One advantage of this design is that we do not store unnecessary data in memory.

Another possibility is to explicitly store each interface flux separately. This requires more memory allocation (12 extra 2D fields, I think, in the worst case scenario). For a model that's high resolution in the vertical this is fairly trivial extra storage. But it might represent a lot of memory for a low-vertical resolution in the vertical, nearly tantamount to an extra 3D field. But perhaps even that is not a huge price to pay...

To spell this out, consider that the computational model for the coupled system is effectively like this:

```
    atmosphere
------------------
     sea ice     |
------------------
      ocean      |
------------------
```

Consider the flux of some quantity denoted by $J$. There are three fluxes: $J_{ao}$ for atmosphere-ocean, $J_{ai}$ for atmosphere-sea-ice, and $J_{io}$ for sea-ice-ocean. The _total_ flux into the top of the ocean $J_o$ is computed using the sea ice concentration $\aleph$:

$$ J_o = \aleph J_{io} + (1 - \aleph) J_{ao} $$

The total flux into the top of the sea ice is just $J^{top}_i = J_{ai}$ and at the bottom of the sea ice is just $J^{bot}_i = J_{io}$.

simone-silvestri:
I also think explicitly computing $J_{ao}$ and $J_{io}$ and later constructing $J_o$ might be better. This is because being able to explore the two fluxes individually will make it easier to debug/analyze. 
Also, we need a $J_i$ for which $J_{ao}$ does not participate, so it is nice to have the individual cross-component fluxes to be able to later discriminate which contributes to which flux. Finally, for coupling purposes, only $J_{ao}$ is required (actually only the turbulent component of $J_{ao}$, the radiative component might need to be handled differently), so it is good to hang on to each individual turbulent component.

src/OceanSeaIceModels/InterfaceComputations/atmosphere_ocean_fluxes.jl
```diff
@@ -0,0 +1,164 @@
...
+                       Mp = atmosphere_fields.Mp.data,
+                       h_bℓ = atmosphere.boundary_layer_height)
+
+    flux_formulation = coupled_model.interfaces.atmosphere_ocean_interface.flux_formulation
+    interface_fluxes = coupled_model.interfaces.atmosphere_ocean_interface.fluxes
+    interface_temperature = coupled_model.interfaces.atmosphere_ocean_interface.temperature
+    interface_properties = coupled_model.interfaces.atmosphere_ocean_interface.properties
+    ocean_properties = coupled_model.interfaces.ocean_properties
+    atmosphere_properties = (thermodynamics_parameters = atmosphere.thermodynamics_parameters,
+                             reference_height = atmosphere.reference_height)
```
simone-silvestri, line 36:
```suggestion
    atmosphere_properties = (thermodynamics_parameters = thermodynamics_parameters(atmosphere),
                             reference_height = reference_height(atmosphere))
```

src/OceanSeaIceModels/InterfaceComputations/atmosphere_sea_ice_fluxes.jl
```diff
@@ -0,0 +1,182 @@
...
+
+    # Estimate interface specific humidity using interior temperature
+    q_formulation = interface_properties.specific_humidity_formulation
+    qₛ = saturation_specific_humidity(q_formulation, ℂₐ, 𝒬ₐ.ρ, Tₛ, Sᵢ)
+
+    # Guess
+    Sₛ = zero(FT) # what should we use for interface salinity?
+    initial_interface_state = InterfaceState(u★, u★, u★, uᵢ, vᵢ, Tₛ, Sₛ, convert(FT, qₛ))
+    land = inactive_node(i, j, kᴺ, grid, Center(), Center(), Center())
+    ice_free = hᵢ == 1
```
simone-silvestri, line 128:
```suggestion
    ice_free = hᵢ == 0
```

simone-silvestri:
I cannot approve this PR since I am the author, if someone approves it, I can merge when the tests pass.

src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
```diff
@@ -81,8 +95,8 @@ end
         ρτx = atmos_ocean_fluxes.x_momentum[i, j, 1]    # zonal momentum flux
         ρτy = atmos_ocean_fluxes.y_momentum[i, j, 1]    # meridional momentum flux
 
-        Qio = sea_ice_ocean_fluxes.heat[i, j, 1]
-        Jˢio = sea_ice_ocean_fluxes.salt[i, j, 1]
+        Qio  = get_ice_heat_flux(i, j, 1, grid, sea_ice_ocean_fluxes)[i, j, 1]
+        Jˢio = get_ice_salt_flux(i, j, 1, grid, sea_ice_ocean_fluxes)[i, j, 1]
```
glwagner, line 99: just remember to give some weight to the value of refactoring if that would help clean things up

PR #296, opened 2024-12-13
Updates diagnostic for computing mixed layer depth

glwagner:
The diagnostic is based on a buoyancy differential. I used a default 3e-5, which is the recommended potential density difference 0.03 (see for example [Treguier et al 2023](https://gmd.copernicus.org/articles/16/3849/2023/)) divided by a reference density ~1e3.

We should add a test and also limit by the static column depth from Oceananigans, @simone-silvestri maybe you can look this over.

I also added an example which is not literated, but we could consider literating it.


glwagner:
Hmm looks like this may depend on #284 so let's merge that first

src/Diagnostics/mixed_layer_depth.jl
```diff
@@ -0,0 +1,89 @@
...
+    Δz = zN - zk
+    z★ = zk - Δz/Δb * (Δb★ - Δb)
+
+    # Special case when domain is one grid cell deep
+    z★ = ifelse(Δb == 0, zN, z★)
+
+    # Apply various criterion
+    h = -z★
+    h = max(h, zero(grid))
+    h = min(h, grid.Lz)
```
simone-silvestri, line 85: you mean here?
```suggestion
    H = static_column_depthᶜᶜᵃ(i, j, grid)
    h = min(h, H)
```

simone-silvestri (approved):

src/Diagnostics/mixed_layer_depth.jl
glwagner: bingo

navidcy:
> I also added an example which is not literated, but we could consider literating it.

That's great. But you also deleted the example that _was_ literated. And then #312 put the non-literated example (that uses GLMakie and sliders and what not) in the docs and made all doc builds fail onwards.

navidcy:
My only point was that if the literated file is deleted then we should remove it from `make.jl` otherwise it might trigger mistakes as in #312.

PR #300, opened 2024-12-16
Change default major basins to 1 in `regrid_bathymetry`

glwagner:
I can't really think of a case where we would want more than one major basin, let alone an infinite number of basins!


xkykai (approved):
🤣

PR #306, opened 2024-12-19
Add a directory option to download data in `ECCORestoring`

simone-silvestri:
we could also call it just `dir` instead of `data_dir` since that's how we call it in `ECCOMetadata` and `JRA55_field_time_series`


glwagner:
`dir` is better for sure. I'm not sure how I would interpret the difference between `dir` and `data_dir`?

navidcy (approved):
lgtm

PR #308, opened 2024-12-19
Remove grid from `OceanSeaIceModel`

simone-silvestri:
I think this should work without problems but let's see


glwagner (approved):
This is good, though in the future we may have to consider what grid to use for flux computations. For the moment, we just use the ocean surface grid.

Issue #253, opened 2024-11-15
Better way to smooth bathymetry than "repeated interpolation"?


glwagner:
I think diffusion or a moving average might be a better way to smooth bathymetry. This would preserve the total ocean volume. Also, rather than specifying "number of passes" I feel like it would be more useful to be able to specify something like the "maximum gradient" or "maximum curvature" and iterate until that criterion is reached.

NoraLoose:
> I think diffusion or a moving average might be a better way to smooth bathymetry. This would preserve the total ocean volume. Also, rather than specifying "number of passes" I feel like it would be more useful to be able to specify something like the "maximum gradient" or "maximum curvature" and iterate until that criterion is reached.

That's basically what we do when we prepare the bathymetry for a ROMS simulation. We perform two consecutive steps, first a domain-wide smoothing (to avoid grid-scale instabilities) and then an iterative local smoothing until the maximum slope criterion is satisfied. For more details, see [the ROMS-Tools documentation](https://roms-tools.readthedocs.io/en/latest/methods.html#topography).

NoraLoose:
The domain-wide smoothing step might be more relevant for a terrain-following coordinate (as in ROMS) than for a z-coordinate.

navidcy:
While working on https://github.com/CliMA/ClimaOcean.jl/pull/475 and in an attempt to understand what the `regrid_bathymetry` method does, we were chatting with @glwagner in the morning trying to understand what the effect of the step-by-step interpolations do (eg interpolate into intermediate resolutions before arriving at your final resolution versus interpolating straight from high-resolution to final coarse resolution.) The figure at https://clima.github.io/ClimaOceanDocumentation/stable/literated/generate_bathymetry/ suggests that intermediate interpolations lead to horizontal smoothing...

We both think that that might be more reasonable is to apply the smoothing to the native_grid of the topography (at the highest resolution) and then interpolate to the user-desired resolution.

https://github.com/CliMA/ClimaOcean.jl/pull/477 is a first attempt to allow for new smoothing operations

glwagner:
Curious @NoraLoose do you smooth the _source_ bathymetry, or only the "regridded" field (eg the bathymetry that you can after interpolation / regridding)? I'm also curious whether we should naively interpolate, or use a proper conservative remapping, which conserve the average depth.

NoraLoose:
For the sake of completeness, I will just say that we are actually performing three steps:

[...]

PR #309, opened 2024-12-19
Distributed regridding of bathymetry

simone-silvestri:
Regridding of bathymetry requires extending periodic dimensions and subsequent passes where we need to define new grids. 

This is not compatible with distributed grids for the moment, so it is better to regrid on the global grid and partition the array in the different cores once this operation is done.

Probably, when we find a better way to smooth the bathymetry (as suggested in #253) we can revisit this simple implementation and split the regridding on different cores.


glwagner (approved):
When does this fail? When the global bottom height does not fit on one GPU?

simone-silvestri:
> When does this fail? When the global bottom height does not fit on one GPU?

I think so.

PR #312, opened 2024-12-22
Fix docs

simone-silvestri:


navidcy (approved):

docs/make.jl
```diff
@@ -13,7 +13,7 @@ const EXAMPLES_DIR = joinpath(@__DIR__, "..", "examples")
 const OUTPUT_DIR   = joinpath(@__DIR__, "src/literated")
 
 to_be_literated = [
-    "inspect_ecco_data.jl",
+    "ecco_inspect_temperature_salinity.jl",
```
navidcy, line 16: this was not good idea; this `ecco_inspect_temperature_salinity.jl` seems to be using GLMakie and generally it wasn't made for docs (it includes sliders etc).

PR #318, opened 2024-12-24
Fix ECCO `metadata_url` and `urls` that both adding a trailing `/`

### LINKED ISSUE #316: ECCO's `metadata_url` and `urls` both add a `/` between the url's path and filename
The `metadata_url` at:

https://github.com/CliMA/ClimaOcean.jl/blob/7ddb30bd90023a614dec04f4c238c38dabb83ad2/src/DataWrangling/ECCO/ECCO_metadata.jl#L150-L151

add a `/` after the prefix. But the prefix is inferred from the `urls` at:

https://github.com/CliMA/ClimaOcean.jl/blob/7ddb30bd90023a614dec04f4c238c38dabb83ad2/src/DataWrangling/ECCO/ECCO_metadata.jl#L199-L202

which _also_ include a `/` at the end.

This results in a double `//`.

Which one should we keep; does it matter for elsewhere?

navidcy:
Close #316


navidcy:
So we can literate this example, but I don't know what to do with the sliders. I don't think we can have sliders in the docs...
At the moment, it looks like:
https://clima.github.io/ClimaOceanDocumentation/previews/PR318/literated/ecco_inspect_temperature_salinity/

glwagner:
If this fixes a bug, we should add a test to ensure the bug doesn't happen again. How did this bug get through?

navidcy:
I'm also wondering because there are tests that download ECCO data; I'll look into that.

simone-silvestri:
Repeated `/` are usually not accounted for when navigating URLs or a (UNIX) file system. 
For example, to access ClimaOcean
https://github.com/CliMA/ClimaOcean.jl/
is equivalent to
https://github.com/CliMA/////ClimaOcean.jl/
or
https://github.com/////CliMA/ClimaOcean.jl
The same goes with cd:
`cd /to/my/path` and `cd /to///my/path///` are equivalent.
In this way, it is possible to compose directories or join paths without needing to think whether we need to add a trailing `/` or not.
However, I am not sure about windows systems.

glwagner:
Ok that is what I suspected --- not a bug, but good to clean up regardless. Let's not mark it as a bug though because that may mislead

simone-silvestri (approved):

navidcy:
Seems like there are errors due to scalar operations at

https://github.com/CliMA/Oceananigans.jl/blob/e02564df8ab90fb8f994fddf465b10d123fdfc3a/src/Operators/spacings_and_areas_and_volumes.jl#L204

See: https://buildkite.com/clima/climaocean-ci/builds/2367#01945504-9e29-4245-8d55-4e7942f35978/187-526

But how come this only came up now? Is this related to the [pin in GPUArrays](https://github.com/CliMA/Oceananigans.jl/blob/e02564df8ab90fb8f994fddf465b10d123fdfc3a/Project.toml#L15)?

simone-silvestri:
yeah looks like. Should go away if we use `Oceananigans@0.95.6`

navidcy:
and none of the tests of Oceananigans caught this?

navidcy:
> and none of the tests of Oceananigans caught this?

OK, none of Oceananigans tests caught this because they are all run within a `CUDA.allowscalar()`!

https://github.com/CliMA/Oceananigans.jl/blob/e02564df8ab90fb8f994fddf465b10d123fdfc3a/test/runtests.jl#L19

We should remove this `CUDA.allowscalar() do ... all-the-tests ... end`

cc @glwagner

simone-silvestri:
I am definitely in favor of that. `allowscalar` should be more granular and used only when necessary.

PR #321, opened 2025-01-03
Correct type instability in fluxes computation

simone-silvestri:


glwagner (approved):
I don't think causes an issue since it gets `convert`d in `setindex` but its a good idea anyways

PR #322, opened 2025-01-06
Introduce an interface to allow attaching a prognostic atmosphere

simone-silvestri:
This PR reorganizes a bit the code to allow for a prognostic atmosphere, which means adding some interfaces that can be extended in the atmosphere we want to couple with


src/OceanSeaIceModels/time_step_ocean_sea_ice_model.jl
```diff
@@ -28,6 +29,10 @@ function time_step!(coupled_model::OceanSeaIceModel, Δt; callbacks=[], compute_
         time_step!(sea_ice)
     end
 
+    # Time step the atmosphere
+    # TODO: allow different time-steps for atmosphere and ocean
+    time_step!(atmosphere)
```
glwagner, line 34: Somehow you have to ensure the time-step is size `Δt`

milankl: Numerically you can't change the time step in SpeedyWeather during the integration. You can set it flexibly at initialization though. Whether ClimaOcean's dt is used in SpeedyWeather or SpeedyWeather's step is used in ClimaOcean I don't mind.

glwagner: I think that ClimaOcean should manage the time-step and that the component models time-step should be set based on what ClimaOcean has.

If component models have some restriction then we need to design a way to manage that. This also means that `TimeInterval` schedule cannot be used in the `Simulation` of `OceanSeaIceModel`.

Why don't we write a function like `set_time_step!(atmosphere, dt)` (also for `ocean`). Then SpeedyWeather will throw an error if `dt` is different from its internal unchangeable time-step.

glwagner: or `set_Δt!(atmosphere, Δt)`.

(we have discussed elsewhere about the fact that we often use "time step" as a noun `Δt` as well as a verb for actually taking a time-step, ie `time_step!`... a little annoying)

glwagner: or `set_time_step_size!(atmos, Δt)`

milankl: normally we choose our time step with `Leapfrog(spectral_grid, Δt_at_T31=Minute(40))` with T31 being our default resolution and a linear scaling towards higher resolutions that's automatically applied. But with https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/650/ I've just created a PR that also allows to `set!` the time step more explicitly

glwagner: Sorry for being dull --- do you mean that #650 makes it so that the user can choose the time-step when setting up a SpeedyWeather simulation?

PR #325, opened 2025-01-08
Update arctic_night.jl for latest Oceananigans

glwagner:
`Buoyancy.model` became `BuoyancyForce.formulation`


examples/arctic_night.jl
waywardpidgeon, line 1: Gee - this is the precisely the same code as for idealized_single_column_formulation.jl - see the .mp4 at the end for example. I saw this because the same buoyancy error occcured in both arctic_night.jl and this example. 
Please let us have the actual arctic_night.jl !

glwagner: Right, I think this script was actually added to the repo by mistake! But what do you mean by this: 

> Please let us have the actual arctic_night.jl !

glwagner: The difference between this script and the "idealized single column formulation" is the inital ocean temperature (0 C instead of 20 C) and the atmos temperature (-5 C in arctic night). I think this one also has sea ice which we expect to grow under these conditions.

waywardpidgeon: Thanks - I did miss those differences. I will explore the git tool for checking them. Suggestion: change the name of the .mp4 output file to say arctic_night.mp4 .

glwagner: Can you submit a PR with that change?

waywardpidgeon: Sorry, my time is too limited, could someone else please do this? 

The change needed is to replace in line 114 of the file arctic_night.jl of the examples directory the filename "idealized_atmosphere.mp4" 
by "arctic_night.mp4" or similar. Thanks.

navidcy: it's already like that; I'm resolving this

navidcy (approved):

PR #330, opened 2025-01-15
Changing the output filename to arctic_night_atmosphere.mp4

waywardpidgeon:
I change the name of the output movie so as not to be confused with the output from the file
idealized_single_column_simulation.jl 

PS I named my branch kab which I hope appears in appropriate places rather than the cumbersome waywardpidgeon!


glwagner (approved):
Thank you @waywardpidgeon !

PR #340, opened 2025-02-14
An experiment to compute flux climatology

simone-silvestri:
I have also added a prescribed ocean, so we have a completely prescribed earth model, where the only computation required are the fluxes


simone-silvestri:
I took the opportunity to check how difficult it would be to have a prescribed ocean in ClimaOcean. Apparently, not too difficult! 

btw, these are the output of the script, ordered as: 
(1) mean -> :balance
(2) std -> :deep
(3) max -> :magma
(4) min -> :haline
mean_fluxes
fluxes_std
fluxes_max
fluxes_min

PR #342, opened 2025-02-15
Allow attaching a prognostic atmosphere take 2

simone-silvestri:
Since the interface changed a bit it is easier to rewrite the PR rather than fix #322, since the changes are minimal. This PR supersedes #322 but has the same goal. 

closes #322


glwagner (approved):

simone-silvestri:
I have added a clock to the `PrescribedAtmosphere` so we can remove the `update_model_field_time_series!(atmosphere)` from the `update_state!(coupled_model)` which means that we do not always have to extend `update_model_field_time_series!` for different atmospheric models, but only
```julia
time_step!(atmosphere, Δt)
interpolate_atmospheric_state!(coupled_model) 
compute_net_atmosphere_fluxes!(coupled_model)
@inline thermodynamics_parameters(atmos) = ...
@inline reference_height(atmos) = ...
@inline boundary_layer_height(atmos) = ...
```

glwagner:
makes sense. maybe somehow we can ensure synchronization of clocks somewhere since i guess everybody will have their own

PR #350, opened 2025-02-24
Generalize flux solver stop criteria to include `FixedIterations`

glwagner:
This PR adds the feature to fix the number of iterations used in the flux solver. cc @wsmoses


simone-silvestri:
wouldn't this be achieved by `tolerance=0`, `maxiter=fixed_iteration_number`?

simone-silvestri:
Is this required for an ulterior motive?

simone-silvestri (approved):

PR #353, opened 2025-02-24
Precompute interpolation indices + BarotropicPotentialForcing

glwagner:
This PR adds a new property called `StateExchanger`. The idea is that this type will hold objects needed to exchange information from components to ComponentInterfaces, and back.

The first application of this is to precompute and store the "interpolation indices" for spatial interpolation. This needs be performed only once and then re-used in `interpolate_atmospheric_state`.

This PR also adds a forcing to velocity components in `ocean_simulation` called BarotropicPotential forcing. This includes the atmospheric sea level pressure effect on ocean momentum (eg meteotsunamis etc)


glwagner:
Note this depends on https://github.com/CliMA/Oceananigans.jl/pull/4146

simone-silvestri (approved):

src/OceanSeaIceModels/InterfaceComputations/component_interfaces.jl
```diff
@@ -42,19 +43,65 @@ struct SeaIceOceanInterface{J, P, H, A}
...
+    # TODO: use kernel_parameters not hard code?
+    # kernel_parameters = interface_kernel_parameters(ocean_grid)
+    
+    atmos_grid = atmosphere.grid
+    arch = architecture(exchange_grid)
+    Nx, Ny, Nz = size(exchange_grid)
+
+    # Make an array of FractionalIndices
+    kᴺ = size(exchange_grid, 3)
+    X1 = _node(1, 1, kᴺ + 1, exchange_grid, c, c, f)
```
simone-silvestri, line 85: will this work on a GPU?

test/test_jra55.jl
```diff
@@ -96,11 +96,11 @@ using ClimaOcean.OceanSeaIceModels: PrescribedAtmosphere
 
         # Random regression test
         CUDA.@allowscalar begin
-            @test target_fts[1, 1, 1, 1] == 222.243136478611
+            @test Float32(target_fts[1, 1, 1, 1]) ≈ Float32(222.24313354492188)
 
             # Only include this if we are filling halo regions within
             # interpolate_field_time_series
-            @test target_fts[Nx + 1, 1, 1, 1] == 222.243136478611
+            @test Float32(target_fts[Nx + 1, 1, 1, 1]) ≈ Float32(222.24313354492188)
```
simone-silvestri, line 103: isn't this a symptom that something has changed in the interpolation routine?

simone-silvestri: maybe it's just that the indices are `Float32` now instead of `Float64`?

src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -33,25 +24,28 @@ end
...
+datestr(md::ECCOMetadata) = string(first(md.dates), "--", last(md.dates))
+datestr(md::ECCOMetadatum) = string(md.dates)
+
+Base.summary(md::ECCOMetadata) = string("ECCOMetadata{", versionstr(md), "} of ",
+                                        md.name, " for ", datestr(md))
 
 """
     ECCOMetadata(name::Symbol; 
-                 dates = DateTimeProlepticGregorian(1993, 1, 1),
+                 dates = DateTime(1993, 1, 1),
```
navidcy, line 45: Is the "TimeProlepticGregorian" not needed?

navidcy (approved):
Few remarks:

- is the StateExchanger tested?
- I'm a bit worried with such big changes without having build the docs; things break and then it takes a while to figure out. But I also understand that we need to move one. It's a tough call...

test/test_jra55.jl
glwagner: yes thats right its a floating point precision thing

glwagner: the change is like 1e-5

src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -33,25 +24,28 @@ end
...
                  version = ECCO4Monthly(),
                  dir = download_ECCO_cache)
-
+version
```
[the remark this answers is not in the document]
glwagner: typo

src/DataWrangling/ECCO/ECCO_metadata.jl
glwagner: `DateTime` is on the proleptic gregorian calendar. The other times from `CFTime` are necessary because they are not proleptic gregorian, however.

src/OceanSimulations/barotropic_potential_forcing.jl
```diff
@@ -0,0 +1,44 @@
...
+        return forcing_barotropic_potential(mf.forcing[n])
+    end
+end
+
+forcing_barotropic_potential(sim::Simulation) = forcing_barotropic_potential(sim.model)
+
+function forcing_barotropic_potential(model::HydrostaticFreeSurfaceModel)
+    u_potential = forcing_barotropic_potential(model.forcing.u)
+    v_potential = forcing_barotropic_potential(model.forcing.v)
+    @assert u_potential === v_potential
```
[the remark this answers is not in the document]
glwagner: They need to be the same

glwagner:
> * I'm a bit worried with such big changes without having build the docs; things break and then it takes a while to figure out. But I also understand that we need to move one. It's a tough call...

that's fair. I should have labeled a docs build. Although are docs build broken right now?

glwagner:
> > * I'm a bit worried with such big changes without having build the docs; things break and then it takes a while to figure out. But I also understand that we need to move one. It's a tough call...
> 
> that's fair. I should have labeled a docs build. Although are docs build broken right now?

I'm not sure its rational not to develop. Having continuous docs builds is a way to speed up our workflow slightly, because we address problems as the occur during development. Yet either with continuous docs builds, or occasional docs builds (eg delayed because of CI issue) we have fundamentally the same process wherein docs breakage due to development must be addressed.

We have to take into account the risks of stopping development as well.

src/OceanSeaIceModels/InterfaceComputations/component_interfaces.jl
glwagner: might need an @allowscalar

glwagner: the issue is for triply-flat grids. I could detect that case manually. But this seemed a bit more direct

navidcy:
> > * I'm a bit worried with such big changes without having build the docs; things break and then it takes a while to figure out. But I also understand that we need to move one. It's a tough call...
> 
> that's fair. I should have labeled a docs build. Although are docs build broken right now?
> 
> I do think we should be careful about user-facing development with no docs builds. Internal refactoring like this PR might be easier to justify though.

true! that's why I approved!
It was a minor note/comment.. I do generally agree with your sediment above!

glwagner:
Ah true! Actually the example change was purely cosmetic. Hopefully no changes _needed_ from this but we will see...

navidcy:
> Ah true! Actually the example change was purely cosmetic. Hopefully no changes _needed_ from this but we will see...

Gotcha!

PR #355, opened 2025-02-26
Lay the groundwork for prognostic atmospheres

glwagner:
This PR gets things going, but a bit of refactoring will remain. In particular, we need to define the interface for an atmosphere, and then define `PrescribedAtmosphere` somewhat later. Right now the interface for an atmosphere is partly contained inside `PrescribedAtmosphere`.


src/OceanSeaIceModels/time_step_ocean_sea_ice_model.jl
```diff
@@ -41,7 +41,7 @@ function time_step!(coupled_model::OceanSeaIceModel, Δt; callbacks=[], compute_
     time_step!(ocean)
 
     # Time step the atmosphere
-    time_step!(atmosphere, Δt)
+    time_step!(atmosphere)
```
simone-silvestri, line 44:
```suggestion
    time_step!(atmosphere, Δt)
```
we probably need this if atmospheres need to adapt their `Δt` to the coupled simulation's `Δt`

glwagner: No, I don't want to use a different design for different simulation types. I am going to design a separate interface for setting dt.

src/OceanSeaIceModels/OceanSeaIceModels.jl
```diff
@@ -45,6 +45,14 @@ sea_ice_concentration(sea_ice::SeaIceSimulation) = sea_ice.model.ice_concentrati
 ##### Some implementation
 #####
 
+# Atmosphere interface
+interpolate_atmosphere_state!(interfaces, atmosphere, coupled_model) = nothing
+compute_net_atmosphere_fluxes!(copuled_model) = nothing
```
simone-silvestri, line 50: I think this is already defined in the top file (`OceanSeaIceModels.jl`)

src/OceanSeaIceModels/PrescribedAtmospheres.jl
```diff
@@ -301,7 +302,7 @@ struct PrescribedAtmosphere{FT, M, G, T, U, P, C, F, I, R, TP, TI}
     downwelling_radiation :: R
     thermodynamics_parameters :: TP
     times :: TI
-    reference_height :: FT
+    surface_layer_height :: FT
     boundary_layer_height :: FT
```
simone-silvestri, line 306: I believe this should be in the `flux_formulation` because it is used only for calculating the gustiness in the `Simi;arityTheoryFluxes`. Maybe for another PR, but to keep in mind

simone-silvestri (approved):

src/OceanSeaIceModels/PrescribedAtmospheres.jl
glwagner: in principle this is a diagnostic variable that must be computed from the atmospheric state

src/OceanSeaIceModels/OceanSeaIceModels.jl
glwagner: IT is defined for PrescribedAtmospheres but not a fallback. I'd like to lay the groundwork for distinguishing between the interface implementation, and PRescribedAtmospheres.

src/OceanSeaIceModels/time_step_ocean_sea_ice_model.jl
glwagner: I have rethought this and I think it's a good idea to combine these

PR #360, opened 2025-02-28
Update ETOPO url

glwagner:
Tests indicate that the old url, which pointed to a NOAA website, does not seem to work any more. This PR updates the url to point to a self-hosted location (which may be temporary, but at least things will run in the interim).

### DIFF
--- a/src/Bathymetry.jl
+++ b/src/Bathymetry.jl
@@ -30,12 +30,18 @@ function __init__()
     global download_bathymetry_cache = @get_scratch!("Bathymetry")
 end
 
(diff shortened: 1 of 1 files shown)


natgeo-wong:
Hey @glwagner I saw this pop up on my GitHub feed. Just wanted to let you know that the NOAA website is back up (it was on maintenance this morning).

On the flip side, given everything, if there is a way to store ETOPO data as an artifact that can be used across different packages eventually it might be for the best.

navidcy:
thanks @natgeo-wong

PR #362, opened 2025-02-28
One degree example enhancement

navidcy:
The plots in the one degree example had x/y ticks that did not correspond to lat/lon value. I removed them until we find a way to plot with proper coordinates on the tripolar grid.


glwagner:
Note that as far as I know the only solution is the one that Imaginocean implements. I don't think there's another way to plot on lat lon. Unless GeoMakie has something.

glwagner (approved):

simone-silvestri (approved):

navidcy:
merged to #461

Issue #344, opened 2025-02-20
Workflow for downloading input forcing files without GPU node internet access


taimoorsohail:
Hi all, 

I am trying to run ClimaOcean on the Gadi supercomputer in Australia, and only the login CPU node has internet access on the HPC (for security reasons). 

This means that I can't run examples that require downloaded files without first manually downloading the input files, placing them in the necessary folders, and then submitting a job to the GPU or CPU nodes on the HPC. This is an OK workaround, but I realised that as others run this model, they may also be using HPC environments that don't have internet access outside of the login node. 

I just wanted to flag this as a potential issue, and to discuss whether it may be worth developing a workflow which avoids this need to manually download input files prior to running the model. This may be something that is unavoidable, but I figured I would flag it! Thanks

glwagner:
The files go to `JULIA_DEPOT_PATH`. Can you set your `JULIA_DEPOT_PATH` to point to a place that is accessible from the GPU nodes?

As you say, I think the way to run these cases is to initiate the simulation on CPU, but at a very coarse resolution and only running for a short amount of time.

Do you think that is an acceptable workflow? Perhaps, rather than using a simulation we could develop a utility that's something like `download_data(model)`? That would avoid the step of having to use `run!(simulation)`.

taimoorsohail:
The issue isn't file access - the GPU nodes are able to access the depot path. The issue is that any attempt to download data (using wget, curl etc.) doesn't work because the GPU nodes don't have internet access. 

Yes, the ideal workflow would be to create a standalone function that can be run on the login node (for example, I have written a bash script with wget that downloads the JRA, ECCO and Bathymetry files into the necessary folders directly from the login node). I may be overblowing the issue, but I think CPU/GPU nodes in many HPCs don't have internet access. So integrating a simple download script that can be run from the login node prior to run!(simulation) would help those running on such HPC environments. 

Hope that makes sense!

glwagner:
Why do you have to change the folders that the data is downloaded into?

taimoorsohail:
I don't change the folders the data is downloaded into. I just manually download the data into the folders that code would ordinarily download the data into (if it had internet access).

navidcy:
If I understand correctly @taimoorsohail is wrote a bash script to do what the method `download_data(model)` would be doing, right?

navidcy:
Actually the CPU nodes on HPCs often don’t have internet access either. So the issue is not GPU specific.

[...]

PR #364, opened 2025-03-03
Moved ECCO downloading to beginning of code in `examples/one_degree_simulation.jl`

taimoorsohail:
In light of #344 I have modified the `ECCO_metadata.jl` and `examples/one_degree_simulation.jl` to download ECCO data directly using the `download_datasets` function prior to defining the grid etc. This is because the memory requirements of regridding the bathymetry were too high and the code failed before the download could be initiated on the login node. 

Happy to discuss if moving the downloads to the top of the code is reasonable - I think that it is, as it also requires some user input re: download directories and dates. 

I have only altered the ECCO downloads so far - ETOPO and JRA55 are still occurring further down. 

I have also altered the `ECCO_metadata.jl` file to print that the data is downloading (if it doesn't exist) or that the data can be found in <metadata.dir>. This is mostly so the user knows what exactly is happening, but happy to remove.


src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -243,13 +243,23 @@ function download_dataset(metadata::ECCOMetadata; url = urls(metadata))
...
+       
+        # Check if all file paths exist
+        all_files_exist = all(isfile(metadata_path(metadatum)) for metadatum in metadata)
+
+        # If all files exist, print the message once
+        if all_files_exist
+            println("Note: ECCO $(metadata.name) data is in $(metadata.dir).")
+        else
+            println("Downloading ECCO data: $(metadata.name)")
+        end
```
glwagner, line 255: I like this message, but it seems like there is an assumption here. Should we put the downloading into the second branch so the message is guaranteed to be correct? In other words:

```julia
if not_downloading
    # say we aren't
else
    # say we are downloading
    # do the downloading
end
```

taimoorsohail: I'm not sure what assumption you mean? I am using exactly the same if statement as the statement which determines if downloading should occur. The reason I don't print these messages in the if loop where downloading happens is because the message then gets repeated by the numbers of years x months that are downloaded/not downloaded, i.e., you will always see 24 downloaded + not downloaded messages for two years, which feels a bit too verbose. 

Instead, the code checks if _all_ files are present in the specified directory, then, the code says either that files were downloaded or they were not downloaded.

glwagner: > I'm not sure what assumption you mean? I am using exactly the same if statement as the statement which determines if downloading should occur.

Sorry, to clarify, it is only an assumption if the two separate `if` statements continue to use the same criteria in the future. Since you haven't guaranteed that explicitly, future developments can cause this statement to become invalid. You could add a test about printing, but its hard to do that. So better to ensure that future developments don't break it through a slight change in the code.

taimoorsohail:
OK - I have attempted to remove the first conditional. I have instead created a missing_files Boolean which switches to true if there are any missing files (or whatever the condition is here). By default, it is false. I think this should resolve the double conditional assumption? @glwagner

src/DataWrangling/ECCO/ECCO_metadata.jl
taimoorsohail: I have added a modification that should account for this - only one conditional is used

glwagner (approved):
Great! Simone mostly touches this code so might be good for him to take a look @simone-silvestri

src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -262,10 +265,13 @@ function download_dataset(metadata::ECCOMetadata; url = urls(metadata))
...
                     throw(ArgumentError(msg))
                 end
-
+		println("Downloading ECCO data: $(metadatum.name)")
                 Downloads.download(fileurl, filepath; downloader, progress=download_progress)
-            end
+	    end
         end
+	if !missing_files
+            println("Note: ECCO $(metadata.name) data is in $(metadata.dir).")
```
simone-silvestri, line 273:
```suggestion
            @info "Note: ECCO $(metadata.name) data is in $(metadata.dir)."
```
we typically use `@info` for printed information (it just has a nicer formatting 😄 )

navidcy:
@taimoorsohail, I think the test fail because since they are coming from a fork, they don't have access to the repo's secrets that are needed to download the ECCO. I'll merge, but perhaps it worths opening PRs directly from the main repo.

We have the convention that branch names should start with the initials of who creates it, all lower case, and words separated with dashes, e.g., `ncc/a-branch-for-a-new-fancy-feature` or e.g. `ncc-ts/a-feature-we-are-implementing-together`.

taimoorsohail:
Thanks @navidcy I will follow that convention from now on and make changes directly from main

navidcy:
I ran it and simulation starts going!

PR #365, opened 2025-03-03
Continue sea-ice coupling effort

simone-silvestri:
This branch continues the coupling effort by splitting interfacial sea-ice ocean flux from the frazil ice formation. A coupled quarter-degree simulation runs. Still the results are not correct because sea ice concentration is not changing without ice dynamics so the sea ice extent is fixed from the initial conditions. 

However, one PR at then time, we can merge this, then fix sea ice


simone-silvestri:
This is an example of a simulation running with `time_step!(sea_ice)` on this PR. Meaning that at least the simulation does not break. I will merge this and continue in a PR that does an arctic simulation for a couple of years. 
Next steps are:
- figure out the correct thermodynamic timestepping
- add stability functions to the sea ice atmosphere fluxes
- add the sea-ice momentum equation

https://github.com/user-attachments/assets/bf65829e-db12-4d53-b6fd-507d55e17906

PR #374, opened 2025-03-05
Try to pick up the ocean model within `run!(::Simulation{<:OceanSeaIceModel})`

glwagner:
@taimoorsohail curious if this is all that is needed.

Still some work to do for proper checkpointing.

### DIFF
--- a/src/OceanSeaIceModels/ocean_sea_ice_model.jl
+++ b/src/OceanSeaIceModels/ocean_sea_ice_model.jl
@@ -28,6 +28,7 @@ struct OceanSeaIceModel{I, A, O, F, C} <: AbstractModel{Nothing}
 end
 
 const OSIM = OceanSeaIceModel
+const OSIMSIM = Simulation{<:OceanSeaIceModel}
 
 function Base.summary(model::OSIM)
     A = nameof(typeof(architecture(model)))
@@ -189,3 +190,8 @@ function above_freezing_ocean_temperature!(ocean, sea_ice::SeaIceSimulation)
     return nothing
 end
 
+# TODO: picking up OceanSeaIceModel simulations from a checkpoint is a WIP
+function set!(sim::OSIMSIM, pickup)
+    set!(sim.model.ocean, pickup)
+    return nothing
+end


taimoorsohail:
Do we also need:

```
set_time_stepper!(timestepper, args...)
```

to deal with variable tendencies? 

https://github.com/CliMA/Oceananigans.jl/blob/3fe5e9e66b027e38cfbb3ff53b456e85b16b3bfd/src/OutputWriters/checkpointer.jl#L148

glwagner:
Doesn't `set!(sim.model.ocean, pickup)` call that?

glwagner:
via

https://github.com/CliMA/Oceananigans.jl/blob/3fe5e9e66b027e38cfbb3ff53b456e85b16b3bfd/src/Simulations/run.jl#L60

and then

https://github.com/CliMA/Oceananigans.jl/blob/3fe5e9e66b027e38cfbb3ff53b456e85b16b3bfd/src/OutputWriters/checkpointer.jl#L244

taimoorsohail:
Yep you're right, I completely missed that... I'm testing it with my code to see if it works

glwagner:
We could add a test here: https://github.com/CliMA/ClimaOcean.jl/blob/main/test/test_simulations.jl

that file claims to do a "time stepping test" but as far as I can tell, does not actually take a time step!

taimoorsohail:
Question: If the set!() function changes the timestep of the ocean model only, doesn't the prescribed atmosphere and radiation become out-of-sync, as it will begin at (1993, 1, 1) regardless of whether a checkpoint is present or not? Whereas the ocean model will start from whatever month the checkpointer was saved at?

glwagner:
> Question: If the set!() function changes the timestep of the ocean model only, doesn't the prescribed atmosphere and radiation become out-of-sync, as it will begin at (1993, 1, 1) regardless of whether a checkpoint is present or not? Whereas the ocean model will start from whatever month the checkpointer was saved at?

This is definitely an issue and why we implemented https://github.com/CliMA/Oceananigans.jl/pull/4148

Next we need to use this feature in ClimaOcean.

I am still trying to understand whether we actually want to add the Checkpointer to the coupled simulation directly. I have to go look through the source code to see if that will work out of the box or whether more changes are needed.

Sorry lots of things are happening --- I am doing quite a few things at the same time so trying to stay afloat! Any help is very greatly appreciated.

glwagner:
There is also https://github.com/CliMA/ClimaOcean.jl/pull/355/files which is related, I think I will update that PR soon and work on making sure the correct time-step is taken

taimoorsohail:
Thanks Greg! Sorry I'm on week 3 of learning Julia, Oceananigans and ClimaOcean so still struggling with it all. Hopefully I'll be able to make a more meaningful contribution when I understand things a bit more! In the meantime, no rush on getting things changed - I will also try to debug on my end.

glwagner:
> Thanks Greg! Sorry I'm on week 3 of learning Julia, Oceananigans and ClimaOcean so still struggling with it all. Hopefully I'll be able to make a more meaningful contribution when I understand things a bit more! In the meantime, no rush on getting things changed - I will also try to debug on my end.

True! It could help if you work on a setup that doesn't require a checkpointer. Workflows that include checkpointing are quite slow, best for production...

navidcy:
> True! It could help if you work on a setup that doesn't require a checkpointer. Workflows that include checkpointing are quite slow, best for production...

I'm not sure what you mean here. How do you run anything longer that few hours if you are not able to restart it? Or if something happens and the internet/ssh connection breaks you should be able to restart it.

I'll try to add the functionality in ClimaOcean; it'll be needed for sure. Just haven't found the time yet :(

glwagner:
> > True! It could help if you work on a setup that doesn't require a checkpointer. Workflows that include checkpointing are quite slow, best for production...
> 
> I'm not sure what you mean here. How do you run anything longer that few hours if you are not able to restart it? Or if something happens and the internet/ssh connection breaks you should be able to restart it.
> 
> I'll try to add the functionality in ClimaOcean; it'll be needed for sure. Just haven't found the time yet :(

Use tmux to ensure sessions remain if connections are broken or to keep jobs running when you need to step away.

I'm only talking about prototyping, it's true that if you want to run for 100s of years you of course need to checkpoint. Just a few years can be done without it though.

navidcy:
> I'm only talking about prototyping, it's true that if you want to run for 100s of years you of course need to checkpoint. Just a few years can be done without it though.

Oh I see, gotcha. I was thinking longer runs, not playing around-prototyping.

navidcy:
Closing this in favor of #381

glwagner:
> > I'm only talking about prototyping, it's true that if you want to run for 100s of years you of course need to checkpoint. Just a few years can be done without it though.
> 
> Oh I see, gotcha. I was thinking longer runs, not playing around-prototyping.

I only dispute the implication that prototyping is playing around --- its not, its important.

navidcy:
Dash should be /. prototyping or playing around!  Sorry

PR #376, opened 2025-03-06
Allow iteration for surface fluxes on inactive cells with `FixedIterations`

glwagner:
This eliminates an `if` statement from the kernel when using `FixedIterations` which can be a major benefit for optimization.


simone-silvestri (approved):

glwagner:
error makes no sense because this PR doesn't touch that

navidcy:
what error?

glwagner:
> what error?

it went away when I re-ran it! well!

PR #383, opened 2025-03-07
(0.4.6) Remove thermodynamics step from ClimaOcean and use new ClimaSeaIce version

simone-silvestri:


glwagner (approved):

glwagner:
You may want to bump ClimaOcean patch version

PR #387, opened 2025-03-09
Add boundary conditions to the default fields in `PrescribedAtmosphere`

### LINKED ISSUE #386: Missing BCs in `PrescribedAtmosphere` fields
This is a problem when we interpolating the atmospheric state.

simone-silvestri:
closes #386


src/OceanSeaIceModels/PrescribedAtmospheres.jl
```diff
@@ -320,32 +321,37 @@ function Base.show(io::IO, pa::PrescribedAtmosphere)
...
+    bcs  = FieldBoundaryConditions(grid, (Center, Center, Nothing))
+    rain = FieldTimeSeries{Center, Center, Nothing}(grid, times; boundary_conditions=bcs)
+    snow = FieldTimeSeries{Center, Center, Nothing}(grid, times; boundary_conditions=bcs)
     return (; rain, snow)
 end
 
 function default_atmosphere_pressure(grid, times)
-    pa = FieldTimeSeries{Center, Center, Nothing}(grid, times)
+    bcs = FieldBoundaryConditions(grid, (Center, Center, Nothing))
+    pa  = FieldTimeSeries{Center, Center, Nothing}(grid, times; boundary_conditions=bcs)
```
glwagner, line 354: Why don't these boundary conditions get added already? This seems like a change that belongs in Oceananigans. What's the motivation for doing this here? You will have to do this every time you make a FieldTimeSeries right, isn't it a bit much?

navidcy: Should we change the default here?

https://github.com/CliMA/Oceananigans.jl/blob/a93dcb28fcca69b3f7867a0c3fc07dbf88136df5/src/OutputReaders/field_time_series.jl#L376

navidcy: I guess even if we do, we'd probably don't add defaults that have a Nothing z-location

navidcy: @simone-silvestri just for clarification: is this only needed for FieldTimeSeries that are restricted on the surface? Or you are saying that we need to include boundary conditions in the constructors of FieldTimeSeries when we plan to interpolate those time series; this happens to be one example of such case you stumbled upon?

glwagner: i'll do it

PR #388, opened 2025-03-09
Fixes docs that don't build on `main`

navidcy:
Fixes docs:

- Infer FT in `SimilarityTheoryFluxes` from the ocean grid


navidcy:
```Julia
julia> grid = LatitudeLongitudeGrid(size=(360*60, 180*60, 1), latitude=(-90, 90), longitude=(0, 360), z=(0, 1))
21600×10800×1 LatitudeLongitudeGrid{Float64, Periodic, Bounded, Bounded} on CPU with 3×3×1 halo and with precomputed metrics
├── longitude: Periodic λ ∈ [0.0, 360.0)  regularly spaced with Δλ=0.0166667
├── latitude:  Bounded  φ ∈ [-90.0, 90.0] regularly spaced with Δφ=0.0166667
└── z:         Bounded  z ∈ [0.0, 1.0]    regularly spaced with Δz=1.0

julia> minimum_xspacing(grid)
0.26954410758617847

julia> minimum_yspacing(grid)
1853.2487774093122
```

navidcy:
probably an Oceananigans bug....

navidcy:
Actually, I think it's correct!
The minimum xspacing just so small because we have a grid that reaches the pole singularities!

I'm wondering if the condition `if resxt < resxn || resyt < resyn ... interpolate ... end` is the right one. Will this ever be met? We need to have a target grid with resolution **finer** than the lat-lon x-spacing near the pole singularity, am I right?

navidcy:
Perhaps the criterion should be based on `minimum(λspacings(grid))` and `minimum(φspacings(grid))`?

glwagner (approved):

glwagner:
What criterion are you referring to? Can you link to the code?

navidcy:
sorry, here:


https://github.com/CliMA/ClimaOcean.jl/blob/1af700add4dfd3dbb0f7fa3bb19e554a3d5d1e09/src/Bathymetry.jl#L188-L205

glwagner:
> sorry, here:
> 
> https://github.com/CliMA/ClimaOcean.jl/blob/1af700add4dfd3dbb0f7fa3bb19e554a3d5d1e09/src/Bathymetry.jl#L188-L205
> 
> `resxn` is the xspacing at the pole singularity so the criterion `resxt < resxn` can never be satisfied?

What is the issue with the criterion not being satisfied?

navidcy:
We have a 1/60th bathymetry and build a 1 degree grid and then got the message that "Skipping passes for interpolating bathymetry.... because the grid you created is not coarser than the native grid."

That read very counter-intuitive to me. Perhaps the problem is the `@info` message?

glwagner:
> sorry, here:
> 
> https://github.com/CliMA/ClimaOcean.jl/blob/1af700add4dfd3dbb0f7fa3bb19e554a3d5d1e09/src/Bathymetry.jl#L188-L205
> 
> `resxn` is the xspacing at the pole singularity so the criterion `resxt < resxn` can never be satisfied?

I can't see the rest of the code. How does meeting the criteria change what happens (I can read the comment, but it doesn't give complete information whereas the code will).

More generally though I would say this code is assuming a uniform spacing. I don't know why this is necessary.

This is by the way a more general with a "fixed iteration" approach to refining the grid. It should be criteria-based; ie smooth / modify bathyemtry until all criteria are met.

glwagner:
> We have a bathymetry on a 1/60th native grid and then we build a 1 degree grid and then got the message along the lines of: "Skipping passes for interpolating bathymetry.... because the grid you created is not coarser than the bathymetry's native grid."
> 
> That read very counter-intuitive to me. Perhaps the problem is the `@info` message?

seems wrong. Do you want me to get rid of the check?

simone-silvestri (approved):

simone-silvestri:
I think there are also problems with the single column simulation (#399)

navidcy:
merge with #461

PR #390, opened 2025-03-09
(0.4.7) Update for latest oceananigans

glwagner:

Issue #389, opened 2025-03-09
Should we decouple download from regridding for bathymetry?


navidcy:
At the moment, `regrid_bathymetry` 

https://github.com/CliMA/ClimaOcean.jl/blob/38197234d2eb245ca0dfe2abfcb2c25ac821fce8/src/Bathymetry.jl#L97

both downloads the bathymetry dataset and regrids. Me and @taimoorsohail were envisioning something like where

https://github.com/CliMA/ClimaOcean.jl/blob/38197234d2eb245ca0dfe2abfcb2c25ac821fce8/src/Bathymetry.jl#L106-L113

are a separate `download_bathymetry` method.

glwagner:
Don't we already have that? What do you mean?

glwagner:
Why can't you just define

```julia
function download_bathymetry
 # No need for @root here, because only rank 0 accesses this function 
 if !isfile(filepath) 
     Downloads.download(url, filepath; progress=download_progress) 
 end
end 
```

and then call `download_bathymetry` inside `regrid_bathymetry`?

There is an old issue about improving the name of `regrid_bathymetry` too I think which has not been addressed.

glwagner:
But if there is a reason to "enforce" a workflow, I am up for it. I like to think of software development as building out features that people can use for a variety of purposes, rather than enforcing a workflow. It's quite trivial to define a function `download_bathymetry` that would be useful, and we don't have to change any existing tests or examples. Not having to change tests or examples makes it something that might take just a few minutes to implement. Would you like me to implement it?

If we have some motivation to actually discontinue support for downloading bathymetry inside `regrid_bathymetry` though, we can do that. I don't see it right away, because `regrid_bathymetry` already checks for an existing file. So if you choose to call `download_bathymetry` prior to `regrid_bathymetry`, it seems like you have effectively achieved downloading and regridding in separate functions.

PR #391, opened 2025-03-09
Add `download_bathymetry`

glwagner:
This PR adds `download_bathymetry()` method which can be either called separately or within `regrid_bathymetry()`.

cc @navidcy @taimoorsohail


src/Bathymetry.jl
```diff
@@ -36,6 +36,28 @@ end
...
+Download the bathymetry from `url` and saves it under `filename` in the directory `dir` and
+return the full filepath where the bathymetry is saved.
+"""
+function download_bathymetry(;
+                             url = etopo_url,
+                             dir = download_bathymetry_cache,
+                             filename = "ETOPO_2022_v1_60s_N90W180_surface.nc")
+
+    filepath = joinpath(dir, filename)
+    # No need for @root here, because only rank 0 accesses this function
```
navidcy, line 53: Is this comment pertinent?

```suggestion

```

glwagner: Uh. I think it's wrong. We need @root. This is just an assumption that isn't always true and I don't know why one would make this assumption or how it is enforced.

glwagner:
Is this what you wanted @navidcy ? This does not enforce the need to download bathymetry before regridding as suggested on #389. It just exposes a function that users can invoke to download the bathymetry.

navidcy:
That's actually better than what I was suggesting/implying in #389!

src/Bathymetry.jl
navidcy: Hm... Should we rephrase the comment? or leave it there for now..?

navidcy (approved):

src/Bathymetry.jl
glwagner: No we should add @ root.

navidcy:
btw test break because ClimaOcean now uses ClimaSeaIce 0.2.4 which implies that the 0.2.3 -> 0.2.4 was a breaking change and perhaps it should have been 0.2.3 -> 0.3.0?

we can ignore that because things should be OK after #390 is merged.

glwagner:
> btw test break because ClimaOcean now uses ClimaSeaIce 0.2.4 which implies that the 0.2.3 -> 0.2.4 was a breaking change and perhaps it should have been 0.2.3 -> 0.3.0?

We've defined "breaking" as "breaking the user interface", but this is probably not consistent with semantic versioning / packages meant to be used as libraries or that form the dependencies for other packages. We could update our reasoning

navidcy:
> > btw test break because ClimaOcean now uses ClimaSeaIce 0.2.4 which implies that the 0.2.3 -> 0.2.4 was a breaking change and perhaps it should have been 0.2.3 -> 0.3.0?
> 
> We've defined "breaking" as "breaking the user interface", but this is probably not consistent with semantic versioning / packages meant to be used as libraries or that form the dependencies for other packages. For example `AbstractModel` certainly is not exported but both ClimaSeaIce and ClimaOcean depend on it, so when we change it (like we just did) then both packages need to update. We could update our reasoning

Yes, sorry, perhaps I missed that definition.

It's not what we define, it has to do with semantic versioning and whether the compat entries for packages need to be updated for the package to work. E.g., compathelper won't suggest you change the compat entry to _not_ allow 0.2.4 when already 0.2.3 was OK. It assumes that 0.2.3 -> 0.2.4 just fixed bugs, didn't break anything.

navidcy:
Any idea why distributed tests fail?

https://buildkite.com/clima/climaocean-ci/builds/2930#01957dd8-0c80-46c1-a95b-2d1b666dfe79

Anything to do with `@root` we added?

glwagner:
best way to find out is to remove it

glwagner:
pretty weird to me that `root` cannot be composed though.

navidcy:
I removed it and everything broke :D (But that could be random CI failure thing...)

src/Bathymetry.jl
```diff
@@ -36,6 +37,27 @@ end
...
+return the full filepath where the bathymetry is saved.
+"""
+function download_bathymetry(; url = etopo_url,
+                             dir = download_bathymetry_cache,
+                             filename = "ETOPO_2022_v1_60s_N90W180_surface.nc")
+
+    filepath = joinpath(dir, filename)
+
+    if !isfile(filepath)
+        Dowwnloads.download(url, filepath; progress=download_progress)
```
navidcy, line 55:
```suggestion
        Downloads.download(url, filepath; progress=download_progress)
```

navidcy:
Removing `@root` made the distributed tests pass. I'm putting it back in and see if they fail again; if they do then I will leaning to conclude that it's not incidental.

glwagner:
Maybe we can't stack `@root` but that would be a bug it would seem...

navidcy:
What does "stack @root" mean? Is this a technical term or a different way to say "delete @root" from the download_bathymetry method?

taimoorsohail:
This (in part) addresses issue #344 also. Just need to do the same for JRA55.

navidcy:
without `@root` distributed tests were done in 3min; with `@root` they've already been running for 47min.

I'm leaning in dropping `@root` and open an issue

glwagner:
> without `@root` distributed tests were done in 3min; with `@root` they've already been running for 47min.
> 
> I'm leaning in dropping `@root` for this PR and open an issue?

def. I would have given up all ready (admirable persistence!)

simone-silvestri:
> What does "stack @root" mean? Is this a technical term or a different way to say "delete @root" from the download_bathymetry method?

We cannot stack root, because root has a `MPI.Barrier()` inside (ensuring all processes are synched coming out of root), so the processes will hang if we do something like
```julia
@root begin
     something...
     @root begin
         something else...
      end
end
```
because this would expand to something like
```julia
if rank == 0
        something
        if rank == 0
            something else...
        end
        Barrier()
end
Barrier()
```
so it means there is a barrier inside an if condition that not all processes can reach that would end up in a deadlock

PR #392, opened 2025-03-09
Always interpolate bathymetry regardless of grid spacing

glwagner:
Is this what you wanted @navidcy ?

### DIFF
--- a/src/Bathymetry.jl
+++ b/src/Bathymetry.jl
@@ -188,24 +188,6 @@ function interpolate_bathymetry_in_passes(native_z, target_grid;
     Nλt, Nφt = Nt = size(target_grid)
     Nλn, Nφn = Nn = size(native_z)
 
-    resxt = minimum_xspacing(target_grid)
-    resyt = minimum_yspacing(target_grid)
-
-    resxn = minimum_xspacing(native_z.grid)
-    resyn = minimum_yspacing(native_z.grid)
-
-    # Check whether we are refining the grid in any directions.
-    # If so, skip interpolation passes, as they are not needed.
-    if resxt < resxn || resyt < resyn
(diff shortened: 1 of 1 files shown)


navidcy:
Not really. I'm not sure actually.
One issue is that I'm not sure what was intended there.

But definitely getting a message that a 1-degree grid is not finer than the 1/60th-degree grid got me confused. I think that was just a bad criterion selection that was basing it to x-spacing from a lat-lon grid that went up to 90 degrees latitude.

navidcy:
x-ref: https://github.com/CliMA/ClimaOcean.jl/pull/388#issuecomment-2709089382

glwagner:
It would be interesting to know what was intended, but on the other hand aren't we sure the offending code is problematic? If there is some motivation to avoid interpolating under certain conditions, that could be implemented in another PR. But actually I would again suggest that we need criteria-based interpolation. I can put in a framework for criteria-based interpolation if you like. Then it would be easy to add a criteria like "don't keep interpolating if both grids are uniformly spaced and the target grid is finer than the source grid". That woudl just be one of many criteria that could be leveraged though.

navidcy:
OK, now I noticed that interpolation still happens further down.

So yes, I think that part of the code  you deleted is problematic. Or something else was actually intended to happen there that was either not communicated well (via the comments/code/message)? @simone-silvestri?

glwagner:
> OK, now I noticed that interpolation _still_ happens further down. So yes, I think that part of the code you deleted is problematic. Or something else was actually intended to happen there that was either not communicated well (via the comments/code/message)? @simone-silvestri?

you have to interpolate "passes" as being "additional interpolation". I think it's hard to understand so I opened a PR to switch to criteria-based which I think we want anyways.

PR #393, opened 2025-03-09
Interpolate bathymetry based on criteria, including combination of criteria

glwagner:
This PR really just demonstrates the simplicity of a framework that uses generic criteria to decide when to stop interpolating, rather than a simple for loop.

Could address @navidcy concerns but not sure.

Supercedes #392 

x-ref #388


glwagner:
Now you can imagine a new criteria based on the gradient of `z`, something like

```julia
struct MaximumGradient
    gradient :: Float64
end

gradient(z) = sqrt(∂x(∂x(z)) + ∂y(∂y(h)))
interpolating(mg::MaximumGradient, pass, z, args...) = maximum(gradient(z)) < mg.gradient
```

glwagner:
Although it's worth stating that I think a better method would actually apply a diffusion operator to the bathymetry. This would open up the possibility of using a diffusivity that's proportional to the slope so smoothing problem areas faster. Maybe interpolation effectively achieves this, not sure.

giordano:
Probably needs something like https://github.com/CliMA/ClimaSeaIce.jl/pull/59

navidcy:
yeap, #390 is dealing with that

PR #395, opened 2025-03-10
(0.5.0) remove the `OrthogonalSphericalShellGrids.jl` package

simone-silvestri:
To capture https://github.com/CliMA/Oceananigans.jl/pull/3913 when merged
Bump major version because it's a breaking change


simone-silvestri:
Looks like it works!

PR #396, opened 2025-03-11
Try adding a precompile workload

glwagner:
Working with @giordano, we noticed that `ocean_simulation` is fairly slow to compile but fast to execute once compiled. This makes it a good candidate for a precompile workload, so we are adding that in this PR.

I would also like to evaluate `OceanSeaIceModel`. I have found that precompiling doesn't seem to help as much but this should be investigated.


Project.toml
```diff
@@ -21,6 +21,7 @@ NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
 Oceananigans = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 OrthogonalSphericalShellGrids = "c2be9673-fb75-4747-82dc-aa2bb9f4aed0"
+PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
```
giordano, line 24: You'll need to add a compat bound as well 😉 "1" could be a decent initial guess

glwagner: ah ya

src/ClimaOcean.jl
```diff
@@ -96,5 +96,19 @@ using ClimaOcean.OceanSeaIceModels: PrescribedAtmosphere
...
 
+using PrecompileTools: @setup_workload, @compile_workload
+using OrthogonalSphericalShellGrids
+
+@setup_workload begin
+    Nx, Ny, Nz = 32, 32, 10
+    @compile_workload begin
+        z = exponential_z_faces(Nz=Nz, depth=6000, h=34)
+        grid = OrthogonalSphericalShellGrids.TripolarGrid(CPU(); size=(Nx, Ny, Nz), halo=(7, 7, 7), z)
+        grid = ImmersedBoundaryGrid(grid, GridFittedBottom((x, y) -> -5000))
```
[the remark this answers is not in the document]
glwagner: i dont like that one

simone-silvestri (approved):

navidcy:
I had noticed it was relatively slow as well. Does this speed things?

glwagner:
I think it will help with `ocean_simulation` but not sure about anything else

glwagner:
although together with https://github.com/CliMA/ClimaOcean.jl/pull/418 we might be much better off

PR #381, opened 2025-03-07
Checkpointer implementation for the coupled simulation

taimoorsohail:
This is an attempt to implement a checkpointer for the coupled simulation. The checkpointer should checkpoint all components of the coupled model (ocean, sea ice, atmosphere, radiation) required to restart the coupled simulation.

Work in progress still with @navidcy 

Potentially superseding #374.

PR #401, opened 2025-03-11
Implementing checkpointer for coupled model

taimoorsohail:
This PR supersedes #381 adding a new run_coupled! function that replaces the run! function in Oceananigans. 

WIP with @navidcy

### DIFF
--- a/src/OceanSeaIceModels/ocean_sea_ice_model.jl
+++ b/src/OceanSeaIceModels/ocean_sea_ice_model.jl
@@ -1,9 +1,11 @@
+export run_coupled!
+
 using Oceananigans
+using Oceananigans.OutputWriters: checkpoint_path
 using Oceananigans.TimeSteppers: Clock
 using Oceananigans: SeawaterBuoyancy
 using ClimaSeaIce.SeaIceThermodynamics: melting_temperature
 using KernelAbstractions: @kernel, @index
-
 using SeawaterPolynomials: TEOS10EquationOfState
 
 import Thermodynamics as AtmosphericThermodynamics  
@@ -14,7 +16,7 @@ import Oceananigans.Architectures: architecture
 import Oceananigans.Fields: set!
 import Oceananigans.Models: timestepper, NaNChecker, default_nan_checker
 import Oceananigans.OutputWriters: default_included_properties
-import Oceananigans.Simulations: reset!, initialize!, iteration
+import Oceananigans.Simulations: reset!, initialize!, iteration, run!
 import Oceananigans.TimeSteppers: time_step!, update_state!, time
 import Oceananigans.Utils: prettytime
 import Oceananigans.Models: timestepper, NaNChecker, default_nan_checker
@@ -29,6 +31,20 @@ struct OceanSeaIceModel{I, A, O, F, C, Arch} <: AbstractModel{Nothing, Arch}
 end
 
 const OSIM = OceanSeaIceModel
+const OSIMSIM = Simulation{<:OceanSeaIceModel}
+
+function run_coupled!(sim::OSIMSIM; pickup)    
+    if pickup
+        checkpoint_file_path = checkpoint_path(pickup, sim.model.ocean.output_writers)
+        set!(sim.model.ocean, checkpoint_file_path)
+        # Setting the atmosphere time to the ocean time
+        sim.model.atmosphere.clock.time = sim.model.ocean.model.clock.time
+    end
+
+    @show sim.model.clock.time
+    run!(sim)
+end
+
 
 function Base.summary(model::OSIM)
     A = nameof(typeof(architecture(model)))
@@ -158,6 +174,16 @@ function default_nan_checker(model::OceanSeaIceModel)
     u_ocean = model.ocean.model.velocities.u
     nan_checker = NaNChecker((; u_ocean))
     return nan_checker
+end
+
+# TODO: picking up OceanSeaIceModel simulations from a checkpoint is a WIP
+function set!(sim::OSIMSIM, pickup::Union{Bool, Integer, String})
+    checkpoint_file_path = checkpoint_path(pickup, sim.model.ocean.output_writers)
+    set!(sim.model.ocean, filepath = checkpoint_file_path)
+    set!(sim.model.ocean, pickup = pickup)
+    # Setting the atmosphere time to the ocean time
+    sim.model.atmosphere.clock.time = sim.model.ocean.model.clock.time
+    return nothing
 end
 
 @kernel function _above_freezing_ocean_temperature!(T, grid, S, ℵ, liquidus)
--- a/examples/checkpointer_mwe.jl
+++ b/examples/checkpointer_mwe.jl
@@ -1,0 +1,107 @@
+using ClimaOcean
+using Oceananigans
+using Oceananigans.Units
+using CFTime
+using Dates
+using Printf
+
+arch = CPU()
+
+Nx = 144
+Ny = 60
+Nz = 40
+
+depth = 6000meters
+z_faces = exponential_z_faces(; Nz, depth)
+
+grid = LatitudeLongitudeGrid(arch;
+                             size = (Nx, Ny, Nz),
+                             halo = (7, 7, 7),
+                             z = z_faces,
+                             latitude  = (-75, 75),
+                             longitude = (0, 360))
+
+ocean = ocean_simulation(grid)
+
+# date = DateTimeProlepticGregorian(1993, 1, 1)
+# set!(ocean.model, T=ECCOMetadata(:temperature; dates=date),
+#                   S=ECCOMetadata(:salinity; dates=date))
+
+radiation = Radiation(arch)
+
+atmosphere = JRA55PrescribedAtmosphere(arch; backend=JRA55NetCDFBackend(41))
+
+coupled_model = OceanSeaIceModel(ocean; atmosphere, radiation)
+
+simulation = Simulation(coupled_model; Δt=10, stop_iteration=10)
+
+wall_time = Ref(time_ns())
+
+function progress(sim)
+    ocean = sim.model.ocean
+    u, v, w = ocean.model.velocities
+    T = ocean.model.tracers.T
+
+    Tmax = maximum(interior(T))
+    Tmin = minimum(interior(T))
+
+    umax = (maximum(abs, interior(u)),
+            maximum(abs, interior(v)),
+            maximum(abs, interior(w)))
+
+    step_time = 1e-9 * (time_ns() - wall_time[])
+
+    msg = @sprintf("Iter: %d, simulation time: %s, atmosphere time: %s, Δt: %s", iteration(sim), prettytime(sim), prettytime(atmosphere.clock.time), prettytime(sim.Δt))
+    msg *= @sprintf(", max|u|: (%.2e, %.2e, %.2e) m s⁻¹, extrema(T): (%.2f, %.2f) ᵒC, wall time: %s",
+                    umax..., Tmax, Tmin, prettytime(step_time))
+
+    @info msg
+
+    wall_time[] = time_ns()
+end
+
+simulation.callbacks[:progress] = Callback(progress, IterationInterval(1))
+
+outputs = merge(ocean.model.tracers, ocean.model.velocities)
+
+ocean.output_writers[:surface] = JLD2OutputWriter(ocean.model, outputs;
+                                                  schedule = IterationInterval(2),
+                                                  filename = "checkpointer_mwe_surface",
+                                                  indices = (:, :, grid.Nz),
+                                                  with_halos = true,
+                                                  overwrite_existing = true,
+                                                  array_type = Array{Float32})
+
+output_dir = "."
+prefix = "checkpointer_mwe"
+
+ocean.output_writers[:checkpoint] = Checkpointer(ocean.model;
+                                                 schedule = IterationInterval(2),
+                                                 prefix = prefix,
+                                                 dir = output_dir,
+                                                 verbose = true,
+                                                 overwrite_existing = true)
+
+# coupled_checkpointer = Checkpointer(coupled_model;
+#                                     schedule = IterationInterval(4),
+#                                     prefix = prefix,
+#                                     dir = output_dir,
+#                                     verbose = true,
+#                                     overwrite_existing = true)
+
+# @show simulation
+
+run_coupled!(simulation, pickup=true)
+
+# @info "simulation run for 10 iterations; you should have a checkpointer at 8"
+
+# checkpoint_file = prefix * "_iteration0.jld2"
+
+# set!(simulation, checkpoint_file)
+
+# coupled_model = OceanSeaIceModel(simulation.model.ocean; atmosphere, radiation)
+
+# simulation = Simulation(coupled_model; Δt=10, stop_iteration=20)
+# simulation.callbacks[:progress] = Callback(progress, IterationInterval(1))
+
+# run!(simulation)
--- a/src/ClimaOcean.jl
+++ b/src/ClimaOcean.jl
@@ -32,6 +32,7 @@ export
     ocean_simulation,
     sea_ice_simulation,
     initialize!,
+    run_coupled!,
     @root, 
     @onrank,
     @distribute,


glwagner:
I don't agree with the design that creates a new function called `run_coupled!`. However, one could extend `run!` for a `Simulation` of `OceanSeaIceModel`. I'm not sure if there is a reason to keep both versions of `run!`.

navidcy:
> I don't agree with the design that creates a new function called `run_coupled!`. However, one could extend `run!` for a `Simulation` of `OceanSeaIceModel`. I'm not sure if there is a reason to keep both versions of `run!`.

Yes, we also agree. We don't like that either!
We were playing around with extending `run!` but couldn't figure out a way to differentiate the methods...

Somehow this:

```julia
function run!(simulation::Simulation{<:OceanSeaIceSimulation}`; pickup)
    if pickup
        ... do extra things to pick up a coupled model
    end

    run!(simulation) # Oceananigans.Simulations.run!
    return nothing
end
```
wasn't working for us yesterday. 

Don't worry too much atm, it's mostly playing around. When we settle or if we get stuck we can ping you.

taimoorsohail:
Linking issue #404 to this PR

glwagner:
> > I don't agree with the design that creates a new function called `run_coupled!`. However, one could extend `run!` for a `Simulation` of `OceanSeaIceModel`. I'm not sure if there is a reason to keep both versions of `run!`.
> 
> Yes, we also agree. We don't like that either! We were playing around with extending `run!` but couldn't figure out a way to differentiate the methods...
> 
> Somehow this:
> 
> ```julia
> function run!(simulation::Simulation{<:OceanSeaIceSimulation}`; pickup)
>     if pickup
>         ... do extra things to pick up a coupled model
>     end
> 
>     run!(simulation) # Oceananigans.Simulations.run!
>     return nothing
> end
> ```
> 
> wasn't working for us yesterday.
> 
> Don't worry too much atm, it's mostly playing around. When we settle or if we get stuck we can ping you.

To implement this, you need to expose an interface for "picking up a simulation", which has a default implementation in Oceananigans, but which can be specialized for the type of Simulation. Then you can specialize for a simulation of `OceanSeaIceModel` here.

navidcy:
After https://github.com/CliMA/ClimaOcean.jl/pull/401/commits/9ab0df109e3cb05aef99f2f44d98b8246857faf9 I can run the checkpointer_mwe.jl with Oceananigans 0.95.26 (not the branch from https://github.com/CliMA/Oceananigans.jl/pull/4216)!

examples/near_global_ocean_simulation.jl
navidcy, line 1: need to undo these changes if not intentional

src/OceanSeaIceModels/ocean_sea_ice_model.jl
```diff
@@ -73,6 +76,26 @@ function initialize!(model::OSIM)
     return nothing
 end
 
+initialize_jld2_file!(filepath, init, jld2_kw, including, outputs, model::OSIM) =
+    initialize_jld2_file!(filepath, init, jld2_kw, including, outputs, model.ocean.model)
+
+write_output!(c::Checkpointer, model::OSIM) = write_output!(c, model.ocean.model)
+
+function set!(sim::OSIMSIM, pickup::Union{Bool, Integer, String})
```
navidcy, line 84: @glwagner should we type this so that it only works for OSIMSIM with PrescribedAtmosphere?

navidcy: [4f5ff2e](https://github.com/CliMA/ClimaOcean.jl/pull/401/commits/4f5ff2e6ef6ae38e71ef1112316fc552205ce32f)

navidcy:
@glwagner can you review? if you are happy we can convert the `checkpointer_mwe.jl` into a test and we are good to go I think

navidcy:
cc @NoraLoose

examples/near_global_ocean_simulation.jl
taimoorsohail: Done - testing Checkpointer now

src/OceanSeaIceModels/ocean_sea_ice_model.jl
```diff
@@ -59,17 +61,33 @@ prettytime(model::OSIM)             = prettytime(model.clock.time)
...
+    checkpoint_file_path = checkpoint_path(pickup, sim.output_writers)
+
+    set!(sim.model.ocean.model, checkpoint_file_path)
+
+    sim.model.clock.iteration = sim.model.ocean.model.clock.iteration
+    sim.model.clock.time = sim.model.ocean.model.clock.time
+
+    # Setting the atmosphere time to the ocean time
+    sim.model.atmosphere.clock.iteration = sim.model.ocean.model.clock.iteration
+    sim.model.atmosphere.clock.time = sim.model.ocean.model.clock.time
```
glwagner, line 89: Rather than extending set! For every atmos model, it might be better to write a function for setting the clock and then extend that appropriately

glwagner: Same comment for ocean and sea ice

glwagner: As in:

set_clock!(atmos, time, iter)

navidcy: Good point regarding defining a `set_clock!` method.

Just slightly confused: this doesn't alleviate the need to extend `set!`, right?

taimoorsohail: I don't think so @navidcy - currently the set! function in Oceananigans does `set!(sim.model, checkpoint_file_path)` when we want it to do `set!(sim.model.ocean.model, checkpoint_file_path)` because of the coupled nature of the model.

navidcy: so you are saying that we can do this _without_ extending `set!`?

glwagner: Sounds like you need to define `set!(::OceanSeaIceMode, filepath)`.

taimoorsohail: No, I think we need to extend `set!` and create the function `set_clock!` as @glwagner suggests. See my latest push which has added that function.

glwagner: One problem with `set!(::OSIMSIM, filepath)` is that, when we eventually support checkpointing diagnostics  / callbacks / output writer states, that support will have to be copy/pasted here. So it'd be better to use `set!(::OceanSeaIceModel, filepath)`.

test/test_ocean_sea_ice_model.jl
```diff
@@ -93,3 +93,49 @@ using ClimaSeaIce.SeaIceThermodynamics: melting_temperature
...
+        run!(simulation, pickup=true)
+
+        # create a new simulation and pick up from the last checkpointer
+        new_simulation = testbed_coupled_simulation(grid; stop_iteration=13)
+
+        run!(new_simulation, pickup=true)
+
+        # test that ocean tiem step and iteration is the same as the atmosphere
+        @test     new_simulation.model.atmosphere.clock.iteration ≈ new_simulation.model.ocean.model.clock.iteration
+        @test     new_simulation.model.atmosphere.clock.time ≈ new_simulation.model.ocean.model.clock.time
```
navidcy, line 139:
```suggestion
        # ensure the ocean and atmosphere time step and iteration are the same
        @test new_simulation.model.atmosphere.clock.iteration ≈ new_simulation.model.ocean.model.clock.iteration
        @test new_simulation.model.atmosphere.clock.time ≈ new_simulation.model.ocean.model.clock.time
```

examples/one_degree_simulation.jl
```diff
@@ -170,19 +170,32 @@ ocean.output_writers[:surface] = JLD2OutputWriter(ocean.model, outputs;
...
+
+directory_checkpointer = "."
+prefix_checkpointer = "one_degree_checkpointer"
+
+simulation.output_writers[:checkpoint] = Checkpointer(ocean.model;
+                                                      schedule = TimeInterval(5days),
+                                                      prefix = prefix_checkpointer,
+                                                      dir = directory_checkpointer,
+                                                      verbose = true,
+                                                      overwrite_existing = true)
```
navidcy, line 184:
```suggestion
# We also save a series of Checkpointers that will enable us to restart the simulation
# from a later point; ideal for longer runs.

simulation.output_writers[:checkpoint] = Checkpointer(ocean.model;
                                                      schedule = TimeInterval(5days),
                                                      prefix = "one_degree_checkpointer",
                                                      overwrite_existing = true)
```

examples/one_degree_simulation.jl
```diff
@@ -170,19 +170,32 @@ ocean.output_writers[:surface] = JLD2OutputWriter(ocean.model, outputs;
...
+
 # ### Ready to run
 
 # We are ready to press the big red button and run the simulation.
 
 # After we run for a short time (here we set up the simulation with `stop_time = 10days`),
 # we increase the timestep and run for longer.
 
-run!(simulation)
+run!(simulation, pickup=true)
```
navidcy, line 193: not sure if we should have the `pickup=true` as the default in the example or explain, e.g., by something along the lines

> If you have run the simulation and you there are checkpointer files produces, next time when you run again you can use `run!(simulation, pickup=true)` to restart from the latest checkpointer.

src/OceanSeaIceModels/ocean_sea_ice_model.jl
```diff
@@ -59,17 +61,37 @@ prettytime(model::OSIM)             = prettytime(model.clock.time)
...
+initialize!(model::OSIM) = initialize!(model.ocean)
+
+initialize_jld2_file!(filepath, init, jld2_kw, including, outputs, model::OSIM) =
+    initialize_jld2_file!(filepath, init, jld2_kw, including, outputs, model.ocean.model)
+
+write_output!(c::Checkpointer, model::OSIM) = write_output!(c, model.ocean.model)
 
-function reset!(model::OSIM)
-    reset!(model.ocean)
+function set_clock!(sim)
```
navidcy, line 79: this _only_ works for coupled simulation

navidcy: [f9087aa](https://github.com/CliMA/ClimaOcean.jl/pull/401/commits/f9087aad4437c4d8322a5d5a3d292a5a63f0b8ec) creates a `set_clock!` for each simulation

examples/one_degree_simulation.jl
glwagner: `cleanup = true` ?

examples/one_degree_simulation.jl
glwagner: I agree, I don't think we should put `pickup=true`.

The idea is that we only need `pickup=true` on the second time we run the script. So it's not appropriate to put it in the script permanently.

src/OceanSeaIceModels/ocean_sea_ice_model.jl
```diff
@@ -76,21 +76,26 @@ initialize_jld2_file!(filepath, init, jld2_kw, including, outputs, model::OSIM)
...
     return nothing
 end
 
 function set!(sim::OSIMSIMPA, pickup::Union{Bool, Integer, String})
     checkpoint_file_path = checkpoint_path(pickup, sim.output_writers)
 
     set!(sim.model.ocean.model, checkpoint_file_path)
-    set_clock!(sim)
+
+    iteration, time = sim.model.ocean.model.clock.iteration, sim.model.ocean.model.clock.time
```
glwagner, line 96: we will want `set_clock` for the ocean simulation too at some point

test/test_ocean_sea_ice_model.jl
```diff
@@ -93,3 +93,49 @@ using ClimaSeaIce.SeaIceThermodynamics: melting_temperature
     end
 end
 
+function testbed_coupled_simulation(grid; stop_iteration=8)
```
navidcy, line 96:
```suggestion
"""
    testbed_coupled_simulation(grid; stop_iteration=8)

Create a test-bed coupled simulation with a Checkpointer.
"""
function testbed_coupled_simulation(grid; stop_iteration=8)
```

src/OceanSeaIceModels/ocean_sea_ice_model.jl
navidcy: yes, should we add now?

glwagner:
Just want to write that @simone-silvestri and I are discussing how to generalize `OceanSeaIceModel` so that it is truly a coupled model without a "principle" component. This will require changing code that assumes the ocean model is the "source of truth". Just an FYI that if it is possible to avoid this assumption in the checkpointing code, it would be nice. Basically there should be symmetry among all the models, eventually. And we would want a `set_clock!` for the ocean too (which may in fact belong in Oceananigans)

glwagner:
Does `set_clock!` also need the previous time-step to be correct? Otherwise stateful time-steppers like AB2 will not be picked up correctly.

examples/one_degree_simulation.jl
taimoorsohail: But `pickup=true` simply informs you there are no checkpointers and initialises from scratch if no checkpointer files are found. Having `pickup=true` permanently ensures that the checkpointer functionality will always work (including when there are no checkpointers). Otherwise the user would have to manually go in and change from `run!(simulation)` to `run!(simulation, pickup=true)` - something that is inconvenient if you have submitted the job to automatically restart in an HPC scheduler.

src/OceanSeaIceModels/ocean_sea_ice_model.jl
navidcy: OK. But given then the output writers/checkpointers are included in the outer-outer simulations, then if we define `set!(::OSIM, filepath)` that won't have access to the checkpointer. Or? Am I missing something?

navidcy: Otherwise, I like the suggestion!

examples/one_degree_simulation.jl
navidcy: Agreed.

taimoorsohail: `pickup=true` already deals well with the case where there are no checkpointers, so I don't see why it needs to be removed :D

navidcy: I think it boils down to your different workflows (and as a matter of fact to the workflow the user will be using).

Do you submit job via slurm/qsub?
Do you run scripts on repl?

glwagner: It doesn't make sense. We should just have `pickup=true` as the default if we put it here.

glwagner: > `pickup=true` already deals well with the case where there are no checkpointers, so I don't see why it needs to be removed :D

it makes the assumption that you actually want to pick up, rather than tweak the simulation and re-run which is what you want to do most of the time. This will encourage bad practices / confuse people.

src/OceanSeaIceModels/ocean_sea_ice_model.jl
navidcy:
```
set!(::OceanSeaIceModel, filepath)
```
done via [4110aca](https://github.com/CliMA/ClimaOcean.jl/pull/401/commits/4110aca3ce4c0f6adf767d9fdaebd0be825504ef)

src/OceanSeaIceModels/PrescribedAtmospheres.jl
```diff
@@ -318,6 +321,36 @@ function Base.show(io::IO, pa::PrescribedAtmosphere)
...
+        checkpointed_clock = file["$addr/clock"]
+
+        # Update model clock
+        set_clock!(model, checkpointed_clock)
+    end
+
+    return nothing
+end
+
+checkpointer_address(::PrescribedAtmosphere) = "HydrostaticFreeSurfaceModel"
```
navidcy, line 338: Ideally we'd like to make the checkpointer also save the clock for the atmosphere and all other models and then read it everything off from the corresponding checkpointer address.

navidcy: I think that's for a future PR..

navidcy: I don't know, I'm apprehensive... I hate this
```
checkpointer_address(::PrescribedAtmosphere) = "HydrostaticFreeSurfaceModel"
```
 because it's a hack. But on the other hand I like the symmetry here:

https://github.com/CliMA/ClimaOcean.jl/blob/9f9ca89561211d535926dd4570d362b677eab6dc/src/OceanSeaIceModels/ocean_sea_ice_model.jl#L98-L108

Essentially the hack tells the checkpointer for the atmosphere that it should go read data from the ocean model checkpointer.

navidcy:
OK, I tried to make the checkpointer not assuming that ocean model is special but I think it's a lot of work. Perhaps we can put it off until later? :(

I was thinking to  define a checkpointer on the `OceanSeaIceModel` and then redefine `write_output!` and all other methods so that the checkpointer goes and writes all the required properties of all the `OceanSeaIceModel` and all its model components (eg atmosphere, ocean, etc). Then one need not assume that ocean is special.

```

**Comment by navidcy** [comment]
I don't know what's the best workflow to do this because it requires some tweaking of Oceananigans to expose some of the functionality of the Checkpointer which then has to be extended. But the precompilation + requirement of tweaking a different package makes my dev workflow very slow and I feel a bit worn out.

Shall we put this off for later after the above-mentioned redesign of the OceanSeaIceModel @glwagner?

File: src/OceanSeaIceModels/PrescribedAtmospheres.jl (hunk 9 above)
Comment (taimoorsohail) [reply src/OceanSeaIceModels/PrescribedAtmospheres.jl:338]: I guess this sets the basis for future code development so in that sense is OK?

**Comment by taimoorsohail** [comment]
> Does `set_clock!` also need the previous time-step to be correct? Otherwise stateful time-steppers like AB2 will not be picked up correctly.

Just checking

> Does `set_clock!` also need the previous time-step to be correct? Otherwise stateful time-steppers like AB2 will not be picked up correctly.

Does the `set_clock!` change [4110aca3ce4c0f6adf767d9fdaebd0be825504ef](https://github.com/CliMA/ClimaOcean.jl/pull/401/commits/4110aca3ce4c0f6adf767d9fdaebd0be825504ef) resolve this @navidcy ?

File: src/OceanSeaIceModels/PrescribedAtmospheres.jl (hunk 9 above)
Comment (glwagner) [reply src/OceanSeaIceModels/PrescribedAtmospheres.jl:338]: why do you have to write 

```julia
checkpointer_address(::PrescribedAtmosphere) = "HydrostaticFreeSurfaceModel"
```

navidcy: I’m changing that
That’s part of the hack I was explaining yesterday I don’t like.

navidcy:
Just note that I’m refactoring this. It’s almost there. But haven’t pushed yet.

navidcy:
x-ref https://github.com/CliMA/Oceananigans.jl/pull/4250

navidcy:
I plan to work on this on weekend/early next week

simone-silvestri:
@taimoorsohail can we close this waiting for Oceananigans v 0.104 that has new checkpointing?

PR #403, opened 2025-03-11
Remove distributed macros from ClimaOcean

simone-silvestri:
probably requires we wait for a bump in Oceananigans after [#4209](https://github.com/CliMA/Oceananigans.jl/pull/4209)


navidcy (approved):
next Oceananigans patch/minor release we can run the CI and if passes it's all good!

PR #405, opened 2025-03-12
Remove need for `solver_stop_criteria` in a flux formulation

simone-silvestri:
this PR pushes the check whether to iterate or not inside the `compute_interface_state` function, so that we can come up with other non iterative flux solvers that do not need to include the `solver_stop_criteria` field


glwagner:
Can you explain the design in a bit more detail? The problem is that neither SimilarityTheory nor CoefficientBased need require iteration --- it depends specifically on some details of their formulations. So I'm not sure this design is achieving its stated goal. The modification that seems to be needed first is to remove solver criteria from CoefficientBased / SimilarityTheory and put somewhere else.

simone-silvestri:
I ll close this as stale

PR #408, opened 2025-03-12
Add more information for the `RepeatYearJRA55()` dataset, including date range correspondence to JRA55

### LINKED ISSUE #407: RepeatYear forcing for JRA55-do reanalysis is 1st May 1990 to 30th April 1991
Based on the analysis by [Stewart et al. 2021](https://doi.org/10.1016/j.ocemod.2019.101557), the Repeat-Year forcing is 1st May 1990 to 30th April 1991. Is there a reason why this is not the default, e.g., here:

https://github.com/CliMA/ClimaOcean.jl/blob/c7ce9523cf182e1de0578b32da1bfbdf79732304/src/DataWrangling/JRA55/JRA55_metadata.jl#L30-L36

I propose we change that.

navidcy:
Closes #407


glwagner:
Check the NetCDF files to ensure this is right

navidcy:
I'll add a test

glwagner:
I don't think its right FYI

navidcy:
Yes, I'll check that; don't worry.

glwagner:
I believe the correct year is 1990, to match with the initial condition / ECCO state

navidcy:
This are the netCDF files I got from the Australian HPC, right?

glwagner:
> This are the netCDF files I got from the Australian HPC, right?

I don't know, is there a way to tell?

glwagner:
There was a comment somewhere which explained that even though the repeat year dates start/end in spring, the data was delivered to span January 1 to Dec 31.

But now maybe I understand the problem, for Jan 1 to mar 1 this should be 1991 not 1990.

indont think we can resolve this properly until Oceananigans supports DateTime but not sure.

navidcy:
Oh now I remember. Then perhaps it's appropriate to add a note in the PrescribedAtmosphere/JRA55RepeatYear docstring. Even for us to remember this (eg I had forgotten).

navidcy:
I changed the PR. Now it just adds a note in the docstring. @glwagner/@simone-silvestri  could you have another look?

src/DataWrangling/JRA55/JRA55_field_time_series.jl
```diff
@@ -145,11 +145,16 @@ new_backend(::JRA55NetCDFBackend, start, length) = JRA55NetCDFBackend(start, len
...
+which describes one of the variables in the "repeat-year forcing" dataset derived
 from the Japanese 55-year atmospheric reanalysis for driving ocean-sea ice models (JRA55-do).
 For more information about the derivation of the repeat-year forcing dataset, see
 
-> Stewart et al. (2020). JRA55-do-based repeat year forcing datasets for driving ocean–sea-ice models, _Ocean Modelling_, **147**, 101557, https://doi.org/10.1016/j.ocemod.2019.101557.
+> Stewart et al. (2020). JRA55-do-based repeat-year forcing datasets for driving ocean–sea-ice models, _Ocean Modelling_, **147**, 101557, https://doi.org/10.1016/j.ocemod.2019.101557.
+
+The repeat year used here corresponds to May 1st, 1990 until April 30th, 1991. However, the
+returned dataset here has dates that range from January 1st to December 31st. This implies
+that the first 4 months of the `JRA55RepeatYear()` dataset are from year 1991 from the JRE55
```
simone-silvestri, line 156:
```suggestion
that the first 4 months of the `JRA55RepeatYear()` dataset are from year 1991 from the JRA55
```

simone-silvestri (approved):

PR #410, opened 2025-03-13
Bugfix: time_step!(model, dt)

### LINKED ISSUE #404: Output writers seems not saving anything?
In multiple examples, the `Checkpointer` function isn't saving anything beyond iteration 0. A modified `near_global_ocean` example that demonstrates this is in my branch and PR #401 https://github.com/CliMA/ClimaOcean.jl/blob/55106babf63e92dfb4b7b31f542eb3c7c921bdd1/examples/near_global_ocean_simulation.jl#L170

But I have been able to recreate this across two examples with both `TimeInterval` and `IterationInterval` schedules. @navidcy

glwagner:
possibly this closes #404


taimoorsohail (approved):

test/test_ocean_sea_ice_model_parameter_space.jl
navidcy, line 1: @simone-silvestri this test was never called so we deleted it; if it's important for some reason we can add it again. Just tagging you here in case you are wondering in the future where did this go...

simone-silvestri: no problem, I guess the parameter space is checked in the GPU version of `test_ocean_sea_ice_model.jl`

PR #411, opened 2025-03-13
Fuse kernels for sea_ice_ocean fluxes computation

simone-silvestri:
just a small cleanup that removes code and improves kernel fusion


src/OceanSeaIceModels/InterfaceComputations/sea_ice_ocean_fluxes.jl
```diff
@@ -47,10 +45,14 @@ end
...
 
     Nz  = size(grid, 3)
-    Qᶠₒ = frazil_heat_flux
-    Qᵢₒ = interface_heat_flux
+    Qᶠₒ = sea_ice_ocean_fluxes.frazil_heat
+    Qᵢₒ = sea_ice_ocean_fluxes.interface_heat
+    Jˢ  = sea_ice_ocean_fluxes.salt
     Tₒ  = ocean_temperature
     Sₒ  = ocean_salinity
+    Sᵢ  = ice_salinity
```
glwagner, line 53: Does this approach of fusing the kernels work when there's no ice salinity? that was the assumption i was trying to avoid by separating them (eg a bit more design would be needed). I guess we can make this assumption, but it's def good to avoid overcreating dependencies by making assumptions

glwagner (approved):
Just be careful about making assumptions about the properties of component models.

PR #412, opened 2025-03-13
Add a prescribed ocean to the src

simone-silvestri:
This PR adds the `PrescribedOceanModel` to the src (shuffling around a couple of methods). This prescribed ocean is useful to perform single-column simulations of a sea ice model sandwiched between a prescribed atmosphere and a prescribed ocean, to verify the performance of the fluxes and the sea ice climatology


src/OceanSimulations/prescribed_ocean_model.jl
```diff
@@ -0,0 +1,96 @@
...
+import Oceananigans.Models: timestepper, update_model_field_time_series!
+
+import ClimaOcean: reference_density, heat_capacity
+import Oceananigans.Architectures: on_architecture
+
+#####
+##### A prescribed ocean...
+#####
+
+struct PrescribedOceanModel{G, C, U, T, F, Arch} <: AbstractModel{Nothing, Arch}
```
glwagner, line 14: We call it `PrescribedAtmosphere`. Is this `PrescribedOcean` or `PrescribedOceanModel`?

I'm not sure really sure what a "prescribed model" is.

simone-silvestri: Hmm, this is because we add a simulation on top, while the atmosphere is a "simulation" rather than a model. I can switch it to a simulation rather than a model

src/OceanSimulations/prescribed_ocean_model.jl
```diff
@@ -0,0 +1,96 @@
...
+                throw(ArgumentError("All variables in the `timeseries` argument must be `FieldTimeSeries`"))
+            v.grid == grid ||
+                throw(ArgumentError("All variables in the timeseries reside on the provided grid"))
+        end
+    end
+
+    τx = Field{Face, Center, Nothing}(grid)
+    τy = Field{Center, Face, Nothing}(grid)
+    Jᵀ = Field{Center, Center, Nothing}(grid)
+    Jˢ = Field{Center, Center, Nothing}(grid)
```
glwagner, line 56: why this? I think this is wrong

src/OceanSimulations/prescribed_ocean_model.jl
```diff
@@ -0,0 +1,96 @@
...
+
+"""
+    PrescribedOceanModel(timeseries=NamedTuple(); grid, clock=Clock{Float64}(time = 0))
+
+Create a prescribed ocean model to be used in combination with ClimaOcean's `OceanSeaIceModel` 
+on a `grid` with a `clock`.
+
+Arguments
+=========
+- `timeseries`: A `NamedTuple` containing time series data for various fields. The named tuple can be empty
```
glwagner, line 32: the name is confusing because when we say "timeseries" it means just one series. But this is many timeseries?

PR #415, opened 2025-03-13
Bugfix in forcing barotropic potential

simone-silvestri:


glwagner (approved):

glwagner:
Might be good to add a test

PR #418, opened 2025-03-14
Make many big objects mutable

glwagner:
This should / may improve compile time. Note, the only objects that need to be immutable are the ones that must be passed _into_ kernels. Otherwise mutable is probably better.


glwagner:
neat

PR #420, opened 2025-03-14
Compute time interpolator inside time interpolation kernel

glwagner:


glwagner:
requires https://github.com/EnzymeAD/Reactant.jl/pull/918 for Reactant to work

src/OceanSeaIceModels/InterfaceComputations/interpolate_atmospheric_state.jl
```diff
@@ -137,7 +140,7 @@ end
     fj = get_fractional_index(i, j, jj)
 
     x_itp = FractionalIndices(fi, fj, nothing)
-    t_itp = time_interpolator
+    t_itp = TimeInterpolator(time_indexing, atmos_times, time)
```
glwagner, line 143: It seems better to compute this outside the kernel. Either way we have to pass `TracedRNumber` in. With the pattern implemented here, there is one `TracedRNumber`, `time`.

With the previous patten, `TimeInterpolator` has `TracedRNumber`s inside.

@wsmoses is there a reason why a single `TracedRNumber` (plus some recomputation as here) is preferred versus passing in a struct that has a few `TracedRNumber`?

PR #421, opened 2025-03-15
[CI] Some improvements to GitHub Actions setup

giordano:
* Install Dependabot
* Trigger on pull requests, pushes to `main` and tags, when editing relevant files
* Add concurrency group to cancel subsequent jobs in PRs. This is particularly useful to cancelling quickly macOS jobs, since there are only 5 concurrent runners available for the entire organisation
* Use `julia-actions/cache` instead of `actions/cache`
* Add timeouts, to avoid hanging jobs hold on to the runners for 6 hours
* Actually use the `matrix.arch` argument to install the desired julia build. At the moment that was completely unused, resulting in duplicate jobs

Similar to https://github.com/CliMA/Oceananigans.jl/pull/4198.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,13 +1,29 @@
 name: CI
+
 on:
-  push:
-    branches: [main]
-    tags: ["*"]
   pull_request:
+    paths:
+      - '.github/workflows/ci.yml'
+      - 'ext/**'
+      - 'src/**'
+      - 'test/**'
+      - 'Project.toml'
+  push:
+    branches:
+      - main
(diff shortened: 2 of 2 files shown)


navidcy (approved):

giordano:
Ah, tests are failing because I opened the PR from a fork and the `pull_request` trigger doesn't have access to secrets. There are two options:

* I keep the same trigger but reopen the PR from a branch here. This would be safer, but in general this means that tests will never work for PRs opened from forks (this was already the case before this PR)
* I change the trigger type to [`pull_request_target`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target), which has access to secrets.

Let me know what's your preference.

glwagner:
Is there a way to allow jobs to run on approval from someone? I'm ok with option 2 though. A little risk is healthy now and again 😈

giordano:
> Is there a way to allow jobs to run on approval from someone?

There's a setting to require approval for all PRs, but that also slows down a lot the development cycle

glwagner:
> > Is there a way to allow jobs to run on approval from someone?
> 
> There's a setting to require approval for all PRs, but that also slows down a lot the development cycle

Agree. Let's do option 2

giordano:
Ok, CI doesn't seem to want to start in this PR (pushing and/or closing-reopening doesn't have an effect).  I'll try with a new PR.

PR #422, opened 2025-03-15
[CI] Some improvements to GitHub Actions setup

giordano:
* Install Dependabot
* Trigger on pull requests, pushes to `main` and tags, when editing relevant files
* Add concurrency group to cancel subsequent jobs in PRs. This is particularly useful to cancelling quickly macOS jobs, since there are only 5 concurrent runners available for the entire organisation
* Use `julia-actions/cache` instead of `actions/cache`
* Add timeouts, to avoid hanging jobs hold on to the runners for 6 hours
* Actually use the `matrix.arch` argument to install the desired julia build. At the moment that was completely unused, resulting in duplicate jobs

Similar to https://github.com/CliMA/Oceananigans.jl/pull/4198.

New attempt at #421.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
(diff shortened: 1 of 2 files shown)


giordano:
Well, if anyone has an idea of why CI is now stuck, please let me know, I'm clueless.

PR #424, opened 2025-03-16
Possibly simplify interpolate_atmosphere_state kernel

glwagner:


giordano:
Let me try again without this PR just as a sanity check, but with it it was looking good after the last commit which fixed the issues I mentioned above.

giordano:
One thing is that I don't quite remember what this was needed for :sweat_smile:

PR #426, opened 2025-03-16
Precompile a bit less?

glwagner:
This PR changes so we only precompile `ocean_simulation` and not `OceanSeaIceModel`. I do think that precompiling `ocean_simulation` is more important but I don't have an idea of how much it helps to precompile `OceanSeaIceModel`.

### DIFF
--- a/src/ClimaOcean.jl
+++ b/src/ClimaOcean.jl
@@ -100,7 +100,6 @@ using PrecompileTools: @setup_workload, @compile_workload
         grid = Oceananigans.OrthogonalSphericalShellGrids.TripolarGrid(CPU(); size=(Nx, Ny, Nz), halo=(7, 7, 7), z)
         grid = ImmersedBoundaryGrid(grid, GridFittedBottom((x, y) -> -5000))
         ocean = ocean_simulation(grid)
-        model = OceanSeaIceModel(ocean)
     end
 end
 


giordano:
> but I don't have an idea of how much it helps to precompile OceanSeaIceModel.

You can compare

```julia
@time @eval OceanSeaIceModel(ocean)
```

with and without this PR, the total wall time and the % of compile time will tell you how much you're saving (and maybe also compare with v0.5.1, i.e. without #396) and then you can judge whether the extra step in the precompile workload (with the extra time spent in precompilation) is worth it.

giordano:
For what is worth, [with ClimaOcean v0.5.1](https://github.com/PRONTOLab/GB-25/actions/runs/13863692540/job/38799909016?pr=40#step:7:266) (without #396):
```
  10026.1 ms  ✓ ClimaOcean
 345080.2 ms  ✓ GordonBell25
```
With [ClimaOcean v0.5.3](https://github.com/PRONTOLab/GB-25/actions/runs/13880412187/job/38838211640?pr=40#step:7:284):
```
 460531.4 ms  ✓ ClimaOcean
 281000.4 ms  ✓ GordonBell25
```
The total precompilation time across the two packages increased, but having precompiled something in ClimaOcean already does seem to have made life easier in downstream packages (which means, we _might_ be able to precompile less). These are the kind of numbers you want to look at to understand whether precompilation workload is useful.

navidcy:
Somehow this https://github.com/CliMA/ClimaOcean.jl/pull/396 has made my development workflow slower... 
not sure if I don't have the optimal workflow..

I was thinking that precompilation would happen once and that would save me lots of time for later... but seems that when I edit the source code somehow precompilation is triggered again (either in a new session or sometimes within a session?) and that takes a lot of time.

Am I possibly doing something wrong? Any ideas?

giordano:
> I was thinking that precompilation would happen once and that would save me lots of time for later

If you modify the source code of the package the package has to be precompiled again

> Any ideas?

https://julialang.github.io/PrecompileTools.jl/stable/#Package-developers:-reducing-the-cost-of-precompilation-during-development

glwagner:
Here's what I get. I ran this code:

```julia
using ClimaOcean
using Oceananigans

Nx, Ny, Nz = 32, 32, 10
z = exponential_z_faces(Nz=Nz, depth=6000, h=34)
grid = Oceananigans.OrthogonalSphericalShellGrids.TripolarGrid(CPU(); size=(Nx, Ny, Nz), halo=(7, 7, 7), z)
grid = ImmersedBoundaryGrid(grid, GridFittedBottom((x, y) -> -5000))
ocean = ocean_simulation(grid)

@time @eval OceanSeaIceModel(ocean)
```

On this PR I got:

```
 19.606008 seconds (463.38 k allocations: 29.839 MiB, 0.04% gc time, 100.00% compilation time)
```

On main I get

```
  0.000318 seconds (462 allocations: 152.734 KiB)
```

seems like its helping

glwagner:
the trickier question is whether people want to use ClimaOcean _without_ using `OceanSeaIceModel`.

Right now, I think its the core of the package and yes, we do.

glwagner:
@navidcy the advice is to use this:

```julia
using ClimaOcean, Preferences
set_preferences!(ClimaOcean, "precompile_workload" => false; force=true)
```

giordano:
> the trickier question is whether people want to use ClimaOcean without using OceanSeaIceModel.

Yes whether the extra precompilation time is worth it for the end users becomes a judgement call.  Of course the alternative is just to make the life easier for the compiler, but that's not always feasible :sweat_smile:

glwagner:
But also one advantage here is that 19s we pay for compilation can be paid off in combination with precompile for other packages, so this really does reduce the visibility of compilation for a user (just not a dev)

glwagner:
I'll close this but we can keep discussing

navidcy:
> @navidcy the advice is to use this:
> 
> ```julia
> using ClimaOcean, Preferences
> set_preferences!(ClimaOcean, "precompile_workload" => false; force=true)
> ```

so I should have Preferences.jl installed in my main Julia environment, right?

giordano:
Not at all, only the LocalPreferences.toml file in your environment. Can also be the top level one of the stack (e.g. the one in you `.julia/environments/v1.X` directory).

giordano:
@navidcy out of curiosity, what machine do you use for development?

PR #433, opened 2025-03-20
Implement atmosphere sea ice fluxes

simone-silvestri:
This PR implements stability functions for atmosphere sea ice fluxes.

It also corrects a couple of solver implementations, in particular, it makes sure that the skin temperature is capped by a melting temperature and stops the iterations when this melting temperature is surpassed.

Also, it multiplies the concentration to fluxes that require it.

In addition, the aim is to correct the sea-ice ocean flux (at the moment is just a placeholder) to make sure results are sensible

The implementation is verified against a simple MITgcm configuration with a ocean with a y temperature gradient and an atmosphere with a x temperature gradient (comparison incoming)


src/OceanSeaIceModels/InterfaceComputations/compute_interface_state.jl
```diff
@@ -17,11 +18,31 @@ end
...
+    maxiter = convergence.maxiter
+    tolerance = convergence.tolerance
+    hasnt_started = iteration == 0
+    reached_maxiter = iteration ≥ maxiter
+    drift = abs(Ψⁿ.u★ - Ψ⁻.u★) + abs(Ψⁿ.θ★ - Ψ⁻.θ★) + abs(Ψⁿ.q★ - Ψ⁻.q★)
+    Tₘ = ℙᵢ.liquidus.freshwater_melting_temperature
+    Tₘ = convert_to_kelvin(ℙᵢ.temperature_units, Tₘ)
+    above_melting = Ψⁿ.T ≥ Tₘ
+    converged = drift < tolerance
+    return !(converged | reached_maxiter | above_melting) | hasnt_started
```
glwagner, line 44: Does this make sense? The intermediate state shouldn't be assumed to be physically meaningful. The temperature may be intermediate above melting but not converged?

This could cause problems with initialziation too because you can't escape from an above freezing initial condition.

glwagner: I also think this would prevent a situation where the surface melting temperature is a function of some state variable, like salinity. Changes in salinity, which change the melting temperature, would cause the solver to fail.

src/OceanSeaIceModels/InterfaceComputations/compute_interface_state.jl
```diff
@@ -17,11 +18,31 @@ end
...
-@inline iterating(Ψⁿ, Ψ⁻, iteration, fixed::FixedIterations) = iteration < fixed.iterations
+@inline iterating(Ψⁿ, Ψ⁻, ℙᵢ, iteration, fixed::FixedIterations) = iteration < fixed.iterations
+
+# Stop if the surface temperature is above freezing
+struct SeaIceConvergenceStopCriteria{FT}
+    tolerance :: FT
+    maxiter :: Int
+end
+
+@inline function iterating(Ψⁿ, Ψ⁻, ℙᵢ, iteration, convergence::SeaIceConvergenceStopCriteria)
```
glwagner, line 34: I think this is a bad design. We should not have a "sea ice specific convergence criteria". It's not really logical. Keep convergence criteria as actual convergence criteria and figure out another way to express the behavior you want to implement.

src/OceanSeaIceModels/InterfaceComputations/compute_interface_state.jl
glwagner: It's also not very logical. "Temperature above melting" has nothing to do with "convergence". I think that expresses the basic flaw in this

simone-silvestri: Without this change the solver will not converge because melting temperature is not in equilibrium with the fluxes. The non equilibrium part of the flux is used to melt sea ice which would not melt otherwise.

simone-silvestri: > It's also not very logical. "Temperature above melting" has nothing to do with "convergence". I think that expresses the basic flaw in this

I don't understand what you mean by this, I don't think it is relevant with this change. The solver is trying to find a temperature which is in equilibrium with incoming fluxes. In certain situations this temperature is _above_ melting. This condition cannot be found by the solver if we continue manually capping the temperature. In this case (since we are in a non-equilibrium condition) we need to stop the solver.

src/OceanSeaIceModels/InterfaceComputations/compute_interface_state.jl
simone-silvestri: What do you propose?

simone-silvestri:
Have been already implemented in #453

PR #437, opened 2025-03-21
Add support for multi-year JRA55

simone-silvestri:
This PR adds a multiyear JRA55. 
TODOs:

- [x] figure out the `set!` function
- [x] cleanup the jra55_metadata file
- [x] add some tests


simone-silvestri:
This should be ready to merge when tests pass. I am already downloading the whole JRA55 dataset through the new `ClimaOcean` implementation. The testing is a bit bare-bones but we can improve on it once we consolidate the use of the `JRA55MultipleYears` dataset

src/DataWrangling/JRA55/JRA55_field_time_series.jl
```diff
@@ -69,44 +70,60 @@ function compute_bounding_indices(longitude, latitude, grid, LX, LY, λc, φc)
...
+#   - ds["time"]:     time coordinate 
+#   - ds["lon"]:      longitude at the location of the variable
+#   - ds["lat"]:      latitude at the location of the variable
+#   - ds["lon_bnds"]: bounding longitudes between which variables are averaged
+#   - ds["lat_bnds"]: bounding latitudes between which variables are averaged
+#   - ds[shortname]:  the variable data
 
-    ds = Dataset(path)
+# Simple case, only one file per variable, no need to deal with multiple files
+function set!(fts::JRA55NetCDFFTSRepeatYear)
```
navidcy, line 108: what does this method do? `set!(fts)` without second argument? What does it set it to?

simone-silvestri: all the information to set the field are contained on the backend and in the time-indices in memory. So basically this command is setting it from stored data

simone-silvestri: It's called here https://github.com/CliMA/Oceananigans.jl/blob/fe5f37b2cfd0ff3ed6e882b1e0496c9546cef991/src/OutputReaders/field_time_series_indexing.jl#L335

navidcy: Gotcha!

test/test_jra55.jl
```diff
@@ -129,5 +129,22 @@ using ClimaOcean.OceanSeaIceModels: PrescribedAtmosphere
...
+        dataset = JRA55MultipleYears()
+        dates = ClimaOcean.DataWrangling.all_dates(JRA55MultipleYears(), :temperature)
+        
+        # These dates correspond to a metadata that crosses between year 1958 and 1959.
+        # Therefore, this will download two files for the two years and concatenate them
+        # when reading the data.
+        start_date = dates[2800]
+        end_date   = dates[3600]
+        backend = JRA55NetCDFBackend(10)
+        Ta = JRA55FieldTimeSeries(:temperature; dataset=JRA55MultipleYears(), start_date, end_date, backend)
```
navidcy, line 143: when we create a `JRA55FieldTimeSeries` with `JRA55MultipleYears()` the time indexing seems to be `Cyclical()`? Does this mean that if we run the simulation longer than the range of the dataset then it will loop from the start? Should we keep this as default?

test/test_jra55.jl
```diff
@@ -129,5 +129,22 @@ using ClimaOcean.OceanSeaIceModels: PrescribedAtmosphere
...
+
+        # Test we can access all the data
+        for t in Ta.times
+            @test Ta[t] isa Field
+        end
```
navidcy, line 148: how does this work for you?

I created `Ta` and got 

```Julia
julia> Ta
640×320×1×802 FieldTimeSeries{JRA55NetCDFBackend} located at (Center, Center, ⋅) of tas at /Users/navid/.julia/scratchspaces/0376089a-ecfe-4b0e-a64f-9c555d74d754/JRA55/tas_input4MIPs_atmosphericState_OMIP_MRI-JRA55-do-1-5-0_gr_195801010000-195812312100.nc
├── grid: 640×320×1 LatitudeLongitudeGrid{Float32, Periodic, Bounded, Flat} on CPU with 3×3×0 halo and with precomputed metrics
├── indices: (:, :, :)
├── time_indexing: Cyclical(period=8.6616e6)
├── backend: JRA55NetCDFBackend(1, 10)
├── path: /Users/navid/.julia/scratchspaces/0376089a-ecfe-4b0e-a64f-9c555d74d754/JRA55/tas_input4MIPs_atmosphericState_OMIP_MRI-JRA55-do-1-5-0_gr_195801010000-195812312100.nc
├── name: tas
└── data: 646×326×1×10 OffsetArray(::Array{Float32, 4}, -2:643, -2:323, 1:1, 1:10) with eltype Float32 with indices -2:643×-2:323×1:1×1:10
    └── max=331.728, min=0.0, mean=274.511

julia> Ta.times
3.02292e7:10800.0:3.888e7
```

thus how do we index with `3.02292e7`?

src/DataWrangling/JRA55/JRA55_field_time_series.jl
glwagner: It might be cleaner or more explicit to have a second argument to this function, for example, a new backend or index range that is being set to.

test/test_jra55.jl
glwagner: I like that as a default

navidcy: Minor: Perhaps just spit out an info when this happens?
Imagine somebody starts a simulation in 2020 and they run for what they think is 4 years but during that time they hit the end of the dataset and they go back to 1958 without them knowing

navidcy: Since we are at it, I never understood what the eg 10 in `JRA55NetCDFBackend(10)` does. Is it how many time-slices of the dataset are loaded into memory?

test/test_jra55.jl
simone-silvestri: Oops this is a bug, it should be `eachindex(Ta.times)`

test/test_jra55.jl
simone-silvestri: yep, it's the number of indices stored in memory, the rest is on disk

glwagner (approved):

navidcy:
Changing `JRA55RepeatYear` -> `RepeatYearJRA55` is a breaking change, right?

test/test_jra55.jl
navidcy: Thanks!

glwagner:
for sure, also is the repeatyear name introduced here?

navidcy:
It wasn't introduced here. It exists already. But after the discussion this PR renamed it from `JRA55RepeatYear()` -> `RepeatYearJRA55()`.

navidcy (approved):
happy to be merged when tests pass
I bumped the minor release since JRA55RepeatYear was renamed

navidcy:
wooo!!!

did you understand in the end why the tests were failing?

simone-silvestri:
yeah I was trying to load the whole `JRA55PrescribedAtmosphere` in memory 😅 . Probably we should change the default of `JRA55PrescribedAtmosphere` to a `JRA55NetCDFBackend`

glwagner:
Did you change the default here? Sounds like an easy thing to change now that will help others get started sooner

simone-silvestri:
Yeah I guess I did! This PR fixes it #458

PR #438, opened 2025-03-22
Add a Reactant Extension

glwagner:
There wiill likely be a few methods that need special treatment.

cc @giordano @wsmoses @simone-silvestri

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -178,9 +178,11 @@ steps:
     commands:
       - "srun julia --project -e 'using Pkg; Pkg.test()'"
     agents:
-      slurm_mem: 10G
-      slurm_cpus_per_task: 1
-      slurm_ntasks: 4
+      slurm_mem: 8G
+      slurm_gpus: 1
+      slurm_cpus_per_task: 8
+      slurm_ntasks: 1
+      slurm_gpus_per_task: 1
 
   - wait: ~
     continue_on_failure: true
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -163,7 +163,7 @@ steps:
     env:
       TEST_GROUP: "reactant"
       GPU_TEST: "false"
-  command: |
+    command: |
       touch LocalPreferences.toml
 
       echo "[Reactant]" >> LocalPreferences.toml
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -163,15 +163,8 @@ steps:
     env:
       TEST_GROUP: "reactant"
       GPU_TEST: "false"
-    command: |
-      touch LocalPreferences.toml
-
-      echo "[Reactant]" >> LocalPreferences.toml
-      echo "xla_runtime = \"IFRT\"" >> LocalPreferences.toml
-
-      cat LocalPreferences.toml
-
-      srun julia --project -e 'using Pkg; Pkg.test()'
+    commands:
+      - "srun julia --project -e 'using Pkg; Pkg.test()'"
     agents:
       slurm_mem: 10G
       slurm_cpus_per_task: 1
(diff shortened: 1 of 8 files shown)


.buildkite/pipeline.yml
```diff
@@ -158,5 +158,44 @@ steps:
       slurm_cpus_per_task: 1
       slurm_ntasks: 4
 
+  - label: "CPU Reactant tests"
+    key: "cpu_reactant_tests"
+    env:
+      TEST_GROUP: "reactant"
+      GPU_TEST: "false"
+  commands: |
```
glwagner, line 166:
```suggestion
  command: |
```

.buildkite/pipeline.yml
```diff
@@ -158,5 +158,44 @@ steps:
       slurm_cpus_per_task: 1
       slurm_ntasks: 4
 
+  - label: "CPU Reactant tests"
+    key: "cpu_reactant_tests"
+    env:
+      TEST_GROUP: "reactant"
+      GPU_TEST: "false"
+  command: |
```
glwagner, line 166:
```suggestion
    command: |
```

src/OceanSeaIceModels/InterfaceComputations/atmosphere_ocean_fluxes.jl
```diff
@@ -108,7 +108,7 @@ end
     downwelling_radiation = (; Qs, Qℓ)
 
     # Estimate initial interface state
-    FT = eltype(grid)
+    FT = typeof(Tᵢ)
```
navidcy, line 111: what's the difference here?

navidcy (approved):

navidcy (approved):

src/OceanSeaIceModels/InterfaceComputations/atmosphere_ocean_fluxes.jl
glwagner: Basically `Ti` is the correct choice here, because it is the type of the variables in `InterfaceState` (typically a floating point). The eltype of the grid, on the other hand, is more general, because it can either be a traced floating point or a constant floating point. The status of the grid eltype might change in the future (eg maybe we will refine the design so that it is always a floating point type). But regardless of how we decide to treat `eltype`, I think this line is expressing the correct pattern.

glwagner: Ok here is a better explanation:

`Ti` is the `eltype` of the arrays. This `eltype` is more general than just a floating point, because in order to transpile Oceananigans/ClimaOcean with Reactant, the eltype of arrays must be "marked" (also known as "traced") as something that can change while the computation is running. This distinguishes the array eltype from the eltype of variables which do _not_ change during a computation (these we refer to as "constants").

Prior to Reactant we did not distinguish between data that can change vs stay the same. Or actually I would say that we more generally simply allow anything at all to change through extensive use of mutable struct. But with Reactant we come into the position of having to specifically state what is changed and what is not.

We can decide to _define_ eltype(grid) as the "underlying" floating point type of an array eltype. I think this might be a valid choice.

However, we retain generality (and in a sense more modularity in the code) if we propagate the array eltypes locally, rather than attempting to always infer the "global eltype" from the grid. The local inference pattern is actually a better / more modular design.

I didn't realize any of this until Reactant came around, so I think this is a welcome improvement to realize that local inference is the correct approach. There are a handful of changes sprinkled into Oceananigans that are similar btw.

test/test_reactant.jl
```diff
@@ -0,0 +1,10 @@
+using Test
+using Reactant
+
+if haskey(ENV, "GPU_TEST")
```
giordano, line 4:
```suggestion
if get(ENV, "GPU_TEST", "false") == "true"
```

giordano:
I'm working on fixing the issue on Windows.

PR #440, opened 2025-03-24
Add parameter for Int in FixedIterations

glwagner:
@giordano


glwagner:
> Can you please also bump the patch version number so we can use it straight away?

yes! I also have the extension in line so was hoping to get them both under the patch if tests finish on time

PR #444, opened 2025-03-26
Separate gravitational_acceleration from flux_formulation

simone-silvestri:
We need it to compute air temperature so it does not really fit in the flux_formulation, but in a more upstream data structure. I have thought to put it directly in `ComponentInterfaces` and then pass it to `atmosphere_properties` inside the compute fluxes function, but I am happy to shift it around somewhere else.

This PR is needed for #433

### DIFF
--- a/src/OceanSeaIceModels/InterfaceComputations/atmosphere_ocean_fluxes.jl
+++ b/src/OceanSeaIceModels/InterfaceComputations/atmosphere_ocean_fluxes.jl
@@ -51,7 +51,8 @@ function compute_atmosphere_ocean_fluxes!(coupled_model)
             atmosphere_data,
             interface_properties,
             atmosphere_properties,
-            ocean_properties)
+            ocean_properties,
+            coupled_model.interfaces.gravitational_acceleration)
 
     return nothing
 end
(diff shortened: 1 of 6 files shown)


src/OceanSeaIceModels/InterfaceComputations/atmosphere_ocean_fluxes.jl
```diff
@@ -36,7 +36,8 @@ function compute_atmosphere_ocean_fluxes!(coupled_model)
     interface_properties = coupled_model.interfaces.atmosphere_ocean_interface.properties
     ocean_properties = coupled_model.interfaces.ocean_properties
     atmosphere_properties = (thermodynamics_parameters = thermodynamics_parameters(atmosphere),
-                             surface_layer_height = surface_layer_height(atmosphere))
+                             surface_layer_height = surface_layer_height(atmosphere),
+                             gravitational_acceleration = coupled_model.interfaces.gravitational_acceleration)
```
glwagner, line 40: we could consider `interface.properties`. Or maybe we should sketch out a larger vision for this.

simone-silvestri: we could start with `interface.properties`.

glwagner: or `OceanSeaIceModel.properties` perhaps, which maybe makes more logical sense (even though the properties are _only_ needed for coupling)

glwagner: Actually `OceanSeaIceModel.interfaces` is mostly an organization principle, since we don't have the concept of a model that doesn't have interfaces

PR #447, opened 2025-03-28
Generalize the concept of `StateExchanger` to allow different types

simone-silvestri:
so for example if we have a particular prognostic simulation we can create a new `StateExchanger` type and extend the `state_exchanger` function to return that particular type.

This is a minimally invasive PR, in the future when we will probably want to generalize this concept more

### DIFF
--- a/src/OceanSeaIceModels/InterfaceComputations/component_interfaces.jl
+++ b/src/OceanSeaIceModels/InterfaceComputations/component_interfaces.jl
@@ -385,7 +385,7 @@ function ComponentInterfaces(atmosphere, ocean, sea_ice=nothing;
                   sea_ice_top    = net_top_sea_ice_fluxes,
                   sea_ice_bottom = net_bottom_sea_ice_fluxes)
 
-    exchanger = StateExchanger(ocean, atmosphere)
+    exchanger = state_exchanger(ocean, atmosphere)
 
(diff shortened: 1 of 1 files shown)


simone-silvestri:
Actually, this is probably not needed, we can just extend the constructor of `StateExchanger`...

src/OceanSeaIceModels/InterfaceComputations/component_interfaces.jl
```diff
@@ -397,6 +397,9 @@ function ComponentInterfaces(atmosphere, ocean, sea_ice=nothing;
                                net_fluxes)
 end
 
+# To extend if we have a prognostic atmosphere
+state_exchanger(ocean, atmosphere) = StateExchanger(ocean, atmosphere)
```
glwagner, line 401: I had a different design --- ther eis an "`atmosphere_state_exchanger`" constructor. This is more composible, ie so you can mix and match different components. 

In other words we don't want to extend `StateExchanger`. We want to generalize `StateExchanger` to full generality, and then allow each component to interact with `StateExchanger` in their own way, with:

* `atmosphere_state_exchanger`
* `ocean_state_exchanger`
* `land_state_exchanger`
* `sea_ice_state_exchanger`

The `StateExchanger` is then the thing that is used to compute fluxes, ie on the exchange grid.

PR #452, opened 2025-03-31
Switch metadata type parameters

simone-silvestri:
This PR switches the type parameter of `Metadata` from `Metadata{D, V}` to `Metadata{V, D}` where `D` is the dates type and `V` is the dataset type

it makes more sense to have the `dataset` being the first parameter in the `Metadata` type since it is required much more than the dates to dispatch.

This PR is propedeutic to #437

### DIFF
--- a/src/DataWrangling/metadata.jl
+++ b/src/DataWrangling/metadata.jl
@@ -2,7 +2,7 @@ using CFTime
 using Dates
 using Base: @propagate_inbounds
 
-struct Metadata{D, V}
+struct Metadata{V, D}
     name  :: Symbol
     dates :: D
     dataset :: V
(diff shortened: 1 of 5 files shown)


src/DataWrangling/metadata.jl
```diff
@@ -2,7 +2,7 @@ using CFTime
 using Dates
 using Base: @propagate_inbounds
 
-struct Metadata{D, V}
+struct Metadata{V, D}
     name  :: Symbol
     dates :: D
     dataset :: V
```
glwagner, line 8:
```suggestion
    dataset :: V
    dates :: D
```

you might as well also do this to reduce the possibility of confusion

glwagner (approved):

PR #453, opened 2025-04-01
Implement sea-ice ocean stress

simone-silvestri:
this PR adds sea-ice ocean stresses to the sea-ice ocean interface

### DIFF
--- a/src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
+++ b/src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
@@ -76,8 +76,10 @@ end
     i, j = @index(Global, NTuple)
     kᴺ = size(grid, 3)
     time = Time(clock.time)
-    ρτx = atmos_ocean_fluxes.x_momentum  # zonal momentum flux                      
-    ρτy = atmos_ocean_fluxes.y_momentum  # meridional momentum flux
+    ρτxao = atmos_ocean_fluxes.x_momentum # atmosphere - ocean zonal momentum flux                      
+    ρτyao = atmos_ocean_fluxes.y_momentum # atmosphere - ocean meridional momentum flux
+    ρτxio = sea_ice_ocean_fluxes.x_momentum # sea_ice - ocean zonal momentum flux                      
+    ρτyio = sea_ice_ocean_fluxes.y_momentum # sea_ice - ocean meridional momentum flux
 
     @inbounds begin
         Sₒ = ocean_salinity[i, j, kᴺ]
@@ -121,20 +123,22 @@ end
     ρₒ⁻¹ = 1 / ocean_properties.reference_density
     cₒ   = ocean_properties.heat_capacity
 
-    τxao = ℑxᶠᵃᵃ(i, j, 1, grid, τᶜᶜᶜ, ρₒ⁻¹, ℵ, ρτx)
-    τyao = ℑyᵃᶠᵃ(i, j, 1, grid, τᶜᶜᶜ, ρₒ⁻¹, ℵ, ρτy)
-    Jᵀao = ΣQao  * ρₒ⁻¹ / cₒ
-    Jˢao = - Sₒ * ΣFao
-
-    ρₒ⁻¹ = 1 / ocean_properties.reference_density
-    cₒ   = ocean_properties.heat_capacity
-
     @inbounds begin
         ℵᵢ   = ℵ[i, j, 1]
         Qio  = sea_ice_ocean_fluxes.interface_heat[i, j, 1]
-        Jˢio = sea_ice_ocean_fluxes.salt[i, j, 1]
-        Jᵀio = Qio * ρₒ⁻¹ / cₒ
 
+        Jᵀao = ΣQao  * ρₒ⁻¹ / cₒ
+        Jˢao = - Sₒ * ΣFao
+        Jᵀio = Qio * ρₒ⁻¹ / cₒ
+        Jˢio = sea_ice_ocean_fluxes.salt[i, j, 1]
+
+        τxao = ℑxᶠᵃᵃ(i, j, 1, grid, τᶜᶜᶜ, ρₒ⁻¹, ℵ, ρτxao)
+        τyao = ℑyᵃᶠᵃ(i, j, 1, grid, τᶜᶜᶜ, ρₒ⁻¹, ℵ, ρτyao)
+        τxio = ρτxio[i, j, 1] * ρₒ⁻¹ * ℑxᶠᵃᵃ(i, j, 1, grid, ℵ)
+        τyio = ρτyio[i, j, 1] * ρₒ⁻¹ * ℑyᵃᶠᵃ(i, j, 1, grid, ℵ)
+
+        τx[i, j, 1] = τxao + τxio
+        τy[i, j, 1] = τyao + τyio
         τx[i, j, 1] = τxao
         τy[i, j, 1] = τyao
         Jᵀ[i, j, 1] = (1 - ℵᵢ) * Jᵀao + Jᵀio
--- a/src/OceanSeaIceModels/InterfaceComputations/sea_ice_ocean_fluxes.jl
+++ b/src/OceanSeaIceModels/InterfaceComputations/sea_ice_ocean_fluxes.jl
@@ -1,5 +1,6 @@
 using Oceananigans.Operators: Δzᶜᶜᶜ
 using ClimaSeaIce.SeaIceThermodynamics: melting_temperature
+using ClimaSeaIce.SeaIceMomentumEquations: x_momentum_stress, y_momentum_stress
 
 function compute_sea_ice_ocean_fluxes!(coupled_model)
     ocean = coupled_model.ocean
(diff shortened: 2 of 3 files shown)


src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
```diff
@@ -121,20 +127,22 @@ end
...
+
+        τxao = ℑxᶠᵃᵃ(i, j, 1, grid, τᶜᶜᶜ, ρₒ⁻¹, ℵ, ρτxao)
+        τyao = ℑyᵃᶠᵃ(i, j, 1, grid, τᶜᶜᶜ, ρₒ⁻¹, ℵ, ρτyao)
+        τxio = ρτxio[i, j, 1] * ρₒ⁻¹ * ℑxᶠᵃᵃ(i, j, 1, grid, ℵ)
+        τyio = ρτyio[i, j, 1] * ρₒ⁻¹ * ℑyᵃᶠᵃ(i, j, 1, grid, ℵ)
 
+        τx[i, j, 1] = τxao + τxio
+        τy[i, j, 1] = τyao + τyio
         τx[i, j, 1] = τxao
         τy[i, j, 1] = τyao
```
navidcy, line 147: the last two lines overwrite the former two, right?

experiments/arctic_simulation.jl
```diff
@@ -134,19 +133,20 @@ using Statistics
...
+    vomax = maximum(abs, ocean.model.velocities.v)
 
     step_time = 1e-9 * (time_ns() - wall_time[])
 
     msg1 = @sprintf("time: %s, iteration: %d, Δt: %s, ", prettytime(sim), iteration(sim), prettytime(sim.Δt))
     msg2 = @sprintf("max(h): %.2e m, max(ℵ): %.2e ", hmax, ℵmax)
-    msg3 = @sprintf("mean(h): %.2e m, mean(ℵ): %.2e ", hmean, ℵmean)
-    msg4 = @sprintf("extrema(T): (%.2f, %.2f) ᵒC, ", Tmax, Tmin)
+    msg3 = @sprintf("max i-U: (%.2f, %.2f) ᵒC, ", uimax, vimax)
+    msg4 = @sprintf("max o-U: (%.2f, %.2f) ᵒC, ", uomax, vomax)
```
navidcy, line 149:
```suggestion
    msg3 = @sprintf("max uᵢ: (%.2f, %.2f) ᵒC, ", uimax, vimax)
    msg4 = @sprintf("max uₒ: (%.2f, %.2f) ᵒC, ", uomax, vomax)
```

src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
simone-silvestri: Oh yes! Good catch, this is a bug

experiments/arctic_simulation.jl
```diff
@@ -63,12 +63,11 @@ using ClimaSeaIce.SeaIceMomentumEquations
 using ClimaSeaIce.Rheologies
 
 # Remember to pass the SSS as a bottom bc to the sea ice!
-SSS = view(ocean.model.tracers.S, :, :, grid.Nz)
+SSS = view(ocean.model.tracers.S.data, :, :, grid.Nz)
```
navidcy, line 66: why you need `.data` here?

experiments/arctic_simulation.jl
```diff
@@ -77,14 +76,14 @@ dynamics = SeaIceMomentumEquation(grid;
                                   coriolis = ocean.model.coriolis,
                                   top_momentum_stress = (u=τua, v=τva),
                                   bottom_momentum_stress = τo,
-                                  ocean_velocities = (u=SSU, v=SSV),
+                                  ocean_velocities = (u=0.1*SSU, v=0.1*SSV),
```
navidcy, line 79: why 0.1 here? or it's just an example, doesn't matter really?
did you just reduce how much the sea ice feels the ocean?

src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
```diff
@@ -121,22 +127,25 @@ end
...
+
+        # Stresses
+        τx[i, j, 1] = τxao + τxio
+        τy[i, j, 1] = τyao + τyio
 
-        τx[i, j, 1] = τxao
-        τy[i, j, 1] = τyao
+        # Tracer fluxes
         Jᵀ[i, j, 1] = (1 - ℵᵢ) * Jᵀao + Jᵀio
         Jˢ[i, j, 1] = (1 - ℵᵢ) * Jˢao + Jˢio
```
navidcy, line 150: Just double checking that these two lines are correct.
Intuitively I'd expect them to be something like:

```julia
J[i, j, 1] = (1 - ℵᵢ) * Jao + ℵᵢ * Jio
```

so that when `ℵᵢ = 0` we have `Jao` and when `ℵᵢ = 1` we have `Jio`. But perhaps `ℵᵢ` is incorporated in the `Jio`?

src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
```diff
@@ -121,22 +127,25 @@ end
...
         Qio  = sea_ice_ocean_fluxes.interface_heat[i, j, 1]
-        Jˢio = sea_ice_ocean_fluxes.salt[i, j, 1]
+
+        Jᵀao = ΣQao  * ρₒ⁻¹ / cₒ
+        Jˢao = - Sₒ * ΣFao
         Jᵀio = Qio * ρₒ⁻¹ / cₒ
+        Jˢio = sea_ice_ocean_fluxes.salt[i, j, 1]
+
+        τxao = ℑxᶠᵃᵃ(i, j, 1, grid, τᶜᶜᶜ, ρₒ⁻¹, ℵ, ρτxao)
+        τyao = ℑyᵃᶠᵃ(i, j, 1, grid, τᶜᶜᶜ, ρₒ⁻¹, ℵ, ρτyao)
```
navidcy, line 140: Sometimes we use interpolation/derivative operators with `ᵃ` and some other times we ensure we prescribe e.g. `ℑxᶠᶜᶜ`. Is what happens here OK. Again just double checking only.

src/OceanSeaIceModels/InterfaceComputations/sea_ice_ocean_fluxes.jl
```diff
@@ -122,5 +146,9 @@ end
 
         # Update previous ice thickness
         h⁻[i, j, 1] = hᵢ[i, j, 1]
+
+        # Momentum stresses
+        τx[i, j, 1] = x_momentum_stress(i, j, Nz, grid, τₒᵢ, clock, sea_ice_fields)
+        τy[i, j, 1] = y_momentum_stress(i, j, Nz, grid, τₒᵢ, clock, sea_ice_fields)
     end
 end
```
navidcy, line 154:
```suggestion
end

```

navidcy (approved):
It'd be ideal if there was a test added? What do you think @simone-silvestri?

experiments/arctic_simulation.jl
simone-silvestri: because otherwise the field can be indexed only with `grid.Nz`, i.e. we get a windowedfield with indices `(:, :, grid.Nz)`. This is something we need to fix in `ClimaSeaIce` probably

experiments/arctic_simulation.jl
simone-silvestri: yep, this is what they use in SI3 apparently and I am not even sure why ahah. In practice, this is just an example, we should investigate these parameters

src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
simone-silvestri: `ℑxᶠᶜᶜ` does not exist yet, unfortunately. We should add it to Oceananigans and I did a couple fo attempts but left it hanging for a while

src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
simone-silvestri: I think you are right. `ℵᵢ` is incorporated in `Jio`, but it is probably better not to do it to make the code more clear. Let's change it in another PR.

test/test_surface_fluxes.jl
```diff
@@ -210,6 +211,50 @@ end
...
+        fill!(ocean.model.velocities.v, 0.2)
+        fill!(ocean.model.tracers.T,   -2.0)
+
+        # Test that we populate the sea-ice ocean stress
+        earth = OceanSeaIceModel(ocean, sea_ice; atmosphere, radiation=Radiation())
+
+        τx = earth.interfaces.sea_ice_ocean_interface.fluxes.x_momentum
+        τy = earth.interfaces.sea_ice_ocean_interface.fluxes.y_momentum
+
+        CUDA.@alloscalar begin
```
simone-silvestri, line 254:
```suggestion
        CUDA.@allowscalar begin
```

test/test_surface_fluxes.jl
```diff
@@ -210,6 +211,50 @@ end
             # Test that the temperature has snapped up to freezing
             @test minimum(ocean.model.tracers.T) == 0
         end
+
+        @info "Testing Surface Fluxes with sea ice..."
+
+        grid = ReactilinearGrid(arch;
```
simone-silvestri, line 217:
```suggestion
        grid = RectilinearGrid(arch;
```

test/test_surface_fluxes.jl
```diff
@@ -210,6 +211,50 @@ end
...
+                                size = (2, 2, 2), 
+                              extent = (1, 1, 1),
+                            topology = (Periodic, Periodic, Bounded))
```
simone-silvestri, line 220:
```suggestion
        grid = RectilinearGrid(arch;
                               size = (2, 2, 2), 
                             extent = (1, 1, 1),
                           topology = (Periodic, Periodic, Bounded))
```

PR #460, opened 2025-04-08
Clean up one degree simulation

glwagner:


navidcy:
merged to #461; let's continue there

PR #461, opened 2025-04-09
Test if examples run

simone-silvestri:
This is just a test to check that the examples run circumventing issue #399 (so we can check also the others examples)

cc @navidcy

### DIFF
--- a/examples/single_column_os_papa_simulation.jl
+++ b/examples/single_column_os_papa_simulation.jl
@@ -62,7 +62,7 @@ set!(ocean.model, T=ECCOMetadatum(:temperature),
 atmosphere = JRA55PrescribedAtmosphere(longitude = λ★,
                                        latitude = φ★,
                                        end_date = DateTime(1990, 1, 31), # Last day of the simulation
-                                       backend = JRA55NetCDFBackend(30))
+                                       backend  = InMemory())
 
 # This builds a representation of the atmosphere on the small grid
 


glwagner:
still getting this bizarre message

```
┌ Info: Skipping passes for interpolating bathymetry of size (21600, 10800, 1) 
│ to target grid of size (256, 128, 40). Interpolation passes may only 
│ be used to coarsen bathymetry and require that the bathymetry 
└ is finer than the target grid in both horizontal directions.
```

i thought we fixed that

glwagner:
we want to merge this, even if we have partial examples it is better than nothing

glwagner:
note its bad practice to change the download dir. It screws things up for users, because they will end up re-downloading lots of data when they run different cases in different locations. For example

image

glwagner:
so we should never manually set the download dir in an example

examples/single_column_os_papa_simulation.jl
```diff
@@ -62,7 +62,7 @@ set!(ocean.model, T=ECCOMetadatum(:temperature),
 atmosphere = JRA55PrescribedAtmosphere(longitude = λ★,
                                        latitude = φ★,
                                        end_date = DateTime(1990, 1, 31), # Last day of the simulation
-                                       backend = JRA55NetCDFBackend(30))
+                                       backend  = InMemory())
```
navidcy, line 65: @glwagner, I think the DO NOT MERGE label was for this change here. Because this is a hack for #399.
Otherwise I agree, better merge the examples that work and then fix more later.

navidcy:
@glwagner:

```Julia
julia> include("one_degree_simulation.jl")
ERROR: LoadError: UndefVarError: `passes` not defined
Stacktrace:
 [1] regrid_bathymetry(native_z::Field{…}, target_grid::OrthogonalSphericalShellGrid{…}, regridding_criteria::Tuple{…})
   @ ClimaOcean.Bathymetry ~/ClimaOcean.jl/src/Bathymetry.jl:247
 [2] regrid_bathymetry(target_grid::OrthogonalSphericalShellGrid{…}; height_above_water::Nothing, minimum_depth::Int64, dir::String, url::String, filename::String, regridding_criteria::Int64, major_basins::Int64)
   @ ClimaOcean.Bathymetry ~/ClimaOcean.jl/src/Bathymetry.jl:216
 [3] top-level scope
   @ ~/ClimaOcean.jl/examples/one_degree_simulation.jl:42
 [4] include(fname::String)
   @ Base.MainInclude ./client.jl:494
 [5] top-level scope
   @ REPL[3]:1
in expression starting at /home/navidcy/ClimaOcean.jl/examples/one_degree_simulation.jl:42
Some type information was truncated. Use `show(err)` to see complete types.
```

I don't know how to fix what happened here. I see you introduced `MustRefine()` and `InterpolationPasses()` types. But I'm not sure what's expected here. Bathymetry gets a `MustRefine` type but where are the passes? They are a field in one of the tuple's elements, right?

Would it be better if we ntroduced the Bathymetry module changes in a different PR? Otherwise it makes the docs hard to fix. :(

I'm tempted to enforce docs being built every time we merge something. Otherwise we end up with situations where docs are broken for months like we are now. But we can discuss this elsewhere.

navidcy:
> note its bad practice to change the download dir. It screws things up for users, because they will end up re-downloading lots of data when they run different cases in different locations. For example
> 
> image

True! I struggle with this myself. Is is the `dir = '.'` kwarg?

PR #462, opened 2025-04-09
Implement extension for SpeedyWeather

glwagner:
This PR adds an extension that will facilitate coupling with SpeedyWeather atmospheric models using OceanSeaIceModel (which may need to be renamed once the extension goes in...)


glwagner:
oy

PR #464, opened 2025-04-12
Change default gustiness to 1

glwagner:
The value 6.5 seems to cause a large overestimation of the momentum flux and in turn, the heat and vapor fluxes as well.

In a simple example which I might get to post later, changing this to 0 caused the momentum flux to divide by 3.

cc @simone-silvestri 

Note Fairall et al 1996 report values in the 1-1.2 range:

image

An alternative model for "low wind speed" is to simply cap the minimum velocity difference at some constant value like 0.5 m/s, eg

https://cosima.org.au/wp-content/uploads/2025/01/ACCESS-OM2-1-025-010deg-2025-01-24-7be8a1c.pdf#page=30.16

@navidcy and I propose to move this parameterization, and other "wind speed" parameterizations, so the `velocity_formulation` object eg in `WindVelocity` and `RelativeVelocity`:

https://github.com/CliMA/ClimaOcean.jl/blob/740c779435ee736aeabfdcfa303fb40ea3f7a93f/src/OceanSeaIceModels/InterfaceComputations/interface_states.jl#L106-L110

### DIFF
--- a/src/OceanSeaIceModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
+++ b/src/OceanSeaIceModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
@@ -58,10 +58,10 @@ end
 
 """
     SimilarityTheoryFluxes(FT::DataType = Float64;
-                           gravitational_acceleration = convert(FT, 9.81),
-                           von_karman_constant = convert(FT, 0.4),
-                           turbulent_prandtl_number = convert(FT, 1),
-                           gustiness_parameter = convert(FT, 6.5),
+                           gravitational_acceleration = 9.81,
+                           von_karman_constant = 0.4,
+                           turbulent_prandtl_number = 1,
+                           gustiness_parameter = 0,
                            stability_functions = default_stability_functions(FT),
                            roughness_lengths = default_roughness_lengths(FT),
                            similarity_form = LogarithmicSimilarityProfile(),
@@ -77,7 +77,7 @@ Keyword Arguments
 
 - `von_karman_constant`: The von Karman constant. Default: 0.4.
 - `turbulent_prandtl_number`: The turbulent Prandtl number. Default: 1.
-- `gustiness_parameter`: The gustiness parameter that accounts for low wind speed areas. Default: 6.5.
+- `gustiness_parameter`: Increases surface fluxes in low wind conditions. Default: 0.
 - `stability_functions`: The stability functions. Default: `default_stability_functions(FT)` that follow the
                          formulation of Edson et al. (2013).
 - `roughness_lengths`: The roughness lengths used to calculate the characteristic scales for momentum, temperature and
@@ -90,7 +90,7 @@ Keyword Arguments
 function SimilarityTheoryFluxes(FT::DataType = Oceananigans.defaults.FloatType;
                                 von_karman_constant = 0.4,
                                 turbulent_prandtl_number = 1,
-                                gustiness_parameter = 6.5,
+                                gustiness_parameter = 0,
                                 stability_functions = edson_stability_functions(FT),
                                 roughness_lengths = default_roughness_lengths(FT),
                                 similarity_form = LogarithmicSimilarityProfile(),
--- a/src/OceanSeaIceModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
+++ b/src/OceanSeaIceModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
@@ -90,7 +90,7 @@ Keyword Arguments
 function SimilarityTheoryFluxes(FT::DataType = Oceananigans.defaults.FloatType;
                                 von_karman_constant = 0.4,
                                 turbulent_prandtl_number = 1,
-                                gustiness_parameter = 0,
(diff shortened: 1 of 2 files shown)


src/OceanSeaIceModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -90,7 +90,7 @@ Keyword Arguments
 function SimilarityTheoryFluxes(FT::DataType = Oceananigans.defaults.FloatType;
                                 von_karman_constant = 0.4,
                                 turbulent_prandtl_number = 1,
-                                gustiness_parameter = 6.5,
+                                gustiness_parameter = 0,
```
glwagner, line 93:
```suggestion
                                gustiness_parameter = 1,
```

src/OceanSeaIceModels/PrescribedAtmospheres.jl
```diff
@@ -371,7 +371,7 @@ end
     PrescribedAtmosphere(grid, times;
                          clock = Clock{Float64}(time = 0),
                          surface_layer_height = 10, # meters
-                         boundary_layer_height = 600 # meters,
+                         boundary_layer_height = 0 # meters,
```
glwagner, line 374:
```suggestion
                         boundary_layer_height = 512 # meters,
```

src/OceanSeaIceModels/PrescribedAtmospheres.jl
```diff
@@ -385,8 +385,8 @@ state with data given at `times`.
 """
 function PrescribedAtmosphere(grid, times;
                               clock = Clock{Float64}(time = 0),
-                              surface_layer_height = convert(eltype(grid), 10),
-                              boundary_layer_height = convert(eltype(grid), 600),
+                              surface_layer_height = 10,
+                              boundary_layer_height = 0,
```
glwagner, line 389:
```suggestion
                              boundary_layer_height = 512,
```

glwagner:
It looks like we actually need gusintess to avoid NaNs in the surface flux tests.

We should look at the surface flux test to understand why.

For now, I am putting back default gustiness.

It's a thorny parameterization question about exactly what we should do at low wind speed.

My latest thought is that it might actually be best for gustiness to be provided by the JRA55 dataset / PrescribedAtmosphere. After all, the low winds are also coming from JRA55.

We can actually implement that kind of change if we want; eg add a field to the PrescribedAtmosphere corresponding to gustiness.

src/OceanSeaIceModels/InterfaceComputations/similarity_theory_turbulent_fluxes.jl
```diff
@@ -77,7 +77,7 @@ Keyword Arguments
 
 - `von_karman_constant`: The von Karman constant. Default: 0.4.
 - `turbulent_prandtl_number`: The turbulent Prandtl number. Default: 1.
-- `gustiness_parameter`: The gustiness parameter that accounts for low wind speed areas. Default: 6.5.
+- `gustiness_parameter`: Increases surface fluxes in low wind conditions. Default: 0.
```
navidcy, line 80:
```suggestion
- `gustiness_parameter`: Increases surface fluxes in low wind conditions. Default: 1.
```

glwagner:
merge?

navidcy:
I haven't tested it yet; shall I do that first?

glwagner:
might be nice to run 1 year of 1 degree and show the difference in surface temp?

simone-silvestri:
Looks good. This parameter looks to be a little all over the place. I think the issue with gustiness = 0 is that the scalar roughness lengths becomes zero making the bulk coefficients infinite. We should try to fix it, even if it seems like the case where ustar is zero should be already taken care of.

glwagner:
> Looks good. This parameter looks to be a little all over the place. I think the issue with gustiness = 0 is that the scalar roughness lengths becomes zero making the bulk coefficients infinite. We should try to fix it, even if it seems like the case where ustar is zero should be already taken care of.

can you link to the code where this happens?

simone-silvestri:
Yeah it shouldn't NaN because of this line in theory:
https://github.com/CliMA/ClimaOcean.jl/blob/0d6d34c84de05428b1703dfb4043cebc37fa121a/src/OceanSeaIceModels/InterfaceComputations/roughness_lengths.jl#L137

glwagner:
> Yeahm it shouldn't NaN because of these lines in theory:
> 
> https://github.com/CliMA/ClimaOcean.jl/blob/0d6d34c84de05428b1703dfb4043cebc37fa121a/src/OceanSeaIceModels/InterfaceComputations/roughness_lengths.jl#L137-L142

This is the momentum roughness length though, not the scalar roughness length?

I am skeptical about this roughness length formulation too. It doesn't make sense to me to define both a "smooth flow" limit, as well as gustiness.

glwagner:
I think the `β` above is different from the `β` that appears in the gustiness parameterization, something to be aware of.

glwagner:
Can we merge this and continue investigations in a PR or issue?

simone-silvestri:
sure let's do that

simone-silvestri (approved):

PR #467, opened 2025-04-14
Adding EN4 and World Ocean Atlas as initialisation datasets

taimoorsohail:
I am working on adding support for initialising and relaxing to WOA and EN4 datasets, to provide some flexibility in terms of the years and various configurations we can run on ClimaOcean. This PR covers such changes.


taimoorsohail:
Unfortunately, the inpainting doesn't seem to work with EN4, with large swathes having super high values. Below is the surface temperature at `t=0` for a tripolar, 1 degree grid. 

Screenshot 2025-04-14 at 9 14 45 pm

The above plot was made using this code: https://github.com/taimoorsohail/ocean-ensembles/blob/main/experiments/local_EN4_run.jl up to line 100

taimoorsohail:
I looked through `EN4.jl` and couldn't see the source of the error... any thoughts @glwagner or @simone-silvestri ?

taimoorsohail:
Temperature in EN4 is in Kelvin (i.e., between 269 and 310ish), could this be the issue?

simone-silvestri:
Nice job with this PR. I have a comment, I am noticing that many methods like `EN4_*` are basically copy and paste of the `ECCO_*` methods with very little change. 
Would it be maybe cleaner to generalize the methods like `ECCOFieldTimeSeries` and `ECCORestoring` to avoid code duplication?

for the masking issue, I don't see any issues at first glance. You can try plotting the mask to check if it corresponds to the masked region, and maybe add a debug statement in this function
https://github.com/CliMA/ClimaOcean.jl/blob/0d6d34c84de05428b1703dfb4043cebc37fa121a/src/DataWrangling/inpaint_mask.jl#L24-L40
to check where and why you get those high numbers and the inpainting does not propagate inside completely inside the bathymetry

glwagner:
We do want to solve the copy/paste issue because we will probably have a lot of different datasets, so avoiding code duplication will becoming important with 5-6 different datasets to work with (plus to make adding additional datasets easy).

But maybe @simone-silvestri we can work on this in another PR if this PR can get through?

I would suggest just doing one dataset per PR if possible?

taimoorsohail:
Thanks for the positive feedback! This PR is definitely not ready for merging. Persistent issues are:

1) Cannot create a coupled model as the initialised properties (T and S) are of order 1e10, so it stops the model from being created (due to inpainting bug; likely due to mismatch between model and simulation grid) --> can be moved to new PR
2) T and S are not properly initialised as one file contains both variables in EN4 compared with ECCO when two separate files exist for both T and S:
      -  `ocean.model.tracers.S == ocean.model.tracers.T` is currently `true` at `t=0`. 

I would like to at least resolve (2) before merging...

src/DataWrangling/DataWrangling.jl
```diff
@@ -127,8 +127,12 @@ include("metadata.jl")
 include("inpaint_mask.jl")
 include("JRA55/JRA55.jl")
 include("ECCO/ECCO.jl")
+include("WOA/WOA.jl")
```
glwagner, line 130:
```suggestion
```

maybe leftover

src/DataWrangling/EN4/EN4.jl
```diff
@@ -0,0 +1,257 @@
...
+    Nx, Ny, Nz, _ = size(metadata)
+    loc = location(metadata)
+    longitude = (0, 360)
+    latitude = (-90, 90)
+    TX, TY = (Periodic, Bounded)
+
+    if variable_is_three_dimensional(metadata)
+        TZ = Bounded
+        LZ = Center
+        z = EN4_z
```
glwagner, line 98: is this the line that needs to be different from `empty_ECCO_field`?

src/DataWrangling/EN4/EN4.jl
```diff
@@ -0,0 +1,257 @@
...
+                                 size = sz,
+                                 topology = (TX, TY, TZ))
+
+    return Field{loc...}(grid)
+end
+
+# Only temperature and salinity need a thorough inpainting because of stability,
+# other variables can do with only a couple of passes. Sea ice variables 
+# cannot be inpainted because zeros in the data are physical, not missing values.
+function default_inpainting(metadata::EN4Metadata)
```
glwagner, line 119: I think if this is the same for ECCO and EN4, then perhaps it can be the fallback method

glwagner:
@taimoorsohail I took a quick look at the PR files but I realized that it will be hard to identify exactly what can stay the same vs needs to change from ECCO stuff based on the code alone. Do you think that you can point out which lines are different between EN4 and ECCO? that will help us find the common abstractions that lead to zero (or minimal --- but in this case I do think zero is possible) code duplication

glwagner:
> Thanks for the positive feedback! This PR is definitely not ready for merging. Persistent issues are:
> 
> 1. Cannot create a coupled model as the initialised properties (T and S) are of order 1e10, so it stops the model from being created (due to inpainting bug; likely due to mismatch between model and simulation grid) --> can be moved to new PR
> 2. Functions need to be made generic to enable more datasets to be added --> should be moved to new PR
> 3. T and S are not properly initialised as one file contains both variables in EN4 compared with ECCO when two separate files exist for both T and S:
>    
>    * `ocean.model.tracers.S == ocean.model.tracers.T` is currently `true` at `t=0`.
> 
> I would like to resolve (3) before merging...

Are you setting the missing values in the EN4 data to NaN? This may be required depending on the dataset. Sometimes people use very large values (like 1e10) to represent NaN. So then you need to manually change these to NaN before trying to inpaint.

src/DataWrangling/EN4/EN4.jl
```diff
@@ -0,0 +1,257 @@
...
+        data = reverse(data, dims=3)
+    else
+        data = ds[shortname][:, :, 1]
+    end        
+
+    close(ds)
+    
+    # Convert data from Union(FT, missing} to FT
+    FT = eltype(field)
+    data[ismissing.(data)] .= 1e10 # Artificially large number!
```
glwagner, line 184:
```suggestion
    data[ismissing.(data)] .= NaN # Artificially large number!
```

glwagner: is this the issue?

glwagner:
Here's a link to the EN4 data: https://www.metoffice.gov.uk/hadobs/en4/

I'm curious if the objectively analyzed data will work as well as the dynamically consistent ECCO data. But it will be good to test to see whether this is indeed a better choice for initializing years outside the ECCO dataset, or not.

src/DataWrangling/EN4/EN4.jl
taimoorsohail: Yes these are the same.

src/DataWrangling/EN4/EN4.jl
taimoorsohail: I think this is part of the issue; it may be that the Boolean mask for EN4 isn't being created properly, such that it is `false` everywhere - in this line: 

This would lead to the final field having a bunch of `missing` values, which are then set to `1e10` in the above.

src/DataWrangling/EN4/EN4.jl
navidcy: Yes, no need for a new method then.
I can help smooth things out.

You understand what fall back is?

@glwagner suggests something:

```Julia
function default_inpainting(metadata)
    do_things
    return some_things
end
```

and if particular datasets want something different they can implement


```julia
function default_inpainting(metadata::EN4Metadata)
    do_special_EN4_things
    return some_things
end
```

taimoorsohail:
EN4 is a good alternative to ECCO. Here are some pros and cons:

**PROS**
1) EN4 is an objectively analysed gridded observational product, so it isn't impacted by model drift or biases.
2) It has monthly coverage from 1900 (dubious reliability before 1970) to 2025, so it gives us a lot of freedom to explore other time ranges, especially now that JRA55-do can be run over its entire 1958-present time range.
3) It is the closest thing we have to pure obs, though we have to be wary of mapping and sampling bias errors (particularly for salinity during the Argo spin-up period).
4) It is full depth; whereas more complex gridded products cut off at 2000m 

**CONS**
1) It is impacted by errors in the mapping method - specifically, for EN4, it reverts to climatology where no observations are available, which can lead to patchiness at times.
2) EN4 is not necessarily physically consistent. Without an underlying dynamical model, the budgets don't close, and the water masses are not necessarily physically consistent.
3) Its mapping method is not particularly sophisticated, newer machine learning or CMIP-based mapping methods reduce errors, though these products are often not full depth.

src/DataWrangling/EN4/EN4.jl
glwagner: I think the inpainting will only replace NaN. If there are 1e10 in the file, those values will not be touched by inpainting. That can in turn lead to 1e10 values in the final field

src/DataWrangling/EN4/EN4.jl
glwagner: could we define a function called `z_faces(metadata)`, and then change the name of the parent method to `empty_field`? Then we only need one method `empty_field`, which finds the z faces for that metadata

src/DataWrangling/EN4/EN4_metadata.jl
```diff
@@ -0,0 +1,176 @@
...
+
+datestr(md::EN4Metadata) = string(first(md.dates), "--", last(md.dates))
+datestr(md::EN4Metadatum) = string(md.dates)
+
+Base.summary(md::EN4Metadata) = string("EN4Metadata{", datasetstr(md), "} of ",
+                                        md.name, " for ", datestr(md))
+
+Base.size(data::Metadata{<:EN4Monthly}) = (360, 173, 42, length(data.dates))
+
+Base.size(::Metadatum{<:EN4Monthly}) = (360, 173, 42, 1)
```
glwagner, line 49: do these two methods overwrite one another?

src/DataWrangling/EN4/EN4.jl
```diff
@@ -0,0 +1,257 @@
...
+    latitude = (-90, 90)
+    TX, TY = (Periodic, Bounded)
+
+    if variable_is_three_dimensional(metadata)
+        TZ = Bounded
+        LZ = Center
+        z = EN4_z
+        halo = (horizontal_halo..., 3)
+        sz = (Nx, Ny, Nz)
+    else # the variable is two-dimensional
```
glwagner, line 101: what are the two-dimensional fields available with EN4?

src/DataWrangling/EN4/README.md
glwagner, line 1: This file can likely be deleted (or adapted?)

glwagner:
After looking through this PR I believe that the definition of `EN4Metadata` and the vertical grid is perhaps the primary additional things we will need to get `EN4` working.

I also think that we need to generalize a few methods. @taimoorsohail let me know if you think that I have these right, here are some methods we need:

* `Field`, to replace `ECCO_field` and `EN4_field`
* `vertical_interfaces` , to get the vertical grid (assuming that the horizontal grid for both covers the sphere on a lat-lon grid, we can generalize this later if we have other data)
* `metadata_mask` to replace `ECCO_mask` and `EN4_mask`
* I think also a definition of missing values

the last two will be used in the `Field` constructor.

I may have missed something so please let me know if so @taimoorsohail !

Project.toml
```diff
@@ -13,6 +13,7 @@ CubicSplines = "9c784101-8907-5a6d-9be6-98f00873c89b"
 DataDeps = "124859b0-ceae-595e-8997-d05f6a7a8dfe"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
+Glob = "c27321d9-0574-5035-807b-f59d2c89b15c"
```
navidcy, line 16: do we need Glob?

src/DataWrangling/EN4/EN4.jl
taimoorsohail: EN4 marks land as `missing`, not a junk value. The `1e10` is introduced somewhere in this code, I am trying to pin down where.

Project.toml
taimoorsohail: Nope, sorry, that was me playing around and it stayed in there

src/DataWrangling/EN4/EN4.jl
```diff
@@ -0,0 +1,257 @@
...
+
+    return field
+end
+
+# Fallback
+EN4_field(var_name::Symbol; kw...) = EN4_field(EN4Metadata(var_name); kw...)
+
+function inpainted_metadata_filename(metadata::EN4Metadata, var::AbstractString)
+    original_filename = metadata_filename(metadata)
+    without_extension = original_filename[1:end-3]
```
taimoorsohail, line 232: Why is this selection `[1:end-3]` taken from the vector of strings in `original_filename`?  I'm not sure why `without_extension` isn't `original_filename[1]`?

src/DataWrangling/EN4/EN4.jl
```diff
@@ -0,0 +1,257 @@
...
+    return without_extension * "_" * var *"_inpainted.jld2"
```
taimoorsohail, line 233: FYI - I added `var` as an addition to the unpainted file names, ensuring that if one file contains multiple variables, each unpainted variable is output as a separate file

src/DataWrangling/EN4/EN4.jl
navidcy: I can do that. Shall I @taimoorsohail?

taimoorsohail: Yes, ideally for EN4 we would want something like:

```Julia
function z_faces(m::EN4Metadata)
netcdf_filename = metadata_filename(metadata)[1] # index 1 because it will be a string of filenames for all dates
data = Dataset(netcdf_filename) 
z_faces = sort(-1.*Array(data["depth_bnds"])[1,:]) # index 1 because this is the lower bound
push!(z_faces, -1.*Array(data["depth_bnds"])[2,end]) #append the last term of index 2, should be -5500 m in EN4
return z_faces
```

taimoorsohail:
I have a hunch that there may be a bug in inpainting for a tripolar grid... It appears that inpainting occurs by `i,j` index, instead of by `lat/lon`. In a `LatLonGrid` these are the same, but in a tripolar grid, they are not, so you get distortion/warping of the land masses. So, the mask for land is distorted when inpainting with both ECCO and EN4. You can see this from the figure above, where the yellow represents landmasses in EN4, and appears to stretch as you get further south.

taimoorsohail:
@navidcy 

The below code should output a figure with proper inpainting (though the distortion of land masses remains in `LatLonGrid` so maybe I was wrong about the bug in land distortion). Remember to delete the inpainted netCDF files as the code is directly pulling from those, and will need to make new ones to reflect changes to the code. 

```Julia
using ClimaOcean
using Oceananigans
using Oceananigans.Units
using CFTime
using Dates
using Printf
using OceanEnsembles
using Oceananigans.Operators: Ax, Ay, Az, Δz
using Oceananigans.Fields: ReducedField
using ClimaOcean.EN4
using ClimaOcean.EN4: download_dataset

# ### EN4 files
@info "Downloading/checking EN4 data"
## We download Gouretski and Reseghetti (2010) XBT corrections and Gouretski and Cheng (2020) MBT corrections

dates = collect(DateTime(2022, 1, 1): Month(1): DateTime(2023, 12, 1))

data_path = expanduser("/Users/tsohail/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/uom/ocean-ensembles/data/")

temperature = Metadata(:temperature; dates, dataset=EN4Monthly(), dir=data_path)
salinity    = Metadata(:salinity;    dates, dataset=EN4Monthly(), dir=data_path)

download_dataset(temperature)

Nx = Integer(360)
Ny = Integer(180)
Nz = Integer(100/2)

arch = CPU()

z_faces = (-4000, 0)

## In case you want to test with a Tripolar Grid
# underlying_grid = TripolarGrid(arch;
#                                size = (Nx, Ny, Nz),
#                                z = z_faces,
#                                halo = (5, 5, 4),
#                                first_pole_longitude = 70,
#                                north_poles_latitude = 55)

underlying_grid = LatitudeLongitudeGrid(arch;
                                        size = (Nx, Ny, Nz),
                                        z = z_faces,
                                        halo = (5, 5, 4),
                                        longitude = (0, 350),
                                        latitude = (-75, 75))

@info "Defining bottom bathymetry"

@time bottom_height = regrid_bathymetry(underlying_grid;
                                  minimum_depth = 10,
                                  interpolation_passes = 0, # 75 interpolation passes smooth the bathymetry near Florida so that the Gulf Stream is able to flow
				  major_basins = 2)

# For this bathymetry at this horizontal resolution we need to manually open the Gibraltar strait.
# view(bottom_height, 102:103, 124, 1) .= -400

@info "Defining grid"

@time grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom_height); active_cells_map=true)

@info "Defining restoring rate"

restoring_rate  = 2 / 365.25days
z_below_surface = z_faces[end-1]

mask = LinearlyTaperedPolarMask(southern=(-80, -70), northern=(70, 90), z=(z_below_surface, 0))

FT = EN4Restoring(temperature, grid; mask, rate=restoring_rate)
FS = EN4Restoring(salinity,    grid; mask, rate=restoring_rate)
forcing = (T=FT, S=FS)

@info "Defining ocean simulation"

@time ocean = ocean_simulation(grid)

@info "Initialising with EN4"

set!(ocean.model, T=Metadata(:temperature; dates=first(dates), dataset=EN4Monthly(), dir=data_path),
                    S=Metadata(:salinity;    dates=first(dates), dataset=EN4Monthly(), dir=data_path))

## Plot the intitalised SST and SSS
using GLMakie
fig = Figure(size = (1500,2000)) # create a new figure
ax1 = Axis(fig[1, 1])            # add an axis to the figure
ax2 = Axis(fig[2, 1])        # add an axis to the figure
axc1 = (fig[1, 2])            # add an axis to the figure
axc2 = (fig[2, 2])        # add an axis to the figure

Tslice = dropdims(interior(view(ocean.model.tracers.T, :, :, Nz)), dims=3)
Sslice = dropdims(interior(view(ocean.model.tracers.S, :, :, Nz)), dims=3)
hm1 = heatmap!(ax1, Tslice.-273.15; colorrange = (-3, 30), colormap = Reverse(:deep))
hm2 = heatmap!(ax2, Sslice; colorrange = (34,38), colormap = :bwr)
Colorbar(axc1, hm1, label = "°C")
Colorbar(axc2, hm2, label = "g/kg")
fig
```

Screenshot 2025-04-15 at 4 23 11 pm

src/DataWrangling/EN4/EN4.jl
simone-silvestri: I think because we want to remove the `.nc` extension and add the `_inpainted.jld2` extension

src/DataWrangling/EN4/EN4.jl
simone-silvestri: I think the `var` is included in the `metadata_filename`. Is this not the case for `EN4`?

src/DataWrangling/EN4/EN4_restoring.jl
```diff
@@ -0,0 +1,337 @@
+using Oceananigans: location
```
simone-silvestri, line 1: I think also this file is basically general and can be reused.

src/DataWrangling/EN4/EN4.jl
taimoorsohail: Ah, OK, that makes sense, thanks!

src/DataWrangling/EN4/EN4.jl
taimoorsohail: Yes, it is, I updated this function again so it doesn't require the extra `var` argument and can pull it directly from `metadata.name`

src/DataWrangling/EN4/EN4_restoring.jl
taimoorsohail: That's right, this is basically a complete copy/paste. The only thing I have done is removed the extra variables that don't exist in EN4, like velocities and sea ice thickness. 

I should note I haven't fully tested if the restoring is working so it may be that we need to come back and pull some functions out to work differently for EN4, but as it is this is basically identical.

src/DataWrangling/EN4/EN4.jl
taimoorsohail: I figured this out - the inpainting is working now; though I suspect there might be some land mask distortion...

taimoorsohail:
> I have a hunch that there may be a bug in inpainting for a tripolar grid... It appears that inpainting occurs by `i,j` index, instead of by `lat/lon`. In a `LatLonGrid` these are the same, but in a tripolar grid, they are not, so you get distortion/warping of the land masses. So, the mask for land is distorted when inpainting with both ECCO and EN4. You can see this from the figure above, where the yellow represents landmasses in EN4, and appears to stretch as you get further south.

@simone-silvestri let me know if you have thoughts on this. I might be wrong about how the land masking and inpainting happens...

src/DataWrangling/EN4/EN4_metadata.jl
```diff
@@ -0,0 +1,197 @@
+using CFTime
```
taimoorsohail, line 1: This file is basically all new, so there is no overlap (as far as I can tell) with the ECCO_metadata file

src/DataWrangling/EN4/EN4_mask.jl
```diff
@@ -0,0 +1,52 @@
...
+
+    return mask
+end
+
+# Default
+EN4_mask(arch::AbstractArchitecture=CPU()) = EN4_mask(Metadata(:temperature, dataset=EN4Monthly()), arch)
+
+@kernel function _set_EN4_mask!(mask, Tᵢ, args...)
+    i, j, k = @index(Global, NTuple)
+    @inbounds mask[i, j, k] = (Tᵢ[i, j, k] == 1e10)
```
taimoorsohail, line 30: This is the only line that is different from ECCO in this file - to do with the fact that missing data in EN4 is replaced with `1e10`, so we use that as the marker of land in EN4.

simone-silvestri:
> > I have a hunch that there may be a bug in inpainting for a tripolar grid... It appears that inpainting occurs by `i,j` index, instead of by `lat/lon`. In a `LatLonGrid` these are the same, but in a tripolar grid, they are not, so you get distortion/warping of the land masses. So, the mask for land is distorted when inpainting with both ECCO and EN4. You can see this from the figure above, where the yellow represents landmasses in EN4, and appears to stretch as you get further south.
> 
> @simone-silvestri let me know if you have thoughts on this. I might be wrong about how the land masking and inpainting happens...

It might be that the inpainting is not very precise, it is a quite primitive algorithm to fill in empty values (just propagates horizontally based on the neighboring values). I am not sure we need much more though, the inpainting should be a way to ensure the simulation does not crash upon initialization because of pockets of zero salinity (or missing values or NaNs), but the initial solution will still undergo initialization shocks, and eventually rebalance to equilibrium.

src/DataWrangling/EN4/EN4_mask.jl
simone-silvestri: looks good!

src/DataWrangling/EN4/EN4_metadata.jl
simone-silvestri: yeah, defining new methods for a particular dataset metadata ideally should be the only addition we need.

src/DataWrangling/EN4/EN4_metadata.jl
```diff
@@ -0,0 +1,197 @@
...
+        for files in unique(missingfiles)
+            @info files
+            ds = Dataset(files, "a")  # open in append mode
+            temp = ds["temperature"]
+            data = temp[:, :, :, 1]
+            data .= ifelse.(ismissing.(data), data, data .- 273.15)
+            temp[:, :, :, 1] .= data
+            temp.attrib["units"] = "degree_Celsius"
+            close(ds)
+        end
```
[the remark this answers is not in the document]
simone-silvestri: I probably would not change the data at source, but when you build a `Field`, a `FieldTimeSeries` or a `Restoring` object. In that case you are loading the data, so we can maybe add a `data = convert_units(data, metadata)` method, where `metadata` should contain all the information you need to convert the units.
Here, for example
https://github.com/CliMA/ClimaOcean.jl/blob/77dbfa0ee5509e4aae40326153e0c916da10c36c/src/DataWrangling/EN4/EN4.jl#L155

taimoorsohail:
> > > I have a hunch that there may be a bug in inpainting for a tripolar grid... It appears that inpainting occurs by `i,j` index, instead of by `lat/lon`. In a `LatLonGrid` these are the same, but in a tripolar grid, they are not, so you get distortion/warping of the land masses. So, the mask for land is distorted when inpainting with both ECCO and EN4. You can see this from the figure above, where the yellow represents landmasses in EN4, and appears to stretch as you get further south.
> > 
> > 
> > @simone-silvestri let me know if you have thoughts on this. I might be wrong about how the land masking and inpainting happens...
> 
> It might be that the inpainting is not very precise, it is a quite primitive algorithm to fill in empty values (just propagates horizontally based on the neighboring values). I am not sure we need much more though, the inpainting should be a way to ensure the simulation does not crash upon initialization because of pockets of zero salinity (or missing values or NaNs), but the initial solution will still undergo initialization shocks, and eventually rebalance to equilibrium.

If we are restoring SST and/or SSS in the simulation (as we are in some of the examples), then the presence of a land mask that doesn't align with our simulation makes a difference. The southern part of Africa, South America and Australia will all have large swathes that are being restored to an inpainted property, rather than the `real` T or S.

src/DataWrangling/EN4/EN4_metadata.jl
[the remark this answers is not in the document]
taimoorsohail: So could we do something like:

```Julia
if shortname == "temperature"    
if variable_is_three_dimensional(metadata)
        data = ds[shortname][:, :, :, 1] .- 273.15
        data = reverse(data, dims=3)
    else
        data = ds[shortname][:, :, 1] .-273.15
    end
    else
    if variable_is_three_dimensional(metadata)
        data = ds[shortname][:, :, :, 1]
        data = reverse(data, dims=3)
    else
        data = ds[shortname][:, :, 1]
    end
end
```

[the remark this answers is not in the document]
navidcy: let me think about this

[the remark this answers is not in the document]
simone-silvestri: I would actually create a method that takes in the data and the metadata and converts to the proper units. 
Like this (you can probably change the name of the function):
```julia
data = convert_units(data, metadata)
```
then
```julia
convert_units(data, metadata) = data
convert_units(data, metadata::EN4Metadata) = metadata.name == :temperature ? data .- 273.15 : data
```

src/DataWrangling/mask.jl
glwagner, line 1: I thnk "masking.jl" might be a better name for this file

src/DataWrangling/metadata.jl
```diff
@@ -178,3 +178,59 @@ function compute_native_date_range(native_dates, start_date, end_date)
 
     return native_dates[start_idx:end_idx]
 end
+
+"""
+    z_faces(metadata)
+
+Return an array with the ``z``-faces of the dataset that `metadata` corresponds to.
+"""
+z_faces(metadata::Metadata{V}) where V =
```
glwagner, line 187:
```suggestion
vertical_interfaces(metadata::Metadata{V}) where V =
```

src/DataWrangling/metadata.jl
```diff
@@ -178,3 +178,59 @@ function compute_native_date_range(native_dates, start_date, end_date)
...
+
+    return Field{loc...}(grid)
+end
+
+shift_longitude_to_0_360(data, metadata) = data
+
+struct Celsius end
+struct Kelvin end
+
+dataset_temperature_units(metadata) = nothing
```
glwagner, line 236: are these valid units? I thikn if there is not a good fallback then it is better not to define it. Once can simply create a function with no methods by writing

```julia
function dataset_temperature_units end
```

src/DataWrangling/inpaint_mask.jl
glwagner, line 1: what's the difference between `inpaint_mask.jl` and `mask.jl`? I might get confused between what is contained within these two --- perhaps we can combine the files, or find more specific names?

src/DataWrangling/EN4/EN4_restoring.jl
glwagner: Should we try to get field working first in this PR, and then tackle restoring in a future PR?

src/DataWrangling/EN4/EN4_mask.jl
glwagner: Can we generalize this concept and define `mask_value(metadata)`, so that we only need to have one mask function?

src/DataWrangling/EN4/EN4_metadata.jl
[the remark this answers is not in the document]
glwagner: I think it will lead to better code to embed the variable names as types or otherwise figure out how to use dispatch here, eg

```julia
convert_metadata_value(value, metadata, ::Temperature) = value - 273.15
```

glwagner:
My main suggestion is to get a simulation working that is initialized from EN4. Then in a future PR, we can work on getting restoring to work. This incremental approach will help us improve the quality of the code I think.

src/DataWrangling/inpaint_mask.jl
navidcy: good point

navidcy:
I'm in the process of cleaning it up a bit so don't bother looking until I ping you.

At the moment this:

```Julia
using ClimaOcean
using CFTime
using GLMakie
using Oceananigans
using Oceananigans.Units
using Dates
using Printf

using ClimaOcean.ECCO
using ClimaOcean.EN4
using ClimaOcean.ECCO: download_dataset

dates = DateTime(1993, 1, 1): Month(1): DateTime(1993, 1, 31)
data_path = "."

for dataset in (ECCO4Monthly(), EN4Monthly())
    temperature = Metadata(:temperature; dates, dataset, dir=data_path)
    salinity    = Metadata(:salinity;    dates, dataset, dir=data_path)

    download_dataset(temperature)
    download_dataset(salinity)
end

Nx, Ny, Nz = 360, 180, 50

underlying_grid = LatitudeLongitudeGrid(size = (Nx, Ny, Nz),
                                        z = (-4000, 0),
                                        halo = (7, 7, 3),
                                        longitude = (0, 360),
                                        latitude = (-90,90))

bottom_height = regrid_bathymetry(underlying_grid;
                                  minimum_depth = 10,
                                  interpolation_passes = 0,
                                  major_basins = 2)

grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom_height); active_cells_map=true)

ocean = ocean_simulation(grid)

# to ensure you are not using an old, wrong inpainted file
rm(joinpath(data_path, "THETA_1993_01_inpainted.jld2"), force=true)
rm(joinpath(data_path, "SALT_1993_01_inpainted.jld2"), force=true)
rm(joinpath(data_path, "EN.4.2.2.f.analysis.g10.199301_temperature_inpainted.jld2"), force=true)
rm(joinpath(data_path, "EN.4.2.2.f.analysis.g10.199301_salinity_inpainted.jld2"), force=true)


for dataset in (ECCO4Monthly(), EN4Monthly())
    temperature = Metadata(:temperature; dates, dataset, dir=data_path)
    salinity    = Metadata(:salinity;    dates, dataset, dir=data_path)

    set!(ocean.model, T=first(temperature),
                    S=first(salinity))

    T = ocean.model.tracers.T
    S = ocean.model.tracers.S

    land = interior(grid.immersed_boundary.bottom_height, :, :, 1) .≥ 0

    Tn = interior(T, :, :, Nz)
    Sn = interior(S, :, :, Nz)

    Tn[land] .= NaN
    Sn[land] .= NaN

    fig = Figure()
    ax1 = Axis(fig[1, 1])
    ax2 = Axis(fig[2, 1])

    hm1 = heatmap!(ax1, Tn; colorrange = (-3, 30), colormap = Reverse(:deep), nan_color=:silver)
    hm2 = heatmap!(ax2, Sn; colorrange = (34, 38), colormap = :bwr, nan_color=:silver)
    Colorbar(fig[1, 2], hm1, label = "°C")
    Colorbar(fig[2, 2], hm2, label = "g/kg")

    Label(fig[0, :], string(typeof(dataset)), tellwidth=false)

    save("initial_T_S_" * string(typeof(dataset)) * ".png", fig)
end
```

gives

initial_T_S_ECCO4Monthly

initial_T_S_EN4Monthly

src/DataWrangling/metadata.jl
navidcy: good point!

src/DataWrangling/mask.jl
navidcy: agree + combine with `inpaint_mask.jl`

glwagner:
nice, that's great! how different are the two?

I think ultimately each dataset (ECCO or EN4) might be expressable in a single file, and everything else is a common utility in DataWrangling

navidcy:
> After looking through this PR I believe that the definition of `EN4Metadata` and the vertical grid is perhaps the primary additional things we will need to get `EN4` working.
> 
> I also think that we need to generalize a few methods. @taimoorsohail let me know if you think that I have these right, here are some methods we need:
> 
> * `Field`, to replace `ECCO_field` and `EN4_field`
> * `vertical_interfaces` , to get the vertical grid (assuming that the horizontal grid for both covers the sphere on a lat-lon grid, we can generalize this later if we have other data)
> * `metadata_mask` to replace `ECCO_mask` and `EN4_mask`
> * I think also a definition of missing values
> 
> the last two will be used in the `Field` constructor.
> 
> I may have missed something so please let me know if so @taimoorsohail !

OK, I had missed this post but I mostly did this. I wrote:
- `dataset_field()` to replace `ECCO_field` and `EN4_field
- `dataset_mask()` to replace `ECCO_mask` and `EN4_mask`
- `vertical_interfaces`
- put more common methods up one level in DataWrangling

navidcy:
The difference of the two datasets in January:

difference_T_S

navidcy:
I was a bit hesitant to add a method for `Field`. Would it be better to extend `Field` instead of the `dataset_field`?

glwagner:
> The difference of the two datasets in January:
> 
> difference_T_S

This is interesting! Temperature is substantially different, but salinity is not, it seems.

navidcy:
Changed `dataset_field` to `Field`. I think it's clearer. I agree it doesn't seem ambiguous.

I'm gonna stop for today. Have a look if you like.

navidcy:
I'd like to generalize test_ecco to test both datasets. Or add a `test_en4.jl`..

src/DataWrangling/EN4/EN4.jl
navidcy: I tried using the above to replace 

https://github.com/CliMA/ClimaOcean.jl/blob/b0c141c9eddeb38bc4163112675cee2251209c6f/src/DataWrangling/EN4/EN4.jl#L39

but it wasn't working. I think it must be something trivial, but I'll stop for today.

src/DataWrangling/EN4/EN4_metadata.jl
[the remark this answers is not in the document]
navidcy: @glwagner's suggestion is ideal. I actually only now notice both @simone-silvestri and @glwagner's replies. At the moment I did:

https://github.com/CliMA/ClimaOcean.jl/blob/b0c141c9eddeb38bc4163112675cee2251209c6f/src/DataWrangling/dataset.jl#L86-L88

src/DataWrangling/EN4/EN4_mask.jl
navidcy: I've done

https://github.com/CliMA/ClimaOcean.jl/blob/b0c141c9eddeb38bc4163112675cee2251209c6f/src/DataWrangling/masking.jl#L241-L254

and defined `default_set_dataset_mask(metadata)`. But your suggestion is even better! Perhaps tomorrow I can implement it.

taimoorsohail:
Thanks @navidcy for the huge effort in modifying this! 

I'll have a closer look tomorrow, but I just wanted to flag that I think there is still some stretching of the land going on prior to inpainting in EN4. My suspicion is that it's because EN4 has length 173 in the latitude dimension (from 89N to 83S) and the interpolation is somehow stretching the landmasses to fit over length 180 in the model grid. This would explain why the land only seems to stretch in the y dimension, and it stretches more in the Southern Hemisphere than the Northern Hemisphere. Anyway, will dig into this a bit more tomorrow!

glwagner:
> Thanks @navidcy for the huge effort in modifying this!
> 
> I'll have a closer look tomorrow, but I just wanted to flag that I think there is still some stretching of the land going on prior to inpainting in EN4. My suspicion is that it's because EN4 has length 173 in the latitude dimension (from 89N to 83S) and the interpolation is somehow stretching the landmasses to fit over length 180 in the model grid. This would explain why the land only seems to stretch in the y dimension, and it stretches more in the Southern Hemisphere than the Northern Hemisphere. Anyway, will dig into this a bit more tomorrow!

Do we have information about the horizontal grid for EN4?

navidcy:
> I'll have a closer look tomorrow, but I just wanted to flag that I think there is still some stretching of the land going on prior to inpainting in EN4. My suspicion is that it's because EN4 has length 173 in the latitude dimension (from 89N to 83S) and the interpolation is somehow stretching the landmasses to fit over length 180 in the model grid. This would explain why the land only seems to stretch in the y dimension, and it stretches more in the Southern Hemisphere than the Northern Hemisphere. Anyway, will dig into this a bit more tomorrow!

Idea: If we have a lat-lon grid from 83S to 83N, would the stretching go away? That might give a hint?

src/DataWrangling/EN4/EN4_restoring.jl
navidcy: I agree with this suggestion.

navidcy:
OK, I also generalised the tests to cover cases with EN4 fields for `set!` and time stepping. Nothing that touches the restoring.

Based on the title of the PR this is ready (except the land-stretching issue that @taimoorsohail mentions).

I'm suggesting that if we drop the EN4_restoring.jl file; merge this PR, and open a new one in which we generalise the dataset restoring functionality and add the restoring + tests for restoring.

src/DataWrangling/EN4/EN4.jl
```diff
@@ -0,0 +1,95 @@
...
+    -20.1158,
+    -10.0475,
+      0.0,
+    ]
+
+function inpainted_metadata_filename(metadata::EN4Metadata)
+    original_filename = metadata_filename(metadata)
+    without_extension = original_filename[1:end-3]
+    var = string(metadata.name)
+    return without_extension * "_" * var *"_inpainted.jld2"
```
taimoorsohail, line 90: I think this function is general and should be moved up one level -- the only difference is that for EN4: we `return without_extension * "_" * var *"_inpainted.jld2"` and for ECCO: we `return without_extension * "_inpainted.jld2"`.

The inclusion of `var` is more general as it works for both cases where variables are in separate files (as in ECCO) and where variables are in the one file (as in EN4).

src/DataWrangling/EN4/EN4_metadata.jl
```diff
@@ -0,0 +1,186 @@
...
+
+or by invoking
+
+```julia
+julia> ENV["EN4_USERNAME"] = "myusername"
+
+julia> ENV["EN4_PASSWORD"] = "mypassword"
+```
+
+within julia.
```
navidcy, line 149: @taimoorsohail do we need this? I haven't set any EN4_USERNAME/EN4_PASSWORD and was able to download things.

navidcy, line 149: @taimoorsohail do we need these instructions? I haven't set any EN4_USERNAME/EN4_PASSWORD and was able to download things.

src/DataWrangling/EN4/EN4.jl
navidcy: I'll try it and see what happens with eg ECCO. Would we have filenames like `THETA_xyz_temperature_inpainted.jld2`?

navidcy: (shouldn't matter even if this is the case because in an ideal world all of these files are somewhere in a scratch directory and not visible to the user?)

src/DataWrangling/EN4/EN4_restoring.jl
```diff
@@ -36,17 +36,17 @@ Adapt.adapt_structure(to, b::EN4NetCDFBackend{N, C}) where {N, C} = EN4NetCDFBac
...
 """
 function EN4NetCDFBackend(length, metadata;
-                           on_native_grid = false,
-                           cache_inpainted_data = false,
-                           inpainting = NearestNeighborInpainting(Inf))
+                          on_native_grid = false,
+                          cache_inpainted_data = false,
+                          inpainting = NearestNeighborInpainting(Inf))
 
     return EN4NetCDFBackend{on_native_grid, cache_inpainted_data}(1, length, inpainting, metadata)
```
glwagner: if the backend stores the metadata, is it possible to have one backend for both ECCO and EN4? This is a minor savings in code so not crucial --- just curious

src/DataWrangling/EN4/EN4_metadata.jl
taimoorsohail: Nope - EN4 is publicly available without any API or PW/username. I think this is a result of me find/replacing ECCO --> EN4. It can be removed!

src/DataWrangling/EN4/EN4.jl
taimoorsohail: Yes that is what the ECCO filename would look like I think!

taimoorsohail:
Resolves #468

.buildkite/pipeline.yml
```diff
@@ -91,10 +91,10 @@ steps:
       slurm_ntasks: 1
       slurm_gpus_per_task: 1
 
-  - label: "GPU ECCO tests"
-    key: "gpu_ecco_tests"
+  - label: "GPU Datasets tests"
+    key: "gpu_datasets_tests"
     env:
-      TEST_GROUP: "ecco"
+      TEST_GROUP: "datasets"
```
glwagner, line 97: will this include JRA55?

src/DataWrangling/EN4/EN4_restoring.jl
navidcy: what backend you referring to? this is part of a file I removed

.buildkite/pipeline.yml
navidcy: hm... good point
in an ideal world yes
but perhaps until then should we call this group `ecco_en4`?

navidcy:
> Resolves #468

two points: 1/ #469 resolved #468, right? also 2/ this comments "resolves #XYZ" must be made on the first comment of the PR for GitHub to automatically link the PR with the issue

src/DataWrangling/EN4/EN4_metadata.jl
navidcy: Removed it ;)
Just wanted to double check!

src/DataWrangling/EN4/EN4_restoring.jl
navidcy: You are referring to `EN4NetCDFBackend`, right? This is not part of this PR so the comment is relevant of the Restoring refactor, right?

.buildkite/pipeline.yml
navidcy: I renamed it for now.

src/DataWrangling/EN4/EN4_restoring.jl
glwagner: oh sorry

navidcy (approved):

src/DataWrangling/EN4/EN4_metadata.jl
```diff
@@ -0,0 +1,155 @@
...
+        return EN4_url_pre2021 * year * ".zip"
+    else
+        return EN4_url_post2021 * year * ".zip"
+    end
+end
+
+## This function is explicitly for the downloader to check if the zip file/extracted file exists,
+## then to download the relevant URL (from above)
+
+function metadata_path_EN4(m::Metadata{<:EN4Monthly})
```
simone-silvestri, line 103:
```suggestion
function metadata_path(m::Metadata{<:EN4Monthly})
```

src/DataWrangling/EN4/EN4_metadata.jl
```diff
@@ -0,0 +1,155 @@
...
+    close(zarchive)
+end
+
+function download_dataset(metadata::Metadata{<:EN4Monthly})
+    dir = metadata.dir
+    missingzips = []
+
+    # Create a temporary directory to store the .netrc file
+    # The directory will be deleted after the download is complete
+    @root mktempdir(dir) do tmp
```
simone-silvestri, line 134:
```suggestion
@root begin
```

src/DataWrangling/EN4/EN4_metadata.jl
```diff
@@ -0,0 +1,155 @@
...
+        ntasks = Threads.nthreads()
+
+        asyncmap(metadata; ntasks) do metadatum # Distribute the download among tasks
+            fileurl = metadata_url(metadatum)
+            zippath, extracted_file = metadata_path_EN4(metadatum)
+            if !isfile(extracted_file) & !isfile(zippath)
+                push!(missingzips, zippath)
+                instructions_msg = "\n See ClimaOcean.jl/src/DataWrangling/EN4/README.md for instructions."
```
simone-silvestri, line 142:
```suggestion
```

src/DataWrangling/dataset.jl
```diff
@@ -0,0 +1,159 @@
...
+    Field(metadata::Metadatum;
+          architecture = CPU(),
+          inpainting = nothing,
+          mask = nothing,
+          horizontal_halo = (7, 7),
+          cache_inpainted_data = false)
+
+Return a `Field` on `architecture` described by `metadata` with `horizontal_halo` size.
+If not `nothing`, the `inpainting` method is used to fill the cells
+within the specified `mask`. `mask` is set to `Dataset_mask` for non-nothing
```
simone-silvestri, line 52:
```suggestion
within the specified `mask`. `mask` is set to `dataset_mask` for non-nothing
```

navidcy:
incorporated https://github.com/CliMA/ClimaOcean.jl/pull/472

navidcy:
thanks @simone-silvestri for the suggestions!

navidcy:
thanks @taimoorsohail!
I'm hoping that all tests will pass now!

simone-silvestri (approved):
looks good!

src/DataWrangling/masking.jl
```diff
@@ -157,3 +228,45 @@ function inpaint_mask!(field, mask; inpainting=NearestNeighborInpainting(Inf))
...
+function dataset_immersed_grid(metadata, architecture = CPU())
+
+    mask = dataset_mask(metadata, architecture)
+    grid = mask.grid
+    bottom = Field{Center, Center, Nothing}(grid)
+
+    # Set the mask with zeros where field is defined
+    launch!(architecture, grid, :xy, _set_height_from_mask!, bottom, grid, mask)
+
+    return ImmersedBoundaryGrid(grid, GridFittedBottom(bottom))
```
simone-silvestri, line 271: Will this work? the mask is 3D while `bottom` is 2D. Probably we need to just
```suggestion
    return ImmersedBoundaryGrid(grid, GridFittedBoundary(mask))
```

navidcy: This is on `main`:

https://github.com/CliMA/ClimaOcean.jl/blob/46487980688544cdffe88acc86d47ce42f162909/src/DataWrangling/ECCO/ECCO_mask.jl#L59

So let's leave this for a different PR. I think there is an example that uses `ECCO_immersed_grid()` but the docs are broken atm.

simone-silvestri: Oh I haven't seen the `_set_height_from_mask!` kernel, no this would definitely work!

PR #472, opened 2025-04-17
Adjust below freezing temperatures at all cells including surface

glwagner:
I think we want to include the surface cell in this. We _additionally_ want to adjust cells that contact sea ice at the surface.


glwagner:
For example we also want to adjust cells which are below freezing even if the ice concentration is 0.

Issue #471, opened 2025-04-16
`first_data()` call in ECCO_metadata.jl


michel2323:
I get 

```julia
ERROR: MethodError: no method matching first_date(::ECCO4Monthly)

Closest candidates are:
  first_date(::Any, ::Any)
   @ ClimaOcean ~/.julia/dev/ClimaOcean/src/DataWrangling/metadata.jl:143

Stacktrace:
 [1] (Metadata{<:Union{…} where {…}, <:Union{…}})(name::Symbol)
   @ ClimaOcean.DataWrangling.ECCO ~/.julia/dev/ClimaOcean/src/DataWrangling/ECCO/ECCO_metadata.jl:31
 [2] top-level scope
   @ ~/git/Oceanrora.jl/near_global_ocean_simulation.jl:88
Some type information was truncated. Use `show(err)` to see complete types.

ClimaOcean.DataWrangling.ECCO.ECCOMetadatum
```

I fixed it with
```julia
function ECCOMetadatum(name;
                       # from first_date(ECCO4Monthly())
                       # to
                       date = first_date(ECCO4Monthly(), name),
                       dir = download_ECCO_cache)
```
but not sure that's correct.

glwagner:
that looks right! PR welcome! I guess all of our tests provide the date explicitly

navidcy:
Seems like `first_date` is _not_ called by the tests at all.

navidcy:
@michel2323 the suggestion you provide is correct and should be implemented.

But inspired from this issue I'm also wondering if it's useful to add the convenience:

```Julia
first_date(metadata) = first(all_dates(metadata))
```

PR #474, opened 2025-04-17
Fix first_date args

michel2323:
Fixes https://github.com/CliMA/ClimaOcean.jl/issues/471

### DIFF
--- a/src/DataWrangling/ECCO/ECCO_metadata.jl
+++ b/src/DataWrangling/ECCO/ECCO_metadata.jl
@@ -12,7 +12,6 @@ import Oceananigans.Fields: location
 import ClimaOcean.DataWrangling: all_dates, metadata_filename, download_dataset,
                                  default_download_directory, dataset_temperature_units,
                                  short_name, dataset_latitude_extent
-import ClimaOcean.DataWrangling.ECCO: download_ECCO_cache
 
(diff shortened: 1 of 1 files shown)


navidcy (approved):

src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -12,6 +12,7 @@ import Oceananigans.Fields: location
 import ClimaOcean.DataWrangling: all_dates, metadata_filename, download_dataset,
                                  default_download_directory, dataset_temperature_units,
                                  short_name, dataset_latitude_extent
+import ClimaOcean.DataWrangling.ECCO: download_ECCO_cache
```
navidcy, line 15: 1/ is this needed?
2/ do we need to import it or using it?

I believe it's not needed since it was defined in ECCO.jl, right?

navidcy, line 15:
```suggestion
```

navidcy:
Tests will fail since they are coming from a fork and don't have access to the repo's secrets to download ECCO. I tested this and it's fine. I think we should merge.

glwagner:
thanks @michel2323

PR #475, opened 2025-04-17
Better interpolation info shown to user

### LINKED ISSUE #476: Bathymetry info messages for skipping interpolation are tad misleading
There was a discussion that started regarding this in https://github.com/CliMA/ClimaOcean.jl/pull/388 (see https://github.com/CliMA/ClimaOcean.jl/pull/388#issuecomment-2709058675) but we haven't done anything about it.

After looking into it I think what the code is doing is correct (that is, what we'd expect it to do) but the info messages to the user are misleading.

The message would often be:

```julia
┌ Info: Skipping passes for interpolating bathymetry of size (21600, 10800, 1) 
│ to target grid of size (360, 180, 100). Interpolation passes may only 
│ be used to coarsen bathymetry and require that the bathymetry 
└ is finer than the target grid in both horizontal directions.
```

But what is meant to say here is that it's skipping the _extra_ interpolation passes that were passed by the user via `interpolation_passes` kwarg. There is _always_ one interpolation pass.

I refactored the messaging to the user in #475

cc @taimoorsohail

navidcy:
Closes #476

Also gives more informative message to the user regarding why the extra interpolation passes are skipped.

```Julia
using ClimaOcean
using Oceananigans

underlying_grid = TripolarGrid(size = (360, 180, 50),
                               z = (-5000, 0),
                               halo = (5, 5, 4),
                               first_pole_longitude = 70,
                               north_poles_latitude = 55)
```
```Julia
bottom_height = regrid_bathymetry(underlying_grid;
                                  major_basins = 2)
[ Info: Interpolating bathymetry of size (21600, 10800, 1) onto a TripolarGrid target grid of size (360, 180, 50)
360×180×1 Field{Center, Center, Nothing} reduced over dims = (3,) on OrthogonalSphericalShellGrid on CPU
├── grid: 360×180×50 OrthogonalSphericalShellGrid{Float64, Periodic, Oceananigans.Grids.RightConnected, Bounded} on CPU with 5×5×4 halo and with precomputed metrics
├── boundary conditions: FieldBoundaryConditions
│   └── west: Periodic, east: Periodic, south: ZeroFlux, north: Zipper, bottom: Nothing, top: Nothing, immersed: Default
└── data: 370×190×1 OffsetArray(::Array{Float64, 3}, -4:365, -4:185, 1:1) with eltype Float64 with indices -4:365×-4:185×1:1
    └── max=6102.41, min=-9452.9, mean=-1960.54
```
```Julia
bottom_height_with_extra_passes = regrid_bathymetry(underlying_grid;
                                                    interpolation_passes = 6,
                                                    major_basins = 2)
[ Info: Interpolation passes of bathymetry size (21600, 10800, 1) onto a TripolarGrid target grid of size (360, 180, 50):
[ Info:     pass 1 to size (18060, 9030, 1)
[ Info:     pass 2 to size (14520, 7260, 1)
[ Info:     pass 3 to size (10980, 5490, 1)
[ Info:     pass 4 to size (7440, 3720, 1)
[ Info:     pass 5 to size (3900, 1950, 1)
[ Info:     pass 6 to size (360, 180, 1)
360×180×1 Field{Center, Center, Nothing} reduced over dims = (3,) on OrthogonalSphericalShellGrid on CPU
├── grid: 360×180×50 OrthogonalSphericalShellGrid{Float64, Periodic, Oceananigans.Grids.RightConnected, Bounded} on CPU with 5×5×4 halo and with precomputed metrics
├── boundary conditions: FieldBoundaryConditions
│   └── west: Periodic, east: Periodic, south: ZeroFlux, north: Zipper, bottom: Nothing, top: Nothing, immersed: Default
└── data: 370×190×1 OffsetArray(::Array{Float64, 3}, -4:365, -4:185, 1:1) with eltype Float64 with indices -4:365×-4:185×1:1
    └── max=6005.79, min=-9210.06, mean=-1960.37
```

cc @taimoorsohail

### DIFF
--- a/src/Bathymetry.jl
+++ b/src/Bathymetry.jl
@@ -121,7 +121,7 @@ function regrid_bathymetry(target_grid;
                            dir = download_bathymetry_cache,
                            url = etopo_url,
                            filename = "ETOPO_2022_v1_60s_N90W180_surface.nc",
-                           interpolation_passes = 1,
+                           smoothing = InterpolationPasses(0),
                            major_basins = 1) # Allow an `Inf` number of "lakes"
 
     if isinteger(interpolation_passes)
--- a/src/Bathymetry.jl
+++ b/src/Bathymetry.jl
@@ -265,10 +265,10 @@ function interpolate_bathymetry_in_passes(native_z, target_grid;
     old_z  = native_z
     TX, TY = topology(target_grid)
 
+    @info "Interpolation passes of bathymetry size $(size(old_z)) onto a $gridtype target grid of size $Nt:"
     for pass = 1:passes - 1
         new_size = (Nλ[pass], Nφ[pass], 1)
-
-        @info "Bathymetry extra interpolation pass $pass with size $new_size"
+        @info "    pass $pass to size $new_size"
 
         new_grid = LatitudeLongitudeGrid(architecture(target_grid), Float32,
                                          size = new_size,
@@ -283,7 +283,8 @@ function interpolate_bathymetry_in_passes(native_z, target_grid;
         old_z = new_z
     end
 
-    @info "Interpolating bathymetry of size $Nn onto a $gridtype target grid of size $Nt"
+    new_size = (Nλ[passes], Nφ[passes], 1)
+    @info "    pass $passes to size $new_size"
     target_z = Field{Center, Center, Nothing}(target_grid)
     interpolate!(target_z, old_z)
 
--- a/src/Bathymetry.jl
+++ b/src/Bathymetry.jl
@@ -216,7 +216,8 @@ function interpolate_bathymetry_in_passes(native_z, target_grid;
 
         gridtype = target_grid isa TripolarGrid ? "TripolarGrid" :
                    target_grid isa LatitudeLongitudeGrid ? "LatitudeLongitudeGrid" :
-                   target_grid isa RectilinearGrid ? "RectilinearGrid" : error("unknown type of target grid")
+                   target_grid isa RectilinearGrid ? "RectilinearGrid" :
+                   error("unknown target grid type")
 
         @info "Interpolating bathymetry onto a $gridtype target grid of size $(size(target_grid))"
         interpolate!(target_z, native_z)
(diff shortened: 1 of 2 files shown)


navidcy:
Since there is always 1 interpolation pass it makes no sense to be able to provide `intepolation_passes = 0` or `interpolation_passes = nothing`. This misleads the user thinking that there won't be any interpolation.

glwagner:
lets do this --- but do we agree that we should actually eliminate this code in favor of better bathymetry smoothing methods?

glwagner:
I'd advise against working on input validation since it is potentially wasted effort

glwagner (approved):

glwagner:
The problem is that the first interpolation is not a "pass". The first interpolation is necessary. What happens after is smoothing. We should not conflate the necessary step with the smoothing. Instead, we need an interface for smoothing. Implementing an interface for bathymetry smoothing is far preferred to complicated validation of a confusing interface in my opinion.

src/Bathymetry.jl
```diff
@@ -126,6 +124,14 @@ function regrid_bathymetry(target_grid;
                            interpolation_passes = 1,
```
glwagner, line 124:
```suggestion
                           smoothing = InterpolationPasses(0),
```

navidcy:
> The problem is that the first interpolation is not a "pass".
Yes, that was part of the misleading messaging; this PR fixes that.

> The first interpolation is necessary. What happens after is smoothing. We should not conflate the necessary step with the smoothing.

> Instead, we need an interface for smoothing.

Totally agree. Let's merge this and work on something like #393. What's wrong with having better messaging until we fix it. The only argument is perhaps wasted of time. But now I've done it, I wasted that time unfortunately.. :(

> Implementing an interface for bathymetry smoothing is far preferred to complicated validation of a confusing interface in my opinion.

Agree

src/Bathymetry.jl
navidcy: ok, you are suggesting we do the refactor of the bathymetry in this PR?

navidcy:
closing this in favor of #478

navidcy:
> lets do this --- but do we agree that we should actually eliminate this code in favor of better bathymetry smoothing methods?

totally agree
and your idea of applying the smoothing to the native grid of the bathymetry _before_ interpolate make sense

PR #479, opened 2025-04-20
Fixes bugs for ECCO2, adds some more love to ECCO2, and  tests

### LINKED ISSUE #470: Bug for ECCO2 dataset metadata
Definitely a bug here

https://github.com/CliMA/ClimaOcean.jl/blob/b4f2e5687eba3e86a85e67661fc6a3e03b2d7091/src/DataWrangling/ECCO/ECCO_metadata.jl#L133-L135

I can fix it but it also points to the need to include the ECCO2 datasets in the test_downloading.jl and/or test_ecco.jl.

navidcy:
Closes #470


src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -73,15 +75,15 @@ function metadata_filename(metadata::Metadatum{<:ECCO4Monthly})
...
 
 function metadata_filename(metadata::Metadatum{<:Union{ECCO2Daily, ECCO2Monthly}})
-    shortname   = short_name(metadata)
-    yearstr  = string(Dates.year(metadata.dates))
-    monthstr = string(Dates.month(metadata.dates), pad=2)
-    postfix = variable_is_three_dimensional(metadata) ? ".1440x720x50." : ".1440x720."
+    shortname = short_name(metadata)
+    yearstr   = string(Dates.year(metadata.dates))
+    monthstr  = string(Dates.month(metadata.dates), pad=2)
+    postfix   = variable_is_three_dimensional(metadata) ? ".1440x720x50." : ".1440x720."
```
glwagner, line 81: i like "postfix" vs "suffix"

glwagner (approved):

src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -56,9 +58,9 @@ Base.size(::Metadatum{<:ECCO2Monthly}) = (1440, 720, 50, 1)
 Base.size(::Metadatum{<:ECCO4Monthly}) = (720,  360, 50, 1)
 
 # The whole range of dates in the different dataset datasets
-all_dates(::ECCO4Monthly, name) = DateTime(1992, 1, 1) : Month(1) : DateTime(2023, 12, 1)
-all_dates(::ECCO2Monthly, name) = DateTime(1992, 1, 1) : Month(1) : DateTime(2023, 12, 1)
-all_dates(::ECCO2Daily, name)   = DateTime(1992, 1, 4) : Day(1)   : DateTime(2023, 12, 31)
+all_dates(::ECCO4Monthly, name) = DateTime(1992, 1, 1) : Month(1) : DateTime(2017, 12, 1)
```
navidcy, line 61: ECCO4 goes up to 2017? I couldn’t see later files in the website..

navidcy:
I don't understand why tests hang. They pass on my laptop...
I might split them into four different sets; one for each dataset we test.

src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -150,6 +152,10 @@ function download_dataset(metadata::ECCOMetadata)
     password = get(ENV, "ECCO_PASSWORD", nothing)
     dir = metadata.dir
 
+    if !isdir(dir)
+        mkpath(dir)
+    end
```
simone-silvestri, line 157: Should we make a directory automatically? I would probably prefer making sure that the users make sure the directory exists rather than automatically touching the filesystem.

simone-silvestri: i.e., just spit an informative error if the directory does not exist

glwagner: I agree. Also using `dir` should be discouraged.

navidcy: I see your point!

I was doing that because I thought we'd better have the ECCO2 and ECCO4 data in different directories. Also different directories for ECCO2 monthly vs daily. 

But perhaps it's better to have 3 different scratch dirs?

https://github.com/CliMA/ClimaOcean.jl/blob/58065e8516db6272699d6530f34b095886c84c11/src/DataWrangling/ECCO/ECCO.jl#L31-L34

glwagner: I think if we want to look at the data, then we need utilities to copy or create symlinks to it. I think it's also not proper to go to the scratch directory. In summary:

1. Always download data to scratch for simulations. This ensures that data is not downloaded repeatedly.
2. If you would like to look at data in a way that is not accessible through ClimaOcean utilities, then copy from scratch to a local location. (Or, implement the feature that is missing in ClimaOcean.)

navidcy: I agree!

But I'm a bit confused. I don't wanna look at any data. Just want to ensure that data is not being overwritten. E.g. both daily and monthly ECCO2 SSH has the same filename (see https://ecco.jpl.nasa.gov/drive/files/ECCO2/cube92_latlon_quart_90S90N/daily/SSH and https://ecco.jpl.nasa.gov/drive/files/ECCO2/cube92_latlon_quart_90S90N/monthly/SSH). Are we sure that if monthly is downloaded after the daily one doesn't overwrite the other?

I was just suggesting above that the cleanest-safest way is that each dataset has its own scratch directory. Perhaps I didn't explain myself very well?

glwagner: oh well for sure if they have the same name then I think you are right, different directories are needed. I didn't realize that part!

glwagner: I didn't understand because you said "we'd better have them in different directories" rather than "we need them in different directories [because the filenames overlap]"

navidcy: oh no worries

so what do you think:

- 3 different scratch directories?
or
- 1 scratch directory for ECCO and subdirectories for ECCO2, ECCO4 / daily, monthly?

src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -73,15 +75,15 @@ function metadata_filename(metadata::Metadatum{<:ECCO4Monthly})
...
+    shortname = short_name(metadata)
+    yearstr   = string(Dates.year(metadata.dates))
+    monthstr  = string(Dates.month(metadata.dates), pad=2)
+    postfix   = variable_is_three_dimensional(metadata) ? ".1440x720x50." : ".1440x720."
 
     if metadata.dataset isa ECCO2Monthly
         return shortname * postfix * yearstr * monthstr * ".nc"
     elseif metadata.dataset isa ECCO2Daily
-        daystr = string(Dates.day(metadata.dates), pad=2)
+        daystr = variable_is_three_dimensional(metadata) ? string(Dates.day(metadata.dates), pad=2) : ""
```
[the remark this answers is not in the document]
navidcy: the sea ice variables exists but their filename does not have day in it; you are correct.
but this is dealt with `variable_is_three_dimensional(metadata)` since that returns true only for u, v, T, and S.

src/DataWrangling/ECCO/ECCO_metadata.jl
navidcy: > Should we make a directory automatically? I would probably prefer having the users make sure the directory exists rather than automatically touching the filesystem.

see [2205be1](https://github.com/CliMA/ClimaOcean.jl/pull/479/commits/2205be174eeec348010b396dc57c1de3d8f7eb33) does the former

src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -37,9 +37,24 @@ function ECCOMetadatum(name;
     return Metadatum(name; date, dir, dataset=ECCO4Monthly())
 end
 
-default_download_directory(::Union{<:ECCO2Monthly}) = joinpath(download_ECCO_cache, "v2", "monthly")
-default_download_directory(::Union{<:ECCO2Daily})   = joinpath(download_ECCO_cache, "v2", "daily")
-default_download_directory(::Union{<:ECCO4Monthly}) = joinpath(download_ECCO_cache, "v4")
+function default_download_directory(::Union{<:ECCO2Monthly})
+    path = joinpath(download_ECCO_cache, "v2", "monthly")
+    !isdir(path) && mkpath(dir)
```
glwagner:
```suggestion
    mkpath(dir)
```

the docstring says "Unlike mkdir, mkpath does not error if path
  (or parts of it) already exists."

src/DataWrangling/ECCO/ECCO_metadata.jl
```diff
@@ -37,9 +37,24 @@ function ECCOMetadatum(name;
...
+    return path
```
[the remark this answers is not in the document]
navidcy: Thanks!

navidcy:
I'm trying to debug the tests for the cycling boundaries and I've noticed this:

https://github.com/CliMA/ClimaOcean.jl/blob/09f492fe3fbe784cb884d849f282c0efbf281b11/src/DataWrangling/metadata.jl#L178

Why do we start at `start_idx - 1` if `start_idx > 1`?

I'd think that the proper thing to do here is that we start from `start_idx - 1` _only_ if also `native_dates[start_idx] > start_date)`, i.e., 


```julia
start_idx = (start_idx > 1 && native_dates[start_idx] > start_date) ? start_idx - 1 : start_idx
```

I'm changing it! But just wanted to double check that I'm not missing something.

cc @simone-silvestri, @glwagner

navidcy:
I actually think it's a bug because if we are asking for a forcing dataset from a data that is already in the dataset, then the forcing dataset will start from the previous time-instance.

navidcy:
I think tests fails because two or more builds try to access and/or write on the same files. E.g the inpainted ones..

src/DataWrangling/metadata.jl
```diff
@@ -26,15 +26,28 @@ Keyword Arguments
...
 function Metadata(variable_name;
                   dataset,
                   dates = all_dates(dataset, variable_name),
-                  dir = default_download_directory(dataset))
+                  dir = default_download_directory(dataset),
+                  start_date = nothing,
+                  end_date = nothing)
+
+    if !isnothing(start_date) && !isnothing(end_date)
+        @info "Slicing date range within $start_date and $end_date"
```
glwagner: I'm worried we are getting carried away with info statements, ClimaOcean is really chatty right now

navidcy: fair, deleting it

navidcy:
Just pointing out that this PR initial purpose is to make ECCO2 work. So I hope to merge this now that it does and then work on Metadata API in a different PR.

glwagner (approved):

navidcy:
I'm merging.

Note: the ECCO2 tests on CPU take quite longer than on GPU. I've noticed that most of the time is when inpainting with Inf happens. Locally that's not the case.

PR #480, opened 2025-04-23
Generalize/Refactor ECCOrestoring 👉🏽 DatasetRestoring to work with any dataset

taimoorsohail:
This PR introduces a general restoring file that replaces `ECCORestoring` and associated functions with a general `Restoring` function that works with both ECCO and EN4.


src/DataWrangling/restoring.jl
```diff
@@ -19,50 +20,50 @@ import Oceananigans.OutputReaders: new_backend, update_field_time_series!
...
 
 """
-    ECCONetCDFBackend(length, metadata;
+    NetCDFBackend(length, metadata;
                       on_native_grid = false,
                       cache_inpainted_data = false,
                       inpainting = NearestNeighborInpainting(Inf))
 
-Represent an ECCO FieldTimeSeries backed by ECCO native netCDF files.
+Represent an observational/reanalysis FieldTimeSeries backed by native netCDF files.
```
navidcy, line 44:
```suggestion
Represent a FieldTimeSeries backed by native netCDF files.
```

src/DataWrangling/restoring.jl
```diff
@@ -19,50 +20,50 @@ import Oceananigans.OutputReaders: new_backend, update_field_time_series!
...
+Base.length(backend::NetCDFBackend)  = backend.length
+Base.summary(backend::NetCDFBackend) = string("NetCDFBackend(", backend.start, ", ", backend.length, ")")
 
-const ECCOFieldTimeSeries{N} = FlavorOfFTS{<:Any, <:Any, <:Any, <:Any, <:ECCONetCDFBackend{N}} where N
+const ObsFieldTimeSeries{N} = FlavorOfFTS{<:Any, <:Any, <:Any, <:Any, <:NetCDFBackend{N}} where N
 
-new_backend(b::ECCONetCDFBackend{native, cache_data}, start, length) where {native, cache_data} =
-    ECCONetCDFBackend{native, cache_data}(start, length, b.inpainting, b.metadata)
+new_backend(b::NetCDFBackend{native, cache_data}, start, length) where {native, cache_data} =
+NetCDFBackend{native, cache_data}(start, length, b.inpainting, b.metadata)
```
navidcy, line 61:
```suggestion
new_backend(b::NetCDFBackend{native, cache_data}, start, length) where {native, cache_data} =
    NetCDFBackend{native, cache_data}(start, length, b.inpainting, b.metadata)
```

src/DataWrangling/restoring.jl
```diff
@@ -19,50 +20,50 @@ import Oceananigans.OutputReaders: new_backend, update_field_time_series!
...
 
-Adapt.adapt_structure(to, b::ECCONetCDFBackend{N, C}) where {N, C} = ECCONetCDFBackend{N, C}(b.start, b.length, nothing, nothing)
+Adapt.adapt_structure(to, b::NetCDFBackend{N, C}) where {N, C} = NetCDFBackend{N, C}(b.start, b.length, nothing, nothing)
 
 """
-    ECCONetCDFBackend(length, metadata;
+    NetCDFBackend(length, metadata;
                       on_native_grid = false,
                       cache_inpainted_data = false,
                       inpainting = NearestNeighborInpainting(Inf))
```
navidcy, line 42: note the alignment

```suggestion
    NetCDFBackend(length, metadata;
                  on_native_grid = false,
                  cache_inpainted_data = false,
                  inpainting = NearestNeighborInpainting(Inf))
```

src/DataWrangling/restoring.jl
navidcy, line 44: I guess it can also be restoring to some model output somebody outputted.

So perhaps just "Represent a FieldTimeSeries backed by native ..."

navidcy, line 44: I guess it can also be restoring to some model output somebody outputted.

So perhaps just

```suggestion
Represent a FieldTimeSeries backed by native netCDF files.
```

src/DataWrangling/restoring.jl
```diff
@@ -19,50 +20,50 @@ import Oceananigans.OutputReaders: new_backend, update_field_time_series!
...
+    return NetCDFBackend{on_native_grid, cache_inpainted_data}(1, length, inpainting, metadata)
 end
 
-Base.length(backend::ECCONetCDFBackend)  = backend.length
-Base.summary(backend::ECCONetCDFBackend) = string("ECCONetCDFBackend(", backend.start, ", ", backend.length, ")")
+Base.length(backend::NetCDFBackend)  = backend.length
+Base.summary(backend::NetCDFBackend) = string("NetCDFBackend(", backend.start, ", ", backend.length, ")")
 
-const ECCOFieldTimeSeries{N} = FlavorOfFTS{<:Any, <:Any, <:Any, <:Any, <:ECCONetCDFBackend{N}} where N
+const ObsFieldTimeSeries{N} = FlavorOfFTS{<:Any, <:Any, <:Any, <:Any, <:NetCDFBackend{N}} where N
```
navidcy, line 58: why `Obs` in the name here? The name implies that it can be only observations? But it can be anything really, right?

navidcy:
This seems great! 

Noting that we should also change the tests that now use `ECCORestoring` to use `Restoring` with different datasets.

src/DataWrangling/restoring.jl
taimoorsohail: Good point! I'll import the general function here

navidcy: Yes, I think we just want to extend the `FieldTimeSeries` method to work with Metadata.

taimoorsohail: Latest commit extends this (https://github.com/CliMA/ClimaOcean.jl/pull/480/commits/94f22ec21d3c60085aa6d983bd0566dd5d2bf768)

taimoorsohail:
I've fixed the ECCO and EN4 tests, they are all working now. However, `test/test_diagnostics.jl` is failing at line 45... I'm not sure what this test is aiming to achieve?

navidcy:
let me have a look at that

navidcy:
Seems that the test_diagnostics.jl hardcodes some values specific for the ECCO4Monthly() dataset. Thus, I would expect that the test_diagnostics.jl pass for ECCO4Monthly() dataset and possibly fail for EN4Monthly(). However, currently they fail for ECCO4 also which hints that something might be fishy in the refactor? I'll have a closer look!

src/DataWrangling/restoring.jl
```diff
@@ -19,50 +20,48 @@ import Oceananigans.OutputReaders: new_backend, update_field_time_series!
...
+                  on_native_grid = false,
+                  cache_inpainted_data = false,
+                  inpainting = NearestNeighborInpainting(Inf))
 
-Represent an ECCO FieldTimeSeries backed by ECCO native netCDF files.
+Represent a FieldTimeSeries backed by native netCDF files.
 Each time instance is stored in an individual file.
 """
-function ECCONetCDFBackend(length, metadata;
+function NetCDFBackend(length, metadata;
```
navidcy, line 47: code alignment is clunky here

(this is a generic comment, applies to many places...)

src/DataWrangling/restoring.jl
```diff
@@ -79,21 +78,21 @@ function set!(fts::ECCOFieldTimeSeries)
...
 """
-    ECCOFieldTimeSeries(metadata::ECCOMetadata [, arch_or_grid=CPU() ];
+    FieldTimeSeries(metadata::Metadata [, arch_or_grid=CPU() ];
                         time_indices_in_memory = 2,
                         time_indexing = Cyclical(),
                         inpainting = nothing,
                         cache_inpainted_data = true)
 
-Create a field time series object for ECCO data.
+Create a field time series object for obs/reanalysis data.
```
navidcy, line 87: don't wanna mention "obs/reanalysis", right?
perhaps something along the lines of:

```suggestion
Create a FieldTimeSeries from a dataset that corresponds to `metadata`..
```

src/DataWrangling/restoring.jl
```diff
@@ -283,30 +281,30 @@ Keyword Arguments
 
 - `rate`: The restoring rate, i.e., the inverse of the restoring timescale (in s⁻¹).
 
-- `dir`: The directory where the native ECCO data is located. If the data does not exist it will
-         be automatically downloaded. Default: `download_ECCO_cache`.
+- `dir`: The directory where the native obs/reanalysis data is located. If the data does not exist it will
+         be automatically downloaded. No default, must be provided.
```
navidcy, line 285:
```suggestion
- `dir` (required): The directory where the native data is located. If the data does not exist it will
         be automatically downloaded.
```

src/DataWrangling/restoring.jl
```diff
@@ -1,8 +1,9 @@
 using Oceananigans: location
 using Oceananigans.Grids: AbstractGrid, node, on_architecture
 using Oceananigans.Fields: interpolate!, interpolate, location, instantiated_location
-using Oceananigans.OutputReaders: Cyclical, TotallyInMemory, AbstractInMemoryBackend, FlavorOfFTS, time_indices
+using Oceananigans.OutputReaders: Cyclical, TotallyInMemory, AbstractInMemoryBackend, FlavorOfFTS, time_indices, FieldTimeSeries
```
navidcy, line 4: why `using Oceananigans.OutputReaders: FieldTimeSeries` and not `import`? Don't we want to add a method?

navidcy:
> Seems that the test_diagnostics.jl hardcodes some values specific for the ECCO4Monthly() dataset. Thus, I would expect that the test_diagnostics.jl pass for ECCO4Monthly() dataset and possibly fail for EN4Monthly(). However, currently they fail for ECCO4 also which hints that something might be fishy in the refactor? I'll have a closer look!

There must be a bug because when I run test_diagnostics line by line on `main` I get

```Julia
julia> Tt = ECCOFieldTimeSeries(Tmeta, grid; time_indices_in_memory=2)
3×3×100×2 FieldTimeSeries{ClimaOcean.DataWrangling.ECCO.ECCONetCDFBackend} located at (Center, Center, Center) on CPU
├── grid: 3×3×100 ImmersedBoundaryGrid{Float64, Bounded, Bounded, Bounded} on CPU with 3×3×3 halo
├── indices: (:, :, :)
├── time_indexing: Cyclical(period=5.3568e6)
├── backend: ECCONetCDFBackend(1, 2)
└── data: 9×9×106×2 OffsetArray(::Array{Float64, 4}, -2:6, -2:6, -2:103, 1:2) with eltype Float64 with indices -2:6×-2:6×-2:103×1:2
    └── max=28.7566, min=0.0, mean=2.68104
```

while on this PR I get: 

```Julia
julia> Tt = FieldTimeSeries(Tmeta, grid; time_indices_in_memory=2)
3×3×100×2 FieldTimeSeries{ClimaOcean.DataWrangling.NetCDFBackend} located at (Center, Center, Center) on CPU
├── grid: 3×3×100 ImmersedBoundaryGrid{Float64, Bounded, Bounded, Bounded} on CPU with 3×3×3 halo
├── indices: (:, :, :)
├── time_indexing: Cyclical(period=5.3568e6)
├── backend: NetCDFBackend(1, 2)
└── data: 9×9×106×2 OffsetArray(::Array{Float64, 4}, -2:6, -2:6, -2:103, 1:2) with eltype Float64 with indices -2:6×-2:6×-2:103×1:2
    └── max=0.0, min=0.0, mean=0.0
```

So on this PR seems like `Tt` is just zeros?

src/DataWrangling/restoring.jl
taimoorsohail: Hm, you are right, but I'm not sure why it is working... Either way, I have moved it to the import line...

taimoorsohail:
OK, looks like the wrong `set!` function was being called (from `Oceananigans`) - that's been fixed now.

navidcy (approved):

src/DataWrangling/restoring.jl
```diff
@@ -255,23 +252,21 @@ Arguments
...
 Keyword Arguments
 =================
 
-- `dataset`: The dataset of the ECCO dataset. Default: `ECCO4Monthly()`.
-
-- `start_date`: The starting date to use for the ECCO dataset. Default: `first_date(dataset, variable_name)`.
+- `dataset`: The dataset; required kwarg if `variable_name` argument is provide.
 
-- `end_date`: The ending date to use for the ECCO dataset. Default: `end_date(dataset, variable_name)`.
+- `dates`: The date range to use for the dataset. Default: `all_dates(dataset, variable_name)`.
```
simone-silvestri, line 269:
```suggestion
- `start_date`: The starting date to use for the ECCO dataset. Default: `first_date(dataset, variable_name)`.

- `end_date`: The ending date to use for the ECCO dataset. Default: `end_date(dataset, variable_name)
```

test/test_diagnostics.jl
```diff
@@ -3,11 +3,15 @@ include("runtests_setup.jl")
 using SeawaterPolynomials: TEOS10EquationOfState
 using Oceananigans.BuoyancyFormulations: buoyancy
 using Oceananigans: location
-using ClimaOcean.DataWrangling.ECCO: ECCOFieldTimeSeries
+using ClimaOcean.DataWrangling: FieldTimeSeries
```
simone-silvestri, line 6:
```suggestion
```
we don't need this line if we extend from Oceananigans

simone-silvestri (approved):
looks good

examples/one_degree_simulation.jl
```diff
@@ -61,8 +61,8 @@ z_below_surface = r_faces[end-1]
 
 mask = LinearlyTaperedPolarMask(southern=(-80, -70), northern=(70, 90), z=(z_below_surface, 0))
 
-FT = ECCORestoring(temperature, grid; mask, rate=restoring_rate)
-FS = ECCORestoring(salinity,    grid; mask, rate=restoring_rate)
+FT = Restoring(temperature, grid; mask, rate=restoring_rate)
+FS = Restoring(salinity,    grid; mask, rate=restoring_rate)
```
glwagner, line 65:
```suggestion
FS = DatasetRestoring(salinity,    grid; mask, rate=restoring_rate)
```

would this be a better name?

"dataset" is provided by context. However, "restoring" is a more general term that would also apply to an idealized restoring. As an example, we often do need to damp velocities to zero as well, but we don't need data for that.

src/DataWrangling/restoring.jl
```diff
@@ -3,66 +3,66 @@ using Oceananigans.Grids: AbstractGrid, node, on_architecture
...
 import Oceananigans.Fields: set!
 import Oceananigans.Forcings: regularize_forcing
-import Oceananigans.OutputReaders: new_backend, update_field_time_series!
+import Oceananigans.OutputReaders: new_backend, update_field_time_series!, FieldTimeSeries
 
 @inline instantiate(T::DataType) = T()
 @inline instantiate(T) = T
 
-struct ECCONetCDFBackend{N, C, I, M} <: AbstractInMemoryBackend{Int}
+struct NetCDFBackend{N, C, I, M} <: AbstractInMemoryBackend{Int}
```
glwagner, line 22:
```suggestion
struct DatasetNetCDFBackend{N, C, I, M} <: AbstractInMemoryBackend{Int}
```

I suggest this because we will eventually need to have a "NetCDFBackend" for other types of netcdf files --- eg ones created by Oceananigans. In this case, we probably don't want to re-use the concept of "dataset" (which I think we would like to apply specifically to external / established data products that can be downloaded). 

I'm flexible on the name, just think that we need to be more specific than just "NetCDF"

navidcy: Gotcha! I think makes sense what you suggest!

navidcy:
probably also JRA55NetCDFBackend could be changed to DatasetBackend; but perhaps that's a different PR

navidcy:
I'm bit confused with this error:

https://buildkite.com/clima/climaocean-ci/builds/3677#01966497-8819-41b9-b8be-836fc3d29784/187-552

especially because running the GPU tests on tartarus pass.

Any ideas?

simone-silvestri:
looks like `DataRestoring` is not correctly built on the GPU since it contains CPU data:
```julia
DatasetRestoring{Oceananigans.OutputReaders.GPUAdaptedFieldTimeSeries{Center, Center, Center, Oceananigans.OutputReaders.Cyclical{Float64}, ClimaOcean.DataWrangling.DatasetBackend{ClimaOcean.DataWrangling.on_native_grid, true, Nothing, Nothing}, Float32, OffsetArrays.OffsetArray{Float32, 4, Array{Float32, 4}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, LatitudeLongitudeGrid{Float32, Periodic, Bounded, Bounded, Oceananigans.Grids.StaticVerticalDiscretization{OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}}, Float32, Float32, OffsetArrays.OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, OffsetArrays.OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, Float32, Float32, OffsetArrays.OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, OffsetArrays.OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, Float32, Float32, Nothing, Int64}, Int64, ClimaOcean.DataWrangling.Temperature, Float64}}, which is not a bitstype:
```

It might either be a bug in the constructor which does not account for arch, or in the test that does not pass `arch` or `grid` to the `DatasetRestoring`. Tomorrow I ll take a look

test/test_ecco4_en4.jl
```diff
@@ -133,19 +121,11 @@ end
...
-@testset "ECCO-specific field utilities" begin
-    for arch in test_architectures
-        A = typeof(arch)
+    @testset "Field utilities" begin
         for name in (:temperature, :salinity)
-            @info "Testing ECCO_field on $A..."
-            metadata = Metadata(name; dates, dataset=ECCO4Monthly())
-            restoring = ECCORestoring(metadata; rate=1/1000, inpainting)
+            metadata = Metadata(name; dates, dataset)
+            restoring = DatasetRestoring(metadata; rate=1/1000, inpainting)
```
simone-silvestri, line 128: for example here there is no `arch` nor `grid`

test/test_ecco4_en4.jl
```diff
@@ -194,45 +172,45 @@ end
...
-        @test @allowscalar T_restoring(1, 1,   10, grid, clock, fields) == T_restoring.rate
-        @test @allowscalar T_restoring(1, 11,  10, grid, clock, fields) == T_restoring.rate / 2
-        @test @allowscalar T_restoring(1, 21,  10, grid, clock, fields) == 0
-        @test @allowscalar T_restoring(1, 80,  10, grid, clock, fields) == 0
-        @test @allowscalar T_restoring(1, 90,  10, grid, clock, fields) == T_restoring.rate / 2
-        @test @allowscalar T_restoring(1, 100, 10, grid, clock, fields) == T_restoring.rate
-        @test @allowscalar T_restoring(1, 1,   5,  grid, clock, fields) == 0
-        @test @allowscalar T_restoring(1, 10,  5,  grid, clock, fields) == 0
+        for name in (:temperature, :salinity)
+            var_restoring = DatasetRestoring(name; dataset, start_date, end_date, mask, inpainting, rate=1/1000)
```
simone-silvestri, line 176: Here as well, since the previous `LatitudeLongitudeGrid` accepts an `arch`, we also might want to do
```suggestion
            var_restoring = DatasetRestoring(name, arch; dataset, start_date, end_date, mask, inpainting, rate=1/1000)
```
here

test/test_ecco4_en4.jl
```diff
@@ -194,45 +172,45 @@ end
...
         @test begin
-            set!(field, ECCOMetadatum(:temperature, date=start_date))
-            set!(field, ECCOMetadatum(:salinity,    date=start_date))
+            set!(field, Metadatum(:temperature; dataset, date=start_date))
+            set!(field, Metadatum(:salinity;    dataset, date=start_date))
             true
         end
 
-        forcing_T = ECCORestoring(:temperature, arch; start_date, end_date, inpainting, rate=1/1000)
+        forcing_T = DatasetRestoring(:temperature;  dataset, start_date, end_date, inpainting, rate=1/1000)
```
simone-silvestri, line 213:
```suggestion
        forcing_T = DatasetRestoring(:temperature, arch; dataset, start_date, end_date, inpainting, rate=1/1000)
```

test/test_ecco4_en4.jl
```diff
@@ -257,7 +233,7 @@ end
         end_date = DateTime(1993, 5, 1)
         dates = start_date : Month(1) : end_date
 
-        T_restoring = ECCORestoring(:temperature, arch; start_date, end_date, inpainting, rate=1/1000)
+        T_restoring = DatasetRestoring(:temperature; dataset, start_date, end_date, inpainting, rate=1/1000)
```
simone-silvestri, line 236:
```suggestion
        T_restoring = DatasetRestoring(:temperature, arch; dataset, start_date, end_date, inpainting, rate=1/1000)
```

test/test_ecco4_en4.jl
navidcy: oh I see! we dropped `arch`

wondering how come tests pass on tartarus GPU... I'll fix this

navidcy: Thanks!

test/test_ecco4_en4.jl
navidcy: Thanks!

PR #481, opened 2025-04-23
(0.6.2) Bump minor version and update ClimaSeaIce

simone-silvestri:
I wanted to bump the minor version to capture #467 so I can use EN4 as an initialization for the EGU workshop on Monday. The advantage is that there is no need for setting up a username and password!

And I also update ClimaSeaIce to the main version


simone-silvestri:
Oh! I just figured out that was already captured by 0.6.1!

PR #483, opened 2025-04-24
Bugfix in default inpainting

simone-silvestri:
it is called `sea_ice_concentration` and not `sea_ice_fraction`


navidcy (approved):
You changed the sea ice thickness tho. This contradicts your first post on the pr. Just double checking that it was intentional?

simone-silvestri:
> You changed the sea ice thickness tho. This contradicts your first post on the pr. Just double checking that it was intentional?

good catch! 😅

PR #484, opened 2025-04-24
Correct sea-ice ocean salt flux by concentration

simone-silvestri:
I think this flux should be multipled by the concentration of the sea-ice in the cell?


glwagner (approved):
this seems right to me

PR #486, opened 2025-04-27
DatasetRestoring only via Metadata

### LINKED ISSUE #485: Strange behaviour in `compute_native_date_range`
I've noticed this:

https://github.com/CliMA/ClimaOcean.jl/blob/09f492fe3fbe784cb884d849f282c0efbf281b11/src/DataWrangling/metadata.jl#L178

Why do we start at `start_idx - 1` if `start_idx > 1`?

I'd think that the proper thing to do here is that we start from `start_idx - 1` _only_ if also `native_dates[start_idx] > start_date)`, i.e., 

```julia
start_idx = (start_idx > 1 && native_dates[start_idx] > start_date) ? start_idx - 1 : start_idx
```

I actually believe this is bug because if we are asking for a forcing dataset starting from a date that is already in the dataset, then the forcing dataset will start from the previous time-instance.

_Originally posted by @navidcy in https://github.com/CliMA/ClimaOcean.jl/issues/479#issuecomment-2831846810_

navidcy:
Only allow DatasetRestoring to be constructed via providing Metadata.

Closes #485

### DIFF
--- a/src/DataWrangling/metadata.jl
+++ b/src/DataWrangling/metadata.jl
@@ -47,6 +47,7 @@ function Metadata(variable_name;
     if !isnothing(dates) && !nothing(start_date)
         throw(ArgumentError("both dates and start_date kwargs were provided"))
     end
+
     if !isnothing(dates) && !nothing(end_date)
         throw(ArgumentError("both dates and end_date kwargs were provided"))
     end
@@ -183,7 +184,7 @@ metadata_filename(metadata) = [metadata_filename(metadatum) for metadatum in met
 """
     compute_native_date_range(native_dates, start_date, end_date)
 
-Compute the range of dates that fall within the specified start and end date.
+Compute the range of `native_dates` that fall within the specified `start_date` and `end_date`.
 """
 function compute_native_date_range(native_dates, start_date, end_date)
     if last(native_dates) < end_date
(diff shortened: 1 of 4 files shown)


simone-silvestri:
Isn't it convenient not to have to always build the metadata?
Do you think it is too much of a burden to maintain this constructor? I think it is very convenient to have because if we build only with the metadata we have to know exactly the frequency, start date and last date of the metadata.

I vote to keep the constructor with the variable name

glwagner:
> Isn't it convenient not to have to always build the metadata? Do you think it is too much of a burden to maintain this constructor? I think it is very convenient to have because if we build only with the metadata we have to know exactly the frequency, start date and last date of the metadata.
> 
> I vote to keep the constructor with the variable name

I see this as important for the user interface. I think we may want users to have metadata explicitly available for any dataset being used during a simulation. I think we need to keep developing features for metadata.

We could put this back after some time but I personally think our workflow is not mature enough to have such a convenience constructor. Slowly, slowly...

glwagner:
Should the examples be updated, or do they already all use explicit metadata?

glwagner (approved):

navidcy:
All examples use explicit metadata. I checked.

navidcy:
@simone-silvestri this PR deletes the extra DatasetRestoring constructor but adds convenience in Metadata constructor to provide start and end date as you suggested in #485.

PR #489, opened 2025-04-28
Add support for Metadata of Copernicus datasets

### LINKED ISSUE #488: How does EN4 data naming work?
I noticed that the filename associated with EN4 metadata does not invoke the variable name, eg the name is

https://github.com/CliMA/ClimaOcean.jl/blob/c8c85b57c57772390fb85e7cdcd092ad907c1bea/src/DataWrangling/EN4/EN4_metadata.jl#L61

even though the `short_name` is actually generated in the function (it is not used):

https://github.com/CliMA/ClimaOcean.jl/blob/c8c85b57c57772390fb85e7cdcd092ad907c1bea/src/DataWrangling/EN4/EN4_metadata.jl#L58

Is this a bug?

glwagner:
This PR adds support for downloading and ingesting datasets from the [Copernicus Marine Data Store](https://data.marine.copernicus.eu/products). These include GLORYS12 as well as the CMEMs analysis / forecast (there are other datasets as well).

Full support requires an extension based on PythonCall and CondaPkg to install the command line interface `copernicusmarine`. There is also a need for users to get an account with Copernicus and set environment variables.

Note, it is possible to avoid the dep on PythonCall and invoke `copernicusmarine` directly from the command line using `run`. After some thought, I decided to use the Python interface because it is a bit nicer for users; for example this allows us to access the python docstrings. We can consider splitting up the ext into a python and conda separate ext in the future though.

Closes #488


glwagner:
@navidcy I propose deleting `dataset_immersed_grid`. Instead, we should generalize `regrid_bathymetry` so that it can build `bottom_height` associated with metadata. What do you think? This makes sense esp for GLORYS because the depth field for GLORYS can be explicitly downloaded. I also think its a nicer way to think about it.

navidcy:
> @navidcy I propose deleting `dataset_immersed_grid`. Instead, we should generalize `regrid_bathymetry` so that it can build `bottom_height` associated with metadata. What do you think? This makes sense esp for GLORYS because the depth field for GLORYS can be explicitly downloaded. I also think its more modular. For example you can build bottom height for any resolution.

Do it!

simone-silvestri:
Sounds good.

test/test_downloading.jl
```diff
@@ -15,10 +15,10 @@ end
...
         @info "Testing that we can download $(typeof(dataset)) data..."
 
-        variables = dataset isa ECCO4Monthly ? keys(ClimaOcean.ECCO.ECCO4_short_names) :
-                    dataset isa ECCO2Monthly ? keys(ClimaOcean.ECCO.ECCO2_short_names) :
-                    dataset isa ECCO2Daily ? keys(ClimaOcean.ECCO.ECCO2_short_names) :
-                    dataset isa EN4Monthly ? keys(ClimaOcean.EN4.EN4_short_names) :
+        variables = dataset isa ECCO4Monthly ? keys(ClimaOcean.ECCO.ECCO4_dataset_variable_names) :
+                    dataset isa ECCO2Monthly ? keys(ClimaOcean.ECCO.ECCO2_dataset_variable_names) :
+                    dataset isa ECCO2Daily ? keys(ClimaOcean.ECCO.ECCO2_dataset_variable_names) :
+                    dataset isa EN4Monthly ? keys(ClimaOcean.EN4.EN4_dataset_variable_names) :
```
navidcy, line 21: add the Copernicus test?

src/DataWrangling/metadata.jl
```diff
@@ -208,49 +220,47 @@ that `metadata` corresponds to.
...
                                  topology = (TX, TY, TZ))
+
     return Field{loc...}(grid)
 end
 
 struct Celsius end
 struct Kelvin end
 
-function dataset_temperature_units end
+dataset_temperature_units(metadata) = Celsius()
```
navidcy, line 266: In #467 I deliberately didn't wanna have a fallback

Why would the default is Celsius? I mean, it does feel natural to me... But my thinking was that whoever adds a dataset needed to ensure that they added a method and they particularly checked the units and added the method accordingly.

navidcy:
@glwagner just letting you know that sometimes datasets are not even on the grid their documentation suggests. 
E.g. with @taimoorsohail we struggled a bit to discover the the EN4 dataset was actually on a grid from 0.5 to 360.5 or something. That's why you see this 

https://github.com/CliMA/ClimaOcean.jl/blob/ca9ab721307af3a99f4d366d7bb08f8511c26195/src/DataWrangling/EN4/EN4.jl#L84-L86

@taimoorsohail thought of a good way to test this and that's how we figured out what the mask of the EN4 grid actually corresponded to. It involved putting an artificially high value for the mask and then adding the field on its native grid and shifting until we saw that the high values of the mask didn't appear. Am I right?

src/DataWrangling/metadata.jl
```diff
@@ -67,17 +67,19 @@ function Metadatum(variable_name;
...
 
-    date isa Union{CFTime.AbstractCFDateTime, Dates.AbstractDateTime} ||
-        throw(ArgumentError("date must be Union{Dates.AbstractDateTime, CFTime.AbstractCFDateTime}"))
-
     return Metadata(variable_name, dataset, date, dir)
 end
 
 # Just the current directory
-default_download_directory(dataset) = pwd()
+function default_download_directory end
```
navidcy, line 74: if this is not `pwd()` then delete the comment above

src/DataWrangling/EN4/EN4_metadata.jl
```diff
@@ -38,43 +44,32 @@ end
...
     return "EN.4.2.2.f.analysis.g10." * yearstr * lpad(string(monthstr), 2, '0') * ".nc"
 end
 
 # Convenience functions
-short_name(data::Metadata{<:EN4Monthly}) = EN4_short_names[data.name]
-
+dataset_variable_name(data::EN4Metadata) = EN4_dataset_variable_names[data.name]
 location(data::EN4Metadata) = EN4_location[data.name]
+dataset_temperature_units(data::EN4Metadata) = Kelvin()
+absolute_latitude_bounds(data::Metadata{<:EN4Monthly, <:Any}) = (-83.5, 89.5)
```
navidcy, line 66: A thought: I think what we really want is a way to reproduce the native grid of each dataset... that's where `absolute_latitude_bounds ` is used for, right? perhaps we implement a `native_grid(dataset)` method and call that within, e.g., 

https://github.com/CliMA/ClimaOcean.jl/blob/6ad11ed62e8f80baa3a67b8b4279b401ac2bb1ce/src/DataWrangling/metadata.jl#L233

src/DataWrangling/DataWrangling.jl
```diff
@@ -168,9 +168,13 @@ include("restoring.jl")
...
 include("ECCO/ECCO.jl")
 include("EN4/EN4.jl")
+include("Copernicus/Copernicus.jl")
 
 using .ECCO
 using .EN4
 using .JRA55
+using .Copernicus
+
+export Copernicus
```
navidcy, line 178: the other dataset modules are not exported; perhaps either export all of them or none?

src/DataWrangling/ECCO/ECCO.jl
```diff
@@ -23,22 +23,24 @@ using Dates
...
 function __init__()
     global download_ECCO_cache = @get_scratch!("ECCO")
 end
 
 include("ECCO_metadata.jl")
-include("ECCO_mask.jl")
 
-const SomeECCODataset = Union{ECCO2Monthly, ECCO4Monthly, ECCO2Daily}
+default_mask_value(::Metadata{<:ECCO4Monthly}) = 0
+reversed_vertical_axis(::Metadata{<:SomeECCODataset}) = true
```
navidcy, line 41: seems you deleted SomeECCODataset so how will reversed_vertical_axis work?

src/DataWrangling/metadata.jl
glwagner: ah yes, i removed this because the default should definitely not be current working

src/DataWrangling/EN4/EN4_metadata.jl
glwagner: yes, I think that would be better! we actually need to go in that direction, because for GLORYS we do not always want to assume that the longitude bounds are 0, 360 (for example), we need to do limited area domains.

src/DataWrangling/DataWrangling.jl
glwagner: ahah this was a chatgpt thing. I'll remove it

src/DataWrangling/metadata.jl
glwagner: Do we have any examples of Kelvin temperature ocean datasets? Defaults just help accelerate the process of implementing new datasets. Faster development = better for all

glwagner: Also its very obvious that the units are wrong so the cost of this default is low I think. Defaults have a higher cost when they are subtle

test/test_downloading.jl
glwagner: I will do that once we have solid limited area support. Also it may be good to split it up because copernicus is a bit slow

src/DataWrangling/metadata.jl
navidcy: Fair!

src/DataWrangling/DataWrangling.jl
navidcy: Hahaha 😝 
chatGPT should have known better!

src/DataWrangling/EN4/EN4_metadata.jl
navidcy: Perhaps a later PR. Either or.

src/DataWrangling/metadata.jl
navidcy: Perhaps we don’t need this method at all. People who add datasets just ensure that the temp is in Celsius?

src/DataWrangling/metadata.jl
```diff
@@ -2,10 +2,25 @@ using CFTime
...
+    latitude :: Y 
+    z :: Z
+end
+
+"""
+    BoundingBox(; longitude=nothing, latitude=nothing, z=nothing)
+
+Create a bounding box with `latitude`, `longitude`, and `z` bounds on the sphere.
+"""
+BoundingBox(; longitude=nothing, latitude=nothing, z=nothing) =
```
simone-silvestri, line 16: I think this is a great way to unify the JRA55 field time series with the already existing general field time series method we have for metadata, where the only difference with the other datasets is that with JRA55, we also provide a latitude and longitude bound (we can unify everything in another PR).

glwagner: We need bounding box pretty generally. Also for ERA5 atmosphere

glwagner: I was debating making it `LatitudeLongitudeBoundingBox`, but I think that almost all datasets will be lat-lon (I don't see a use for Cartesian geometry bounding boxes... yet)

glwagner:
> This meant that the inpainting algorithm didn't recognise those grid cells as 'missing' and needing inpainting, enabling a visualisation of how 'off' the native EN4 grid was from the model grid and enabling manual adjustments to the EN4 data.

Ok, so you are saying that using 1e10 was a debugging method that was not intended to be a permanent part of the code? I don't understand the benefit of 1e10 vs NaN for that though. Here I am just interpreting "missing" as a `NaN` value (with extra catches for extreme values + special mask values for particular datasets). Then we inpaint the NaNs.

glwagner:
> The debugging was changing the 1e10 to some other value so that ClimaOcean didn't recocnize those cells as "land" and thus didn't inpaint them. Then we plotted the fields and we were able to see whether the land from the native dataset was shifted or not. We didn't merge any debugging thing.

Right, but we do want to inpaint land in general, right? So we want to inpaint 1e10 values? Sorry, I am still confused.

navidcy:
1/ the use of 1e10 value is confusing!!!!

2/ We don't want to inpaint land. At the moment, ClimaOcean treats 1e10 as land. For some confusing reason, see 1/
If we change the value we replace the missing with to something else, e.g., 2e10 then those dataset values are not treated as land and inpaint inpaints them. So that was a "trick" to check whether the land in the dataset was the same as the land by ETOPO.

navidcy:
> 1/ the use of 1e10 value is confusing!!!!

I've seen that this PR changes the 1e10 value with NaNs which make more sense to me. So that confusion is gone.

glwagner:
Ah ok. Another method to figure out if the grid is wrong is to use `mask_immersed_field!`. For example, you can 1) build a field with _no_ inpainting (therefore missing values will be NaN), and then 2) put the data on an immersed boundary grid and mask the immersed field with another value. Then you will see if there is a systematic mismatch between 1) where the masking is done and 2) the NaN values, indicating missing value in your data.

This can all be done within a script, requiring no source code changes.

src/DataWrangling/metadata.jl
glwagner: EN4 is Kelvin! Touche!

glwagner: kudos to them also. Kelvin is the True Way

glwagner:
Ok, so this PR introduces the concept of "native_*_interfaces" for longitude, latitude, z. This generalizes to any LatitudeLongitudeGrid (stretched or not). We won't shift the data to 0, 360 anymore, so we can do any grid.

A future improvement, in my opinion, will be to implement an interface that allows us to infer the grid from the NetCDF file directly. I think this is a better method because it does not require interpretation, and is robust for example to changes in the native grid. I won't go so far in this PR though.

glwagner:
Note that for ETOPO we do build the grid directly from file, so such an interface could be the next step when merging Bathymetry + field Metadata.

navidcy (approved):

navidcy:
@glwagner I think this PR broke the diagnostics tests. They fail now (locally). Just include `test_diagnostics.jl` and you'll see that the computed mixed-layer depth is different. Also it seems that it doesn't change when we set T, S to the second instance from the corresponding time-serieses?

src/DataWrangling/metadata_field.jl
```diff
@@ -0,0 +1,257 @@
...
+
+    grid = LatitudeLongitudeGrid(arch, FT; halo, longitude, latitude, z,
+                                 size = (Nx, Ny, Nz))
+
+    return grid
+end
+
+"""
+    Field(metadata::Metadatum;
+          architecture = CPU(),
```
navidcy, line 42:
```suggestion
    Field(metadata::Metadatum, arch=CPU();
```

src/DataWrangling/metadata_field.jl
```diff
@@ -95,7 +95,7 @@ function Field(metadata::Metadatum, arch=CPU();
         data = ds[dsname][:, :, :, 1]
 
         # Many ocean datasets use a "depth convention" for their vertical axis
-        if reversed_vertical_axis(metadata)
+        if reversed_vertical_axis(metadata.dataset)
```
glwagner, line 98: @navidcy this was the issue

navidcy: something else might be going on because the values don't seem to change for Tt[1] and Tt[2], right?

navidcy:
@glwagner something's fishy because this:

```Julia
using ClimaOcean
using ClimaOcean.ECCO
using Dates

start = DateTime(1993, 1, 1)
stop  = DateTime(1993, 2, 1)
dates = range(start; stop, step=Month(1))

dataset = ECCO4Monthly()
Tmeta = Metadata(:temperature; dataset, dates)
Tt = FieldTimeSeries(Tmeta)

Tt[10, 10, :, 1]
``` 

gives: 

#### on main

```Julia
56-element OffsetArray(::Vector{Float32}, -2:53) with eltype Float32 with indices -2:53:
 0.0
 0.0
 0.06469269
 0.06469269
 1.8760892
 4.086734
 ⋮
 0.00039837073
 0.16970068
 0.015579106
 0.015579106
 0.0
 0.0
```

#### on _this_ PR

```julia
56-element OffsetArray(::Vector{Float32}, -2:53) with eltype Float32 with indices -2:53:
  0.0
  0.0
 -9.1390275f-6
 -9.1390275f-6
 -8.996571f-6
 -8.037224f-6
 -7.888045f-6
 -7.774998f-6
 -7.266368f-6
 -5.8777205f-6
  ⋮
 -3.7124753f-16
 -2.8022524f-9
  8.2148727f-26
  2.7189068f-26
  3.674263f-30
  1.1685582f-26
  1.1685582f-26
  0.0
  0.0
```

src/DataWrangling/DataWrangling.jl
```diff
@@ -173,8 +173,8 @@ that `metadata` corresponds to.
 function z_interfaces end
 function longitude_interfaces end
 function latitude_interfaces end
+function reversed_vertical_axis end
 
-reversed_vertical_axis(metadata) = false
```
navidcy: I applaud not having default assumptions

glwagner:
ok @navidcy so is the problem inpainting then?

glwagner:
Ok so one big change this PR makes is that the native_grid for each field is not shifted / rotated around to be between 0, 360. The native grid is more directly defined as the grid for hte data. As a result ECCO used to be circshifted but as of this PR is not

For example here is the surface temperature: 

Screenshot 2025-05-09 at 5 06 49 PM

I believe on `main` we should see a circshift.

code same as above butr with

```julia
using GLMakie
heatmap(interior(Tt[1], :, :, Tt.grid.Nz), colorrange=(-2, 30))
```

glwagner:
ok but there was a big bug: https://github.com/CliMA/ClimaOcean.jl/pull/489/commits/e270deb149682c281765abbbf1a32148916c34c5

navidcy:
ok, I hadn't seen that!

navidcy:
> ok @navidcy so is the problem inpainting then?

I'm not sure. But I noticed that sometimes I had inpainted files that were inpainted from main but then this PR found them and used them and didn't re-inpaint with it's bug fix so perhaps that's were some of the inconsistencies were originating from. So I thought it might be a good idea to delete inpainted files at the beginning of every test suite? See [a6bac8b](https://github.com/CliMA/ClimaOcean.jl/pull/489/commits/a6bac8bb25c8d1de2fd48d70c8dbfedb0983b58e).

glwagner:
I believe that tests are effectively passing (there is an unrelated problem downloading JRA55, I think).

We should strive to separate the tests out if possible, because it seems its common for tests to fail because servers are down. So we should have a different test for RepeatYearJRA55 than for MultiYearJRA55, and also different for ECCO vs EN4

On deleting inpainted files --- that's a good idea, we should add an option to `Field` that will force delete the cached file. I think that is a better approach than the more fragile approach of trying to find the inpainted files manually and deleting them, which introduces a dependency between that and how we decide to save the data.

PR #492, opened 2025-04-30
Correct net sea ice fluxes

simone-silvestri:
I am not sure why I removed the concentration here, but not having the concentration in the top fluxes leads to the formation of sea ice in the tropics (radiative transfer leads to a negative balance, which grows sea ice). 

Probably, the only flux that can lead to the formation of new sea ice from a ``no-sea-ice'' condition should be the frazil heat flux. which means that the only way the atmosphere can lead to sea ice formation is through cooling the ocean that freezes

We should probably add the sea ice to the climatology experiment when reworking it as an example

### DIFF
--- a/src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
+++ b/src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
@@ -183,7 +183,8 @@ function compute_net_sea_ice_fluxes!(coupled_model)
     kernel_parameters = interface_kernel_parameters(grid)
 
     sea_ice_surface_temperature = coupled_model.interfaces.atmosphere_sea_ice_interface.temperature
-
+    ice_concentration = sea_ice_concentration(sea_ice)
+    
     launch!(arch, grid, kernel_parameters,
             _assemble_net_sea_ice_fluxes!,
             top_fluxes,
@@ -193,6 +194,7 @@ function compute_net_sea_ice_fluxes!(coupled_model)
             atmosphere_sea_ice_fluxes,
             sea_ice_ocean_fluxes,
             freshwater_flux,
+            ice_concentration,
             sea_ice_surface_temperature,
             downwelling_radiation,
             sea_ice_properties,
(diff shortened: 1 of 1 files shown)


navidcy (approved):

src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
```diff
@@ -239,7 +243,7 @@ end
     Qu = upwelling_radiation(Ts, σ, ϵ)
     Qd = net_downwelling_radiation(i, j, grid, time, α, ϵ, Qs, Qℓ)
 
-    ΣQt = Qd + Qu + Qc + Qv
+    ΣQt = (Qd + Qu + Qc + Qv) * ℵi # If ℵi == 0 there is no heat flux from the top!
```
simone-silvestri, line 246: actually I am not completely sure if this must be multiplied by `ℵi` or it should be
```julia
ΣQt = ifelse(ℵi == 0, zero(grid), Qd + Qu + Qc + Qv)
```
for sure if there is no sea ice `ΣQt` should be zero, but if there is, probably `ℵi` does not play a role because the sea ice thickness is computed per column and not per area.

glwagner: is this the net flux into sea ice?

simone-silvestri: yeah, from the atmosphere

navidcy: yes, in or out of the sea ice from the atmosphere

navidcy: before it was becoming negative at night and because sea-ice was T=0C it was forming sea ice in the equator!

but I also agree it should be ifelse, not multiply by ℵi

simone-silvestri: I think this is the same problem https://github.com/CliMA/ClimaSeaIce.jl/pull/57
Here I was multiplying times the concentration. I will try the `ifelse` in this PR

glwagner: or

```julia
(Qd + Qu + Qc + Qv) * (ℵi == 0)
```

which I think is easier to read.

PR #502, opened 2025-05-07
Add a container for radiative fluxes

simone-silvestri:
which makes it much easier to disentangle fluxes and assess if the radiation is correct

PR #504, opened 2025-05-07
Rework radiative properties

simone-silvestri:
In main we need to have local (spatial and temporal) information to evaluate possible space-time-dependent radiative properties inside the solver. As a consequence, `LatitudeDependentAlbedo` (and `TabulatedAlbedo` would not work in combination with a `SkinTemperature` formulation).

This PR reworks a bit the flow of information (in addition to fixing a bug in `stateindex`) to allow passing radiative properties which are functions.

and also reintroduces a method which was needed but deleted in #502

### DIFF
--- a/src/ClimaOcean.jl
+++ b/src/ClimaOcean.jl
@@ -59,15 +59,15 @@ const SKOFTS = SomeKindOfFieldTimeSeries
     return a(λ, φ, z, time)
 end
 
-@inline function stateindex(a::Tuple, i, j, k, grid, time)
+@inline function stateindex(a::Tuple, i, j, k, grid, time, args...)
     N = length(a)
     ntuple(Val(N)) do n
-        stateindex(a[n], i, j, k, grid, time)
+        stateindex(a[n], i, j, k, grid, time, args...)
     end
 end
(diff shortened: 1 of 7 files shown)


glwagner:
I recall there is some problem with how radiation properties are stored as well. They are stored redundantly as interface properties and also in OceanSeaIceModel. Should we address this higher-level issue first before sorting out the lower-level details?

In particular can you show what the new user interface looks like after this PR?

glwagner:
I'm also wondering whether it makes sense to recompute radiation (eg emission from the ocean surface) if we would have a radiation solver in the code. On the other hand, it does seem that if we want to compute the SkinTemperature including the variation of outgoing emissions with surface temperature, we _need_ to recompute it. Can you lay out the basics of the problem so that we can come up with a future proof design?

simone-silvestri:
This PR is not changing the user interface, but only address some bugs in the internals of the solver. Indeed, the radiation needs to be passed twice to make sure that the component interfaces and the radiation in the model have the same properties. Probably we should change it at the top level

glwagner (approved):

PR #505, opened 2025-05-07
Adding an ECCO prescribed atmosphere

simone-silvestri:
This is just a very coarse atmosphere to test fluxes (or maybe for very coarse ocean simulations that would prefer to be forced by a low frequency atmosphere). Will wait for the copernicus PR.

### DIFF
--- a/src/DataWrangling/ECCO/ECCO.jl
+++ b/src/DataWrangling/ECCO/ECCO.jl
@@ -25,7 +25,7 @@ using Scratch
 
 import ClimaOcean.DataWrangling: vertical_interfaces, variable_is_three_dimensional,
                                  shift_longitude_to_0_360, inpainted_metadata_path,
-                                 longitude_shift
+                                 longitude_shift, maybe_massage_data
 
 download_ECCO_cache::String = ""
 function __init__()
@@ -34,6 +34,7 @@ end
 
 include("ECCO_metadata.jl")
 include("ECCO_mask.jl")
+include("ECCO_atmosphere.jl")
 
 const SomeECCODataset = Union{ECCO2Monthly, ECCO4Monthly, ECCO2Daily}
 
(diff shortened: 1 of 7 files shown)


navidcy:
Conflicts have risen indeed :)

simone-silvestri:
> Conflicts have risen indeed :)

Should be all solved, I think we can merge

src/DataWrangling/metadata_field.jl
```diff
@@ -150,6 +150,8 @@ end
 struct ShiftSouth end
 struct AverageNorthSouth end
 
+reversed_sign(dataset, val_name) = false
+
```
navidcy, line 154: do we need the default? it might create issues?
was thinking that each submodule should define this

simone-silvestri: This follows the design of what we do for temperature units and position in space. 
However, I was also thinking that a cleaner way to implement data mangling to comply to ClimaOcean conventions would be to have it done for submodules (i.e. remove also the general `temperature_units` and push it in the `EN4` module as well as the averaging for the u-velocity).

glwagner: what does "reversed sign" mean? It uses the same concept as "reversed vertical axis"? But I don't think it is refering to the function `reverse`, is it?

src/DataWrangling/metadata_field.jl
```diff
@@ -158,6 +160,8 @@ struct AverageNorthSouth end
 @inline mangle(i, j, k, data, ::ShiftSouth) = @inbounds data[i, j-1, k]
 @inline mangle(i, j, k, data, ::AverageNorthSouth) = @inbounds (data[i, j+1, k] + data[i, j, k]) / 2
 
+@inline maybe_reverse_sign(datum, reverse::Bool) = ifelse(reverse, - datum, datum)
```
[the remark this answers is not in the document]
glwagner: If this has to do with vector conventions, then it might help readability to use a type that specifically refers to the vector convention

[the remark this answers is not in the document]
navidcy: I think it has to do with different conventions about heat flux going up or down that various datasets decide to use...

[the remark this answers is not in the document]
glwagner: yeah so like... `flux_direction` = 1 / -1?

src/DataWrangling/metadata_field.jl
simone-silvestri: right it was just the sign (from - to + and from + to minus). However, I removed it now that the convention for downwelling radiation is positive.

simone-silvestri: Nevermind once again, ECCO's convention is aligned with ours, no need to switch signs... 😅

simone-silvestri:
This PR should be ready to merge.

Issue #508, opened 2025-05-08
Are `CUDA.set_runtime_version!(v"12.6"; local_toolkit = true) ` and `CUDA.precompile_runtime()` required in the CI?


navidcy:
Are these required? 

https://github.com/CliMA/ClimaOcean.jl/blob/9dc402c4897d549d9a468737802b9af8c1564b73/test/runtests.jl#L13-L14

It takes 40min on the CI!

cc @simone-silvestri, @Sbozzolo

Sbozzolo:
If you don't precompile it, it will be precompiled as part of everything else the first time you use CUDA. In this particular case, I don't think it would be helpful to keep it because everything is in the same julia session anyway.

PR #509, opened 2025-05-08
Better comment in CI + remove CUDA precompile runtime

navidcy:
related to #508


glwagner (approved):

navidcy:
Actually even without these two lines, the initialize part of the CI still takes ~50min!

simone-silvestri:
Yeah, these two lines might be superfluous, but they do not affect runtime. The runtime issue is that we are downloading all the data we need for testing in the initialization. I propose to store pre-dowloaded files on the testing location (the Caltech cluster or Tartarus when we move this to Tartarus) and test the downloading on github action.

glwagner:
@Sbozzolo the issue we are facing is that the `initialize` step (which blocks all subsequent steps) takes 47 minutes. @navidcy hypothesized that the issue was `precompile_runtime()` but removing it (in this PR) doesn't seem to help.

glwagner:
> Yeah, these two lines might be superfluous, but they do not affect runtime. The runtime issue is that we are downloading all the data we need for testing in the initialization. I propose to store pre-dowloaded files on the testing location (the Caltech cluster or Tartarus when we move this to Tartarus) and test the downloading on github action.

This makes sense to me.

navidcy:
But if I look at the buildkite log the 40min seems to be pre compilation or something. Is it downloading and I am not seeing it?

glwagner:
@navidcy not sure where you are seeing the timings. That said it might be nice to add timings to `download_dataset`

navidcy:
> @navidcy not sure where you are seeing the timings. 

Screenshot 2025-05-09 at 5 50 41 pm

simone-silvestri:
that is the downloading, it is part of the initialization procedure, the comment is wrong.

navidcy:
> that is the downloading, it is part of the initialization procedure, the comment is wrong.

Ok! 👍🏼

simone-silvestri:
by the comment is wrong I mean this one:
https://github.com/CliMA/ClimaOcean.jl/blob/0d6a2eac808d12d65032179d3863cd479607fe61/.buildkite/pipeline.yml#L27
Should be something more like:
 - "echo '--- Initialize the ClimaOcean test suite'"

PR #514, opened 2025-05-12
Fix a bug in `Field(metadata)`

simone-silvestri:
the `Field` constructor needs the donwloaded data for particular datasets (like Copernicus) to be able to build the `native_grid`. So the `download_dataset` should be performed before `native_grid`

### DIFF
--- a/src/DataWrangling/metadata_field.jl
+++ b/src/DataWrangling/metadata_field.jl
@@ -57,6 +57,11 @@ function Field(metadata::Metadatum, arch=CPU();
                halo = (3, 3, 3),
                cache_inpainted_data = true)
 
+    # Make data available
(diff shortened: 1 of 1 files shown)


src/DataWrangling/metadata_field.jl
```diff
@@ -57,6 +57,9 @@ function Field(metadata::Metadatum, arch=CPU();
                halo = (3, 3, 3),
                cache_inpainted_data = true)
 
+    # Ensure data is available
```
glwagner, line 60: the comment is a little confusing. It's not about "ensuring" anything -- we just need the data

navidcy: #515

Issue #491, opened 2025-04-30
Implementing `Metadata` for different bathymetry products


taimoorsohail:
I am trying to use a different input bathymetry file for running a ClimaOcean simulation, but I noticed the `Bathymetry.jl` file is very much coded for the ETOPO data. For example, the code assumes the `Dataset` has `lat` and `lon` dimensions, and is on a `LatLonGrid` with longitude bounds between `-180` and `180`. 

I think we should implement a bathymetry metadata, as we did with JRA/EN4/ECCO fields, so that we simply specify the bathymetry metadata, and write separate metadata files which handle the native grid, longitude extent, etc. for each bathymetry product. 

I'm just opening up this issue to see if this is a good idea - perhaps there's an easier way to implement this...

simone-silvestri:
That sounds good. The objective is eventually to use `Metadata` for all dataset products we include in the codebase. 
This also justifies moving the `Bathymetry.jl` module inside the `DataWrangling` module

glwagner:
I would hold off --- I am planning to refactor this once #489 goes through.

Note that the point of `regrid_bathymetry` is really just for automatically building bathymetry. If you already have a file available, there is little point in using `regrid_bathymetry`. You should simply directly interpolate the data and then put it into your `ImmersedBoundaryGrid`:

```julia
bottom_height = Field{Center, Center, Nothing}(underlying_grid)
interpolate!(bottom_height, my_bathymetry_data)
grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom_height))
```

As for code design, my plan is to reuse the existing Metadata for bathymetry --- ie supporting "dateless" Metadata. I don't think we should re-implement a new type for a bathymetry metadata. But I am open to discussing this.

glwagner:
More broadly, `Metadata` is only needed for:

1) automatically downloading data when porting to new machines (this is important for the package in general, but not for individual users who have already downloaded data)
2) setting up restoring

Otherwise with manually downloaded data there is no motivation for using Metadata.

taimoorsohail:
Great! I'll hold off till https://github.com/CliMA/ClimaOcean.jl/pull/489 is done :)

glwagner:
@taimoorsohail maybe also we can work on a Metadata tutorial after this. It'd be nice to showcase all of the different data products that can be downloaded automatically and incorporated into simulations, I think it is a very useful feature of ClimaOcean.

[...]

PR #518, opened 2025-05-13
Add Bathymetry metadata to allow using diverse bathymetry datasets

taimoorsohail:
This ~is a work in progress -~ PR is introducing a Bathymetry metadata to download and apply diverse bathymetry datasets. At the moment I am just implementing ETOPO, may do ACCESS in the future.

~The code currently isn't working but I thought I would push it so I can get help from others!~ (@navidcy edit)

Resolves https://github.com/CliMA/ClimaOcean.jl/issues/491


taimoorsohail:
I have added a new Bathymetry folder (equivalent to DataWrangling) - not sure if this is the best approach!

src/DataWrangling/ETOPO/ETOPO.jl
```diff
@@ -0,0 +1,81 @@
...
+function __init__()
+    global download_ETOPO_cache = @get_scratch!("ETOPO")
+end
+
+ETOPO_bathymetry_variable_names = Dict(
+    :temperature => "temperature",
+    :salinity    => "salinity"
+)
+
+struct ETOPOBathy end
```
navidcy, line 27: perhaps `ETOPOBathymetry` is a better name? Bathy means nothing to me.
Or perhaps `ETOPO2022`? That's the dataset we are downloading, right?

src/DataWrangling/Bathymetry.jl
```diff
@@ -433,40 +404,6 @@ function remove_minor_basins!(zb, keep_major_basins)
...
-    else
-        bottom_height = regrid_bathymetry(grid; kw...)
-        jldsave(filename, bathymetry = Array(interior(bottom_height)))
-    end
-
-    return bottom_height
-end
-
-retrieve_bathymetry(grid, ::Nothing; kw...) = regrid_bathymetry(grid; kw...)
-retrieve_bathymetry(grid; kw...)            = regrid_bathymetry(grid; kw...)
```
navidcy, line 470: nice, time to let go of `retrieve_bathymetry`

src/DataWrangling/ETOPO/ETOPO.jl
glwagner: For sure, just the name for the dataset. For other cases we don't add qualifiers like "bathymetry" or "ocean state" so I suggest just the name of the dataset and no more

src/DataWrangling/Bathymetry.jl
```diff
@@ -1,62 +1,33 @@
 module Bathymetry
```
glwagner, line 1: the name should be `module ETOPO` right?

navidcy: There are some bathymetry-related methods that don't necessarily belong in the ETOPO module, e.g., `regrid_bathymetry!`. These could be part of the DataWrangling module, e.g., gathered together in a file called `bathymetry_utils.jl` or something; they don't need their own module. And then we have a separate module for every dataset, e.g., ETOPO, GEBCO, etc.

glwagner: I think we want to have `Bathymetry` module for things related to manipulating bathymetry and then `ETOPO`, etc, for the data downloading

navidcy: That also works

src/Bathymetry.jl
```diff
@@ -132,27 +106,15 @@ function regrid_bathymetry(target_grid;
         return throw(ArgumentError("interpolation_passes has to be an integer ≥ 1"))
     end
 
-    filepath = download_bathymetry(; url, dir, filename)
+    filepath = metadata_path(m)
     dataset = Dataset(filepath, "r")
 
+    grid = native_grid(m, architecture(target_grid); halo = (10, 10, 1))
```
taimoorsohail, line 112: This file can't seem to 'see' the `native_grid` methods I have imported, I have no idea why!

src/Bathymetry.jl
```diff
@@ -161,30 +123,12 @@ function regrid_bathymetry(target_grid;
...
-                                        latitude  = (φ₁_data, φ₂_data),
-                                        longitude = (λ₁_data, λ₂_data),
-                                        z = (0, 1),
-                                        halo = (10, 10, 1))
-
     native_z = Field{Center, Center, Nothing}(native_grid)
     set!(native_z, z_data)
     fill_halo_regions!(native_z)
 
+    ### We still need to fix the grids so they are the same - how did Greg do this?
```
taimoorsohail, line 130: @glwagner I have defined a `native_grid` and `target_grid` - I just need to get them to align without using `circshift` or similar. I couldn't figure out how you did this for the EN4/ECCO datasets?

simone-silvestri: We do not need to align grids if they are in the same longitude range. Oceananigans will interpolate the data taking into account the circularity of the longitude dimension. 
For example, for Oceananigans, a longitude range of (-180, 180) is equivalent to a range (180, 360) concatenated to (0, 180). Another example: a range (700, 720) is equivalent to (340, 360) and so on...

src/DataWrangling/ETOPO/ETOPO.jl
```diff
@@ -0,0 +1,84 @@
...
+    :salinity    => "salinity"
+)
+
+struct ETOPOBathymetry end
+
+default_download_directory(::ETOPOBathymetry) = download_ETOPO_cache
+reversed_vertical_axis(::ETOPOBathymetry) = true
+longitude_interfaces(::ETOPOBathymetry) = (-180,180)
+latitude_interfaces(::ETOPOBathymetry) = (-90,90)
+is_three_dimensional(::ETOPOBathymetry) = false
```
taimoorsohail, line 43: For some reason, Julia doesn't 'see' this method or the two above it... I think I am importing and exporting correctly, especially because `all_dates` seems to work below

taimoorsohail: So it shows 

```Julia
julia> is_three_dimensional(ETOPOmetadata)
true
```

when it should be `false`

navidcy: the way you defined it here, `is_three_dimensional` expects a dataset as argument, not a metadata.

src/DataWrangling/ETOPO/ETOPO.jl
```diff
@@ -0,0 +1,82 @@
...
+    :temperature => "temperature",
+    :salinity    => "salinity"
+)
+
+struct ETOPOBathymetry end
+
+default_download_directory(::ETOPOBathymetry) = download_ETOPO_cache
+reversed_vertical_axis(::ETOPOBathymetry) = true
+longitude_interfaces(::ETOPOBathymetry) = (-180,180)
+latitude_interfaces(::ETOPOBathymetry) = (-90,90)
```
navidcy, line 42:
```suggestion
longitude_interfaces(::ETOPOBathymetry) = (-180, 180)
latitude_interfaces(::ETOPOBathymetry) = (-90, 90)
```

navidcy: ocd :)

src/DataWrangling/metadata_field.jl
```diff
@@ -17,11 +17,16 @@ end
 
 function native_grid(metadata::Metadata, arch=CPU(); halo = (3, 3, 3))
     Nx, Ny, Nz = size(metadata)
+    if is_three_dimensional(metadata)
+        z = z_interfaces(metadata)
+    else
+        z = (0,1)
```
navidcy, line 23:
```suggestion
        z = (0, 1)
```

but why this choice? seems ad hoc?

src/DataWrangling/ETOPO/ETOPO.jl
glwagner: where is this method used? is it used to build the grid?

src/Bathymetry.jl
glwagner: We don't use circshift anymore because it is not a generic solution. I don't think you need a custom native_grid implementation for this.

taimoorsohail:
OK - PR is ready for review :) Thanks all, everything should work now!

src/Bathymetry.jl
taimoorsohail: Yes, I didn't use a custom `native_grid` implementation now

src/DataWrangling/ETOPO/ETOPO.jl
taimoorsohail: Fixed this now - I no longer use `is_three_dimensional`

src/DataWrangling/metadata_field.jl
taimoorsohail: I think in the end the bottom height is set onto a `Field{Center, Center, Nothing}` so it's completely arbitrary

simone-silvestri:
I am not sure why, but GitHub didn't let me add comments but only review comments, which are quite inconvenient.

navidcy:
I can add comments!

src/Bathymetry.jl
```diff
@@ -115,12 +90,9 @@ Keyword Arguments
                   the smallest basins are removed first. `major_basins = 1` retains only the largest basin.
                   If `Inf` then no basins are removed. Default: 1.
 """
-function regrid_bathymetry(target_grid;
+function regrid_bathymetry(target_grid, m::Metadatum;
```
simone-silvestri, line 93: It is quite inconvenient to have to build a `Metadatum` object to download the bathymetry when (generally speaking) the only information you need is the dataset. Let's add a method like:
```julia
function regrid_bathymetry(target_grid; dataset = ETOPOBathymetry(), kw...)
    metadatum = # Build metadata object...
    return regrid_bathymetry(target_grid, metadatum; kw...)
end
```

simone-silvestri, line 93:
```suggestion
function regrid_bathymetry(target_grid, metadatum::Metadatum;
```
a little more verbose

src/Bathymetry.jl
```diff
@@ -132,27 +104,17 @@ function regrid_bathymetry(target_grid;
         return throw(ArgumentError("interpolation_passes has to be an integer ≥ 1"))
     end
 
-    filepath = download_bathymetry(; url, dir, filename)
-    dataset = Dataset(filepath, "r")
+    arch = architecture(target_grid)
 
+    grid = native_grid(m, arch; halo = (10, 10, 1))
```
simone-silvestri, line 109:
```suggestion
    grid = native_grid(metadatum, arch; halo = (10, 10, 1))
```

src/Bathymetry.jl
```diff
@@ -132,27 +104,17 @@ function regrid_bathymetry(target_grid;
...
     FT = eltype(target_grid)
 
-    φ_data = dataset["lat"][:]
-    λ_data = dataset["lon"][:]
+    filepath = metadata_path(m)
```
simone-silvestri, line 112:
```suggestion
    filepath = metadata_path(metadatum)
```

src/Bathymetry.jl
```diff
@@ -161,30 +123,12 @@ function regrid_bathymetry(target_grid;
...
-                                        longitude = (λ₁_data, λ₂_data),
-                                        z = (0, 1),
-                                        halo = (10, 10, 1))
-
-    native_z = Field{Center, Center, Nothing}(native_grid)
+    native_z = Field{Center, Center, Nothing}(grid)
     set!(native_z, z_data)
     fill_halo_regions!(native_z)
 
+    ### We still need to fix the grids so they are the same - how did Greg do this?
```
simone-silvestri, line 130:
```suggestion
```

src/DataWrangling/ETOPO/ETOPO.jl
```diff
@@ -0,0 +1,84 @@
...
+
+download_ETOPO_cache::String = ""
+function __init__()
+    global download_ETOPO_cache = @get_scratch!("ETOPO")
+end
+
+ETOPO_bathymetry_variable_names = Dict(
+    :temperature => "temperature",
+    :salinity    => "salinity"
+)
```
simone-silvestri, line 36:
```suggestion
```
I guess this is vestigial code remaining from copying other datasets

src/DataWrangling/ETOPO/ETOPO.jl
```diff
@@ -0,0 +1,84 @@
...
+longitude_interfaces(::ETOPOBathymetry) = (-180, 180)
+latitude_interfaces(::ETOPOBathymetry) = (-90, 90)
+Base.size(dataset::ETOPOBathymetry, variable) = (21600, 10800, 1)
+Base.size(dataset::ETOPOBathymetry) = (21600, 10800, 1)
+
+all_dates(dataset::ETOPOBathymetry) = nothing
+all_dates(dataset::ETOPOBathymetry, variable) = nothing
+function first_date(dataset::ETOPOBathymetry, variable)
+    return nothing
+end
```
simone-silvestri, line 51:
```suggestion
first_date(dataset::ETOPOBathymetry, variable) = nothing
```

src/DataWrangling/ETOPO/ETOPO.jl
```diff
@@ -0,0 +1,84 @@
...
+last_date(dataset::ETOPOBathymetry, variable) = nothing
+
+const ETOPOMetadatum = Metadatum{<:ETOPOBathymetry, <:Any, <:Any}
+
+const ETOPO_url  = "https://www.dropbox.com/scl/fi/6pwalcuuzgtpanysn4h6f/" *
+"ETOPO_2022_v1_60s_N90W180_surface.nc?rlkey=2t7890ruyk4nd5t5eov5768lt&st=yfxsy1lu&dl=0"
+
+z_interfaces(::ETOPOMetadatum)= [0,1]
```
simone-silvestri, line 59:
```suggestion
z_interfaces(::ETOPOMetadatum)= (0, 1)
```

src/DataWrangling/ETOPO/ETOPO.jl
```diff
@@ -0,0 +1,84 @@
...
+
+function download_dataset(metadatum::ETOPOMetadatum)
+    dir = metadatum.dir
+
+    # Create a temporary directory to store the .netrc file
+    # The directory will be deleted after the download is complete
+    fileurl  = metadata_url(metadatum)
+    filepath = metadata_path(metadatum)
+
+    if !isfile(filepath)
```
simone-silvestri, line 77:
```suggestion
    @root if !isfile(filepath)
```
to download only on the root process

src/DataWrangling/metadata.jl
```diff
@@ -142,7 +142,7 @@ function Base.show(io::IO, metadata::Metadata)
 end
 
 # Treat Metadata as an array to allow iteration over the dates.
-Base.length(metadata::Metadata) = length(metadata.dates)
+Base.length(metadata::Metadata) = isnothing(metadata.dates) ? 1 : length(metadata.dates)
```
simone-silvestri, line 145:
```suggestion
Base.length(metadata::Metadata) = length(metadata.dates)
```
`Base.length` is defined below for `Metadatum` which should cover the case when `dates == nothing`

src/DataWrangling/ETOPO/ETOPO.jl
```diff
@@ -0,0 +1,84 @@
...
+Base.size(dataset::ETOPOBathymetry) = (21600, 10800, 1)
+
+all_dates(dataset::ETOPOBathymetry) = nothing
+all_dates(dataset::ETOPOBathymetry, variable) = nothing
+function first_date(dataset::ETOPOBathymetry, variable)
+    return nothing
+end
+last_date(dataset::ETOPOBathymetry, variable) = nothing
+
+const ETOPOMetadatum = Metadatum{<:ETOPOBathymetry, <:Any, <:Any}
```
simone-silvestri, line 54:
```suggestion
const ETOPOMetadatum = Metadatum{<:ETOPOBathymetry}
```
`Metadatum` has only one type parameter (see the definition in `metadata.jl`)

src/DataWrangling/metadata.jl
```diff
@@ -157,8 +157,12 @@ Base.length(metadata::Metadatum) = 1
...
 
 @inline function Base.iterate(m::Metadata, i=1)
-    if (i % UInt) - 1 < length(m)
-        return Metadata(m.name, m.dataset, m.dates[i], m.bounding_box, m.dir), i + 1
+    if i - 1 < length(m)
+        if m.dates == nothing
+            return m, i + 1
+        else
+            return Metadata(m.name, m.dataset, m.dates[i], m.bounding_box, m.dir), i + 1
+        end
```
simone-silvestri, line 165:
```suggestion
   if (i % UInt) - 1 < length(m)
        return Metadata(m.name, m.dataset, m.dates[i], m.bounding_box, m.dir), i + 1
```
same here. `Base.iterate` is defined below for `Metadatum`

simone-silvestri:
Ok I managed to convert everything to comments

test/test_ETOPO.jl
```diff
@@ -0,0 +1,46 @@
+using ClimaOcean.DataWrangling: download_dataset
+using ClimaOcean.DataWrangling.ETOPO
+using ClimaOcean
+using NCDatasets
+using Oceananigans
+
+
+data_path = expanduser("/Users/tsohail/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/uom/ocean-ensembles/data/")
```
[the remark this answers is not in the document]
navidcy: I dealt with that ;)

src/DataWrangling/ETOPO/ETOPO.jl
navidcy: Yeah, we probably don't need it. At the moment it's not used.

src/Bathymetry.jl
navidcy: Actually, this is good point. Also, if we add a method like this then this PR won't be introducing breaking API changes since all examples would continue to work without changes. Then we can change the bump minor release to just bump a patch release.

src/DataWrangling/metadata_field.jl
navidcy: Wondering whether we should use a Flat z dimension for the bathymetry's native grid. Would that work with interpolating onto a grid with finite Nz? @simone-silvestri?

src/Bathymetry.jl
navidcy: Also, btw, we need to update the docstring of this method.

src/DataWrangling/ETOPO/ETOPO.jl
```diff
@@ -0,0 +1,76 @@
...
+
+z_interfaces(::ETOPOMetadatum)= (0, 1)
+metadata_url(::ETOPOMetadatum) = ETOPO_url
+metadata_filename(metadatum::ETOPOMetadatum) = "ETOPO_2022_v1_60s_N90W180_surface.nc"
+
+function download_dataset(metadatum::ETOPOMetadatum)
+    dir = metadatum.dir
+
+    # Create a temporary directory to store the .netrc file
+    # The directory will be deleted after the download is complete
```
navidcy, line 65:
```suggestion
```

src/DataWrangling/ETOPO/ETOPO.jl
simone-silvestri: well I think it is also wrong, the dataset only has a bottom height

src/Bathymetry.jl
```diff
@@ -310,7 +255,7 @@ function regrid_bathymetry(target_grid::DistributedGrid; kw...)
     # shared, we could easily have OOM errors.
     # We perform the reconstruction only on rank 0 and share the result.
     bottom_height = if arch.local_rank == 0
-        bottom_field = regrid_bathymetry(global_grid; kw...)
+        bottom_field = regrid_bathymetry(global_grid, args...; kw...)
         bottom_field.data[1:Nx, 1:Ny, 1]
     else
         zeros(Nx, Ny)
```
simone-silvestri, line 261: If we keep this if-condition, we need to ensure that there is no `@root` in the `regrid_bathymetry` function; otherwise, we would encounter a deadlock. Probably we need to rethink a bit the `regrid_bathymetry` function to allow a `@root` in the `download_dataset`. This would mean having a child function that regrids the bathymetry assuming that the data is already downloaded and splitting `regrid_bathymetry` into
- `download_dataset` (has a `@root` inside)
- function that actually regrids the downloaded data
in this way, the distributed `regrid_bathymetry` will have the same structure without incurring deadlocks.

glwagner:
> Thanks, I should have mentioned that `test/test_ETOPO.jl` will be removed before merging as I have incorporated tests into `test_bathymetry.jl`

We may want to have a different file for each data product. (We also need to separate ECCO and EN4)

src/Bathymetry.jl
taimoorsohail: I don't understand:

1) `regrid_bathymetry` is no longer used to "download the bathymetry" - that is now in `download_dataset`. 
2) We will still need to redefine `regrid_bathymetry`, just with `dataset` as an added kwarg instead of `metadatum` - so we will still introduce breaking changes for all examples that load a bathymetry. 
3) I am happy to switch `metadata` with `dataset` in the `regrid_bathymetry`, though I am unsure of what the added benefit of this would be - simply that we are ensuring the user doesn't have to predefined their metadata? 
4) In my opinion, letting the user predefined metadatum first is helpful, because then they can at least inspect the metadatum, like the directory, etc.

navidcy: @simone-silvestri was suggesting the above method as an _additional_ convenience method. He's not suggesting we drop the method with the metadadum.

I implemented this in 

https://github.com/CliMA/ClimaOcean.jl/blob/6bb3984272140e907b217acdd89c9adc27d16f47/src/Bathymetry.jl#L143-L147

navidcy: > regrid_bathymetry is no longer used to "download the bathymetry" - that is now in download_dataset.

hm... ok, I might have missed that. I think similarly to other datasets, download should be also included in methods that use the data (eg in `set!`). This way eg regrid_bathymetry will download the data if it's not there. But regrid_bathymetry _won't be the only way_ to download the data.

src/Bathymetry.jl
```diff
@@ -116,14 +86,15 @@ Keyword Arguments
...
                            minimum_depth = 0,
-                           dir = download_bathymetry_cache,
-                           url = etopo_url,
-                           filename = "ETOPO_2022_v1_60s_N90W180_surface.nc",
                            interpolation_passes = 1,
                            major_basins = 1) # Allow an `Inf` number of "lakes"
 
+    if isnothing(metadata)
+        throw(ArgumentError("Metadata must be provided"))
+    end
```
navidcy, line 97: if metadata needs to be provided then no need to give it a default `nothing` value upstairs

src/Bathymetry.jl
```diff
@@ -116,14 +86,15 @@ Keyword Arguments
                   If `Inf` then no basins are removed. Default: 1.
 """
 function regrid_bathymetry(target_grid;
+                           metadata = nothing,
```
navidcy, line 89:
```suggestion
                           metadata,
```

navidcy: but why change to kwarg instead of arg? was this intentional?

src/Bathymetry.jl
taimoorsohail: Not sure I understand. 'regrid_bathymetry' doesn't call 'download_dataset' in this PR. So I don't see where the deadlock would be?

src/DataWrangling/DataWrangling.jl
```diff
@@ -202,12 +203,14 @@ end
 # Datasets
 include("JRA55/JRA55.jl")
 include("ECCO/ECCO.jl")
-include("EN4.jl")
+include("EN4/EN4.jl")
```
navidcy, line 206: how was EN4 working with the wrong file path? 😳

src/DataWrangling/ETOPO/ETOPO.jl
```diff
@@ -0,0 +1,75 @@
...
+)
+
+struct ETOPO2022 end
+
+default_download_directory(::ETOPO2022) = download_ETOPO_cache
+reversed_vertical_axis(::ETOPO2022) = true
+longitude_interfaces(::ETOPO2022) = (-180, 180)
+latitude_interfaces(::ETOPO2022) = (-90, 90)
+Base.size(::ETOPO2022, variable) = (21600, 10800, 1)
+Base.size(::ETOPO2022) = (21600, 10800, 1)
```
navidcy, line 44:
```suggestion
Base.size(::ETOPO2022) = (21600, 10800, 1)
Base.size(dataset::ETOPO2022, variable) = size(dataset)
```

src/Bathymetry.jl
navidcy: if it's a kwarg then we can't disambiguate between methods that need the metadata and the convenience that uses the default dataset and constructs the metadata.

```Julia
function regrid_bathymetry(target_grid, metadata; ...)
    do all the things
    return target_z
end

# convenience
function regrid_bathymetry(target_grid; dataset=ETOPO2022(), kw...)
    metadata = ...
    return regrid_bathymetry(target_grid, metadata; kw...)
end
```

taimoorsohail: Ah! Thanks, didn't realise. Happy to change it back.

src/Bathymetry.jl
simone-silvestri: ah, as navid I also assumed that `regrid_bathymetry` downloads the data if necessary. I think we should keep this feature

src/Bathymetry.jl
simone-silvestri: I also prefer to have it as a positional argument, since we have it everywhere as a positional argument

src/Bathymetry.jl
taimoorsohail: Oh I see! Yes I guess I didn't think about what would happen if the dataset doesn't exist, other than it would throw an error! I agree, although maybe we should rename the function to something more descriptive like 'fetch_and_regrid_bathymetry'? I remember I was confused why the regridding function was also downloading.

navidcy: Other methods download if needed, e.g., `set!`.
I think it's ok to download if needed, no need to include the "download_" bit as part of all methods that potentially do that.

The problematic situation before was that there wasn't any other way to download the bathymetry. But now there is.

So I was thinking that now we could download the bathymetry like this:

```Julia
using ClimaOcean

metadatum = Metadatum(:bottom_height; dataset=ETOPO2022())
download_dataset(metadatum)
```

but also we could just do

```Julia
using ClimaOcean

metadatum = Metadatum(:bottom_height; dataset=ETOPO2022())
regrid_bathymetry(metadatum)
```

and `regrid_bathymetry` would download if needed; similarly to what other methods do.

What do you think?

simone-silvestri: Sounds good. It is convenient to download the dataset inside the `regrid_bathymetry` function. However, what we cannot have is:
```julia
function download_dataset(::ETOPOBathymetry)
    @root begin # Only rank 0 enters here...
         # Download stuff...
    end # All ranks (included 0) wait for _All_ the ranks to reach this point (There is an explicit barrier in @root)
end
```
that would be the preferred implementation of `download_dataset` (because data is downloaded only on the root process), in combination with
```julia
function regrid_bathymetry(grid::DistributedGrid, args...; kwargs...)
    bottom_height = if architecture(grid).local_rank == 0
          regrid_bathymetry(global_grid, args...; kwargs...) # Only rank 0 enters here but since there is an explicit barrier in this step we will incur in a deadlock. I.e. -> not all ranks take this path, but the barrier assumes all of them do
    else
        zeros(Nx, Ny)
    end
end
```

simone-silvestri: So a workaround is to split the downloading of the bathymetry from the regridding:
```julia
function regrid_bathymetry(grid, metadata; kwargs...)
    download_dataset(metadata)
    return actually_regrid_bathymetry(grid, metadata; kwargs...)
end
```
so that the following is legal:
```julia
function regrid_bathymetry(grid::DisatributedGrid, metadata; kwargs...)
    download_dataset(metadata)
    bottom_heigth = if arch.local_rank == 0
           actually_regrid_bathymetry(grid, metadata; kwargs...)
   else 
           zeros()
   end
   # Scatter the bathymetry to the other cores...
   return bottom_heigth
end
```
because we do the downloading outside the divergent path for ranks.
(`actually_regrid_bathymetry` is probably not the right name, feel free to choose the most communicative name)

navidcy:
```Julia
using ClimaOcean
using ClimaOcean.ETOPO
using Oceananigans

arch = CPU()
Nx, Ny, Nz = 360, 180, 2

underlying_grid = TripolarGrid(arch;
                               size = (Nx, Ny, Nz),
                               z = (-6000, 0),
                               halo = (5, 5, 4),
                               first_pole_longitude = 70,
                               north_poles_latitude = 55)

bathymetry_kwargs = (height_above_water = nothing,
                     minimum_depth = 10,
                     interpolation_passes = 1,
                     major_basins = 2)

@info "Bathymetry using default ECCO2022 dataset..."
bottom_height_1 = regrid_bathymetry(underlying_grid; bathymetry_kwargs...)

@info "Bathymetry using Metadata..."
ETOPOmetadata = Metadatum(:bottom_height, dataset=ETOPO2022())

filepath = ClimaOcean.DataWrangling.download_dataset(ETOPOmetadata)

bottom_height = regrid_bathymetry(underlying_grid, ETOPOmetadata; bathymetry_kwargs...)

using GLMakie
fig = Figure()
ax = Axis(fig[1, 1])
hm = heatmap!(ax, interior(bottom_height, :, :, 1), colormap=:deep)
contour!(ax, interior(bottom_height, :, :, 1), levels=[0.], color=:black)
Colorbar(fig[1, 2], hm)
display(fig)
```
display

src/Bathymetry.jl
navidcy: I agree with the above except the "actually_regrid_bathymetry" name :)

simone-silvestri: what about `regrid_bathymetry` for the exterior function that downloads the dataset and `_regrid_bathymetry` for the interior function that assumes that data is downloaded?

navidcy: we think alike! I did that in 650cc1b257f6f63e4e1ee6f6f6fa6f1ac6d06215 before your comment :D

navidcy:
I'm happy with this PR! But a second pair of eyes would be nice.

src/Bathymetry.jl
```diff
@@ -135,6 +148,38 @@ function regrid_bathymetry(target_grid; dataset = ETOPO2022(), kw...)
     return regrid_bathymetry(target_grid, metadatum; kw...)
 end
 
+# Regridding bathymetry for distributed grids, we handle the whole process
+# on just one rank, and share the results with the other processors.
+function regrid_bathymetry(target_grid::DistributedGrid, args...; kw...)
```
simone-silvestri, line 153:
```suggestion
function regrid_bathymetry(target_grid::DistributedGrid, metadatum; kw...)
    download_dataset(metadatum)
```

navidcy: I left it as `args...` so that the convenience method with the default dataset also works.

navidcy: nice catch; thanks!

navidcy:
I think this is ready and should be merged if tests pass. Any more comments?

It's noteworthy that although it generalises something the PR has a net negative of code lines ;)

navidcy:
No sure why this error happens:

https://buildkite.com/clima/climaocean-ci/builds/4068#0196dcf5-14b1-4ece-8038-dde7ba89bacb/187-466

I did run the `test_diagnostics.jl` on tartarus GPU and all was good...

navidcy:
The test_distributed_utils.jl needs to be modified since it's using a method for regrid_bathymetry that doesn't exist anymore:

https://github.com/CliMA/ClimaOcean.jl/blob/4e6c618ac2d9dbad7b0124347e5bb4a3c61fe6fd/test/test_distributed_utils.jl#L60-L63

glwagner:
Can someone show what the new syntax looks like?

navidcy:
I posted an example just upstairs. 
but also the previous syntax works. All examples should work out of the box.

navidcy:
https://github.com/CliMA/ClimaOcean.jl/pull/518#issuecomment-2886113352

glwagner:
Question: is the `/` necessary here? If so, we should make it so that its not (and omit it from code).

```julia
 global_height = regrid_bathymetry(global_grid; 
                                   dir = "./", 
                                   filename = "trivial_bathymetry.nc", 
                                   interpolation_passes=10) 
```

glwagner:
I don't think we should support specifying a `dir` in `regrid_bathymetry` if that's what you mean by previous syntax being supported. I think it'd be better to always go through the `Metadata` interface if info about the data is required.

navidcy:
> Question: is the `/` necessary here? If so, we should make it so that its not (and omit it from code).
> 
> ```julia
>  global_height = regrid_bathymetry(global_grid; 
>                                    dir = "./", 
>                                    filename = "trivial_bathymetry.nc", 
>                                    interpolation_passes=10) 
> ```



> I don't think we should support specifying a `dir` in `regrid_bathymetry` if that's what you mean by previous syntax being supported. I think it'd be better to always go through the `Metadata` interface if info about the data is required.

We don’t support dir. 
the code you commented on is old.


but now I’m thinking that we should have bumped minor release. Will do that later.

glwagner:
yes that'd be great! I would first put proper diffusion smoothing in first as a feature (which is not very hard, the docs stuff will take longer), and then build up docs around a carefully designed interface

PR #520, opened 2025-05-13
Add wind-dependent Charnock coefficient

simone-silvestri:


src/OceanSeaIceModels/InterfaceComputations/roughness_lengths.jl
```diff
@@ -1,7 +1,7 @@
-struct MomentumRoughnessLength{FT, V}
+struct MomentumRoughnessLength{FT, G, V}
     gravitational_acceleration :: FT
     air_kinematic_viscosity :: V
-    gravity_wave_parameter :: FT
+    gravity_waves :: G
```
glwagner, line 4:
```suggestion
    wave_formulation :: G
```

src/OceanSeaIceModels/InterfaceComputations/roughness_lengths.jl
```diff
@@ -12,6 +12,25 @@ struct ScalarRoughnessLength{FT, V, R}
     maximum_roughness_length :: FT
 end
 
+struct WindDependentGravityWaveParameter{FT}
```
glwagner, line 15:
```suggestion
struct WindDependentWaveFormulation{FT}
```

navidcy:
Let's see how the simulations change in the docs preview. Then we can revert back to the default `gravity_wave_parameter = 0.011` before merging.

simone-silvestri:
> Let's see how the simulations change in the docs preview. Then we can revert back to the default `gravity_wave_parameter = 0.011` before merging.

If the results are more physical, we can keep this as a default.

simone-silvestri:
I don't see too much difference in the maximum temperature...

navidcy:
We have a preview:

https://clima.github.io/ClimaOceanDocumentation/previews/PR520/literated/near_global_ocean_simulation/

Not sure if the results look much different. At least not by eyeballing (my eyes at least).

navidcy (approved):

glwagner:
I support merging but I don't think we should make this default

glwagner:
Can we change this PR to just implement the Charnock coefficient rather than making it default? I think we are working on fluxes and we should keep all the defaults relatively simple if possible.

navidcy:
OK!

PR #521, opened 2025-05-14
Correct atmosphere - sea ice heat flux

simone-silvestri:
We decided that this was the correct formulation but haven't changed the src code to reflect it yet. This change has already happened in [ClimaSeaIce](https://github.com/CliMA/ClimaSeaIce.jl/commit/0f2444add91738186f5a2c37d0886e9e7c3b6d1c)


navidcy:
Add a compat entry requirement for ClimaSeaIce?

navidcy (approved):

simone-silvestri:
I think the current compact is fine, that change was just in the examples.

PR #522, opened 2025-05-14
Fix a couple of bugs with `Metadata`

simone-silvestri:


src/DataWrangling/Copernicus/Copernicus.jl
```diff
@@ -20,6 +20,7 @@ import ClimaOcean.DataWrangling:
     latitude_interfaces,
     z_interfaces,
     metadata_filename,
+    metadata_path,
```
navidcy, line 23: `metadata_path` is a method of Oceananigans.Fields?

simone-silvestri: it's a method of `ClimaOcean.DataWrangling`

navidcy: Yes. That was my impression. Then how come we import it here from Oceananigans.Fields? Typo?

simone-silvestri: I think you misread the changes. We are importing `metadata_path` from `ClimaOcean.DataWrangling`. From Oceananigans.Fields we are importing `location`
https://github.com/CliMA/ClimaOcean.jl/blob/64b821b0986435474ff486d2d2efa429d641bdc0/src/DataWrangling/Copernicus/Copernicus.jl#L12-L25

glwagner: we shouldn't need to extend the path. why do we need to extend it here?

navidcy: definitely I misread...

src/DataWrangling/Copernicus/Copernicus.jl
```diff
@@ -58,8 +58,8 @@ copernicusmarine_dataset_id(::GLORYSMonthly) = "cmems_mod_glo_phy_my_0.083deg_P1
 struct CMEMSHourlyAnalysis <: CopernicusDataset end
 copernicusmarine_dataset_id(::CMEMSHourlyAnalysis) = "cmems_mod_glo_phy_anfc_0.083deg_PT1H-m"
 
-CopernicusMetadata{D} = Metadata{<:CopernicusDataset, D}
-CopernicusMetadatum = Metadatum{<:CopernicusDataset}
+const CopernicusMetadata{D} = Metadata{<:CopernicusDataset, D}
```
navidcy, line 61: Metadata has 3 params, right?

simone-silvestri: Metadatum has one. Let's merge this then I can add the additional `B` parameter to `Metadatum`

navidcy: Yes.

navidcy (approved):

navidcy:
if CI passes then let's do it...

src/DataWrangling/Copernicus/Copernicus.jl
navidcy: if CI passes then let's do it...

PR #540, opened 2025-05-23
Print globally averaged surface temperature for one-degree example

glwagner:
This PR adds printing of the globally-averaged surface temperature in the one-degree example. I'd also like to add a few time-series. I'm thinking:

* globally-averaged surface temperature
* mixed layer depth
* globally-averaged fluxes

### DIFF
--- a/examples/one_degree_simulation.jl
+++ b/examples/one_degree_simulation.jl
@@ -44,8 +44,8 @@ grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom_height);
 #
 # We include temperature and salinity surface restoring to ECCO data from 1993 thoughout the water column.
 
-start_date = DateTime(1993, 1, 1)
-stop_date = DateTime(1993, 12, 1)
+start_date = DateTime(1993, 6, 1)
+stop_date = DateTime(1994, 5, 1)
 dates = range(start_date, step=Month(1), stop=stop_date)
 ecco_temperature = Metadata(:temperature; dates, dataset=ECCO4Monthly())
 ecco_salinity = Metadata(:salinity; dates, dataset=ECCO4Monthly())
@@ -117,7 +117,7 @@ function progress(sim)
     u, v, w = ocean.model.velocities
     T = ocean.model.tracers.T
     e = ocean.model.tracers.e
-    Tmin, Tmax = minimum(T), maximum(T)
+    Tmin, Tmax, Tavg = minimum(T), maximum(T), mean(T)
     emax = maximum(e)
     umax = (maximum(abs, u), maximum(abs, v), maximum(abs, w))
 
@@ -125,7 +125,7 @@ function progress(sim)
 
     msg1 = @sprintf("Time: %s, iter: %d", prettytime(sim), iteration(sim))
     msg2 = @sprintf(", max|u|: (%.1e, %.1e, %.1e) m s⁻¹, ", umax...)
-    msg3 = @sprintf(", extrema(T): (%.1f, %.1f) ᵒC, ", Tmin, Tmax)
+    msg3 = @sprintf(", extrema(T): (%.1f, %.1f) ᵒC, mean(T): %.1f", Tmin, Tmax, Tavg)
     msg4 = @sprintf(", maximum(e): %.2f m² s⁻², ", emax)
     msg5 = @sprintf(", wall time: %s \n", prettytime(step_time))
 
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -17,9 +17,9 @@ const EXAMPLES_DIR = joinpath(@__DIR__, "..", "examples")
 const OUTPUT_DIR   = joinpath(@__DIR__, "src/literated")
 
 to_be_literated = [
-    "single_column_os_papa_simulation.jl",
+    # "single_column_os_papa_simulation.jl",
     "one_degree_simulation.jl",
-    "near_global_ocean_simulation.jl"
+    # "near_global_ocean_simulation.jl"
 ]
 
 for file in to_be_literated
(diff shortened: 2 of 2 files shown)


glwagner:
Perhaps a plot of zonally-averaged fluxes and SST would be nice too? To show seasonality?

examples/one_degree_simulation.jl
```diff
@@ -44,8 +44,8 @@ grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom_height);
 #
 # We include temperature and salinity surface restoring to ECCO data from 1993 thoughout the water column.
 
-start_date = DateTime(1993, 1, 1)
-stop_date = DateTime(1993, 12, 1)
+start_date = DateTime(1993, 6, 1)
+stop_date = DateTime(1994, 5, 1)
```
navidcy, line 48: why?

examples/one_degree_simulation.jl
```diff
@@ -117,15 +117,15 @@ function progress(sim)
     u, v, w = ocean.model.velocities
     T = ocean.model.tracers.T
     e = ocean.model.tracers.e
-    Tmin, Tmax = minimum(T), maximum(T)
+    Tmin, Tmax, Tavg = minimum(T), maximum(T), mean(T)
```
navidcy, line 120: does the `mean` ignore land? or do we need to do `compute!(Field(Average(T)))[1, 1, 1]`?

glwagner: Yes, all reductions of `Field` will ignore immersed cells

glwagner: A lot of code has previously used various workarounds, including using `interior`, because conditional reductions often did not compile on the GPU. But since our revamp of conditional operations, I have found that we can use simple clean syntax now.

We may also want something like `mean(view(T, :, :, Nz))` which would compute the surface average

examples/one_degree_simulation.jl
```diff
@@ -117,15 +117,15 @@ function progress(sim)
...
+    Tmin, Tmax, Tavg = minimum(T), maximum(T), mean(view(T, :, :, ocean.model.grid.Nz))
     emax = maximum(e)
     umax = (maximum(abs, u), maximum(abs, v), maximum(abs, w))
 
     step_time = 1e-9 * (time_ns() - wall_time[])
 
     msg1 = @sprintf("Time: %s, iter: %d", prettytime(sim), iteration(sim))
     msg2 = @sprintf(", max|u|: (%.1e, %.1e, %.1e) m s⁻¹, ", umax...)
-    msg3 = @sprintf(", extrema(T): (%.1f, %.1f) ᵒC, ", Tmin, Tmax)
+    msg3 = @sprintf(", extrema(T): (%.1f, %.1f) ᵒC, mean(T): %.1f", Tmin, Tmax, Tavg)
```
glwagner, line 128:
```suggestion
    msg3 = @sprintf(", extrema(T): (%.1f, %.1f) ᵒC, mean(T(z=0)): %.1f", Tmin, Tmax, Tavg)
```

examples/one_degree_simulation.jl
glwagner: because then the solution won't be biased to heat up as much at the beginning (eg starting during northern hemisphere winter)

navidcy:
> docs didn't build, I guess i need another commit

you need to add the label _before_ you create the PR if you want the docs to built from the first commit

navidcy:
I run the 1-degree simulation for 5 years and changed the color range on the SST to go up to 38.

Here's a movie:

https://clima.github.io/ClimaOceanDocumentation/previews/PR540/literated/one_degree_simulation/

Seems like the very big SSTs are just in the Java sea?

cc @taimoorsohail

navidcy:
https://github.com/user-attachments/assets/ad2c6e7f-8b18-4759-b252-48cc463e4ca6

navidcy (approved):

taimoorsohail:
For what it's worth, I ran the code with the `MutableVerticalDiscretization` and it warmed up WAY more - above 60C. 

This may be a secondary issue, or may point us to what is causing this (milder) warm up. I thought I would flag it though.

taimoorsohail:
flux_issue_RYF1deg_mutablegrid
flux_issue_RYF1deg_w_WENO
flux_issue_RYF1deg_wo_mutable

examples/one_degree_simulation.jl
```diff
@@ -44,8 +45,8 @@ grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom_height);
 #
 # We include temperature and salinity surface restoring to ECCO data from 1993 thoughout the water column.
 
-start_date = DateTime(1993, 1, 1)
-stop_date = DateTime(1993, 12, 1)
+start_date = DateTime(1993, 6, 1)
+stop_date = DateTime(1994, 5, 1)
 dates = range(start_date, step=Month(1), stop=stop_date)
 ecco_temperature = Metadata(:temperature; dates, dataset=ECCO4Monthly())
```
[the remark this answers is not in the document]
taimoorsohail: Oops - yep that's right!

navidcy:
> For what it's worth, I ran the code with the `MutableVerticalDiscretization` and it warmed up WAY more - above 60C.
> 
> This may be a secondary issue, or may point us to what is causing this (milder) warm up. I thought I would flag it though.

@simone-silvestri are these you think related to why z-star is behaving differently (or wrong)?

- https://github.com/CliMA/Oceananigans.jl/pull/4546 
- https://github.com/CliMA/Oceananigans.jl/pull/4487
- https://github.com/CliMA/Oceananigans.jl/pull/4546

?

PR #557, opened 2025-06-05
Bypass certificate verification in downloader

nefrathenrici:
The ECCO server is sending an [SSL certificate chain](https://www.ssllabs.com/ssltest/analyze.html?d=ecco.jpl.nasa.gov) that includes Entrust intermediate certificates. Since Entrust root certificates have been distrusted by many platforms, this causes certificate validation failures.

This PR works around the issue by bypassing certificate verification in the `netrc_downloader`. While the data transfer is still encrypted disabling certificate verification means we no longer confirm the server’s authenticity, which does introduce some security risks.

### DIFF
--- a/src/DataWrangling/DataWrangling.jl
+++ b/src/DataWrangling/DataWrangling.jl
@@ -84,8 +84,11 @@ end
 function netrc_downloader(username, password, machine, dir)
     netrc_file = netrc_permission_file(username, password, machine, dir)
     downloader = Downloads.Downloader()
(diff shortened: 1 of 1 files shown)


nefrathenrici:
> I guess this PR is a fix for these tests failing? https://github.com/CliMA/ClimaOcean.jl/actions/runs/15485716733/job/43599831576

Yes, this is the same [error](https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/381#01973dd2-1b34-45f6-8f14-73b62e7214d9/174-307) I ran into using ClimaOcean in the coupler.

PR #558, opened 2025-06-06
Convert type in `InterfaceState`

simone-silvestri:
The flux computation is failing because of type conversion which leads to `u★` having a different type than `u` in `InterfaceStates`. In particular, the error is 
```julia
ERROR: LoadError: MethodError: no method matching InterfaceState(::Float64, ::Float64, ::Float64, ::Float32, ::Float32, ::Float32, ::Float32, ::Float64, ::Bool)

Closest candidates are:
  InterfaceState(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any)
   @ ClimaOcean ~/development/ClimaOcean.jl/src/OceanSeaIceModels/InterfaceComputations/interface_states.jl:374
  InterfaceState(::FT1, ::FT1, ::FT1, ::FT2, ::FT2, ::FT2, ::FT2, ::FT2, ::Bool) where {FT1, FT2}
   @ ClimaOcean ~/development/ClimaOcean.jl/src/OceanSeaIceModels/InterfaceComputations/interface_states.jl:363
```
because in this case, the atmosphere, the ocean, and the interface have different float types. 
I think we are missing the possibility of changing float types in some constructors, so I'll take a look at that later.
In the meantime, this PR ensures the code does not error when the types are different for the different components.

### DIFF
--- a/src/DataWrangling/DataWrangling.jl
+++ b/src/DataWrangling/DataWrangling.jl
@@ -84,8 +84,11 @@ end
 function netrc_downloader(username, password, machine, dir)
     netrc_file = netrc_permission_file(username, password, machine, dir)
     downloader = Downloads.Downloader()
-    easy_hook  = (easy, _) -> Downloads.Curl.setopt(easy, Downloads.Curl.CURLOPT_NETRC_FILE, netrc_file)
-
+    easy_hook  = (easy, _) -> begin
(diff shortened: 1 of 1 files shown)


navidcy:
I'm wondering if this is not the right place to convert and if we merge it will be harder to find out where the convert is missing in a constructor or something?

Could you post a MWE that reproduces this error?

But also I'm happy to be merged if this might be obstructing progress in other fronts.

navidcy (approved):

PR #559, opened 2025-06-08
Include sea ice in the  `one_degree_simulation.jl`

simone-silvestri:
I think, now that the coupling is working we could be ambitious and add the sea ice to the examples. It will help us go forward.


simone-silvestri:
The reason for the crash might be https://github.com/CliMA/ClimaSeaIce.jl/pull/79. I forgot I was using that branch for the omip simulation but I had not opened a PR yet.

simone-silvestri:
Looks like the simulation runs now!
https://clima.github.io/ClimaOceanDocumentation/previews/PR559/literated/one_degree_global_ocean_surface.mp4

I ll fix a bit the visualization.

examples/one_degree_simulation.jl
```diff
@@ -77,17 +60,31 @@ free_surface       = SplitExplicitFreeSurface(grid; substeps=70)
...
-ocean = ocean_simulation(grid; momentum_advection, tracer_advection, forcing, free_surface,
+ocean = ocean_simulation(grid; momentum_advection, tracer_advection, free_surface,
                          closure=(eddy_closure, horizontal_viscosity, vertical_mixing))
 
 @info "We've built an ocean simulation with model:"
 @show ocean.model
 
+# ### Sea Ice simulation
+# We also need to build a sea ice simulation. We use the default configuration:
+# and EVP rheology and a zero-layer thermodynamic model that advances thickness
```
navidcy, line 71:
```suggestion
# EVP rheology and a zero-layer thermodynamic model that advances thickness
```

navidcy:
The simulation blows up tho soon..

simone-silvestri:
I think the problem with the simulation crashing was me removing the `h = 34` in the vertical grid. This leads to very large spacings at the surface (~40m), while probably we want smaller spacings (closer to 5m) (we should definitely change this kwarg because it is not very clear). 
However, smaller vertical spacings lead to higher CFL constraints, which do not allow us to evolve the model with 20mins timestep.

simone-silvestri:
I think we can merge this PR, this is the output https://clima.github.io/ClimaOceanDocumentation/previews/PR559/literated/one_degree_simulation/

glwagner:
You may want to compute (a*u) (eg concentration times speed) rather than speed itself

navidcy:
> You may want to compute (a*u) (eg concentration times speed) rather than speed itself

what about masking with `h` instead with speed itself; see [8953fe2](https://github.com/CliMA/ClimaOcean.jl/pull/559/commits/8953fe2188d50e8a511f93239302e24314570581)

glwagner:
Instead of `h` you may want to use "effective thickness" which is `a * h` --- this represents the average thickness in a cell. The point is that you can have 1e-9 conccentration but 1m thickness; this represents basically 0 ice volume.

navidcy:
So is this label wrong?

https://github.com/CliMA/ClimaOcean.jl/blob/b43daee8b9c8b81b3828c05ad2bb95ea0f29922b/examples/one_degree_simulation.jl#L269

That's not really effective thickness?

What's the `a` variable you are referring to -- how do I access it? `seaice.model.a`? No...

Looking at 

https://github.com/CliMA/ClimaSeaIce.jl/blob/29101ce037fdd77114f646e28a91c890d112c0bc/src/sea_ice_model.jl#L11-L32

I can't figure it out.

navidcy:
look at the seasonal cycle; it's supposed to start from June and go for a year... doesn't it look strange?

https://github.com/user-attachments/assets/7f05d8ae-7696-47e7-ae29-4eedce58b620

How does the atmosphere knows where to start?

navidcy:
Seasonal cycle looks right now!
The sea ice speed might need some tweaking or converted to effective speed. But I'm merging and we can tweak later.

PR #456, opened 2025-04-03
OMIP prototype

simone-silvestri:
requires #437 and #453

PR #561, opened 2025-06-13
Fix bug on sea ice fluxes

navidcy:
From #456

### DIFF
--- a/src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
+++ b/src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
@@ -261,7 +261,7 @@ end
     @inbounds begin
         Ts = surface_temperature[i, j, kᴺ]
         Ts = convert_to_kelvin(sea_ice_properties.temperature_units, Ts)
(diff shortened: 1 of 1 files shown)


simone-silvestri:
Right, I guess for now it is irrelevant because we have a zero-layer model, and concentration is a reduced field in z. I think we haven't decided yet how a multi-layer model will look, but I guess the concentration will always remain two-dimensional?

glwagner:
> Right, I guess for now it is irrelevant because we have a zero-layer model, and concentration is a reduced field in z. I think we haven't decided yet how a multi-layer model will look, but I guess the concentration will always remain two-dimensional?

Correct, concentration, velocity, and total thickness are two-dimensional

navidcy:
Sorry. “This” referred to the changes proposed in this PR. I extracted them from the #456 and was thinking that if it’s a bugfix it deserves to be fixed for all the package, not only when we merge the #456.

glwagner:
> Sorry. “This” referred to the changes proposed in this PR. I extracted them from the #456 and was thinking that if it’s a bugfix it deserves to be fixed for all the package, not only when we merge the #456.

Ah ok. I agree, it is not needed for #456.

PR #565, opened 2025-06-18
(0.7.0) Revamp VerticalGrids + add Docs for VerticalGrids

navidcy:
This is ready for review.

The PR changes the user interface for creating vertical grids.

Also adds a section in the Docs describing the vertical grids. Preview should be here but it might take few hours until you get the final version

https://clima.github.io/ClimaOceanDocumentation/previews/PR565/vertical_grids/

Next should come a PR with a vertical grid implementation that allows us to specify surface spacing, Nz and depth. Will probably involve iterative process, e.g. in the spirit of https://doi.org/10.1016/j.ocemod.2017.03.012


src/VerticalGrids.jl
```diff
@@ -78,18 +81,28 @@ function stretched_vertical_faces(; surface_layer_Δz = 5.0,
...
+@inline exponential_profile(z, Lz, h) = expm1((z + Lz) / h) / expm1(Lz / h)
 
-function exponential_z_faces(; Nz, depth, h = Nz / 4.5)
+"""
+    exponential_z_faces(; Nz, depth, scale = depth / 5)
+
+Return an array of ``z``-faces of length `Nz + 1` with exponential spacing that spans
+the range from the surface (``z=0``) down to ``z = -```depth`.
+"""
+function exponential_z_faces(; Nz, depth, scale = depth / 5)
```
glwagner, line 92: Can you convert this to a callable object so that we only have one breaking change? Like this:

```julia
struct ExponentialFaces{FT}
    size :: Int
    extent :: FT
    scale :: FT
end

# write a function that returns the interfaces
```

navidcy: > I don't know what is specific about "z" to this either

You are referring to the name? Yes it's not, good point.

navidcy: have a look now?
I added types for ExponentialFaces and StretchedFaces.
I'm not that happy for the user-interface now... Have a look at the at https://github.com/CliMA/ClimaOcean.jl/blob/ncc/vertical-grids/docs/src/vertical_grids.md for a demo. Suggestions welcome!

navidcy: (I commented out the examples. When we settle on user interface I can update all examples and uncomment them.)

src/VerticalGrids.jl
```diff
@@ -26,74 +29,139 @@ function (stretching::LinearStretching)(Δz, z)
     return (1 + c) * Δz
 end
 
+struct StretchedFaces{FT, S}
+    extent :: FT
+    top_layer_minimum_spacing :: FT
+    top_layer_height :: FT
+    constant_bottom_spacing_depth :: FT
```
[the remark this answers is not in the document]
navidcy: Yes. That was my thinking exactly. Keep it z-specific to increase readability. Seems strange someone would want this applied laterally..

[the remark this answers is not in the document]
navidcy: One could say that since these belong to `VerticalGrids` they can be all z-specific? The exponential tho was easy to generalise.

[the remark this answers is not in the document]
glwagner: It's all about designing a stretched coordinate, so the ideas can be applied to any direction in rectilinear or lat lon grid, as well as the vertical direction for orthogonal spherical shell grid.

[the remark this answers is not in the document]
navidcy: It's true. Perhaps these do belong in Oceananigans at the end of the day..

navidcy:
How do you like this syntax?

```Julia
zgrid = exponential_vertical_faces(; Nz, depth, scale) # return the grid type, e.g., an ExponentialFaces type
zf = z_faces(zgrid) # returns the faces based on the grid type
```

docs/src/vertical_grids.md
```diff
@@ -121,7 +119,8 @@ But with the larger ``h / L`` is, the smaller the rate the spacings increase wit
 A ridiculously large value of ``h / L`` (approaching infinity) gives a uniform grid:
 
 ```@example vgrids
-z_faces(exponential_vertical_faces(; Nz, depth, scale = 1e20*depth))
+zgrid = exponential_vertical_faces(; Nz, depth, scale = 1e20*depth)
+z_faces(zgrid)
```
glwagner, line 123: what does "zgrid" and z_faces mean here?

src/VerticalGrids.jl
glwagner: I don't think we should introduce new functions called `z_faces` and `z_centers` (not sure I am interpreting that correctly). Instead, let's build the grid and use the `znodes` interface to visualize the grid?

navidcy: Perhaps I never really understood your original suggestion? z_faces builds the grid. Let me think of a better name for it.

docs/src/vertical_grids.md
navidcy: zgrid returns the ExponentialFaces type and z_faces builds the array of the faces given the type

I don't like it; but I'm trying to follow your suggestion to have a special type for each vertical grid. But perhaps in the course of doing that I'm mixing a bit of my thinking with your thinking....?

src/VerticalGrids.jl
navidcy: It could be a matter of (re)naming `z_faces`? Because `z_faces` method _does_ return the faces for the `ExponentialFaces` as you were originally suggesting and it dispatches on the `::ExponentialFaces` or `::StretchedFaces`, or other new vertical grid types we will introduce.

glwagner: I'm wondering if this can be done without introducing any new functions, which I think would be better from a docs/maintenance/simplicity perspective

navidcy: We can brainstorm over Zoom? My main purpose on this PR was to document the current grids and plot examples so we have an idea the effect the parameters (eg `h`, which was changed to `scale`) have on the final grid.

src/VerticalGrids.jl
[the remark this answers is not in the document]
navidcy: If the grids would take `z` as argument we could dispatch on the types of the vertical grids.

navidcy:
@glwagner, have a look now!

docs/src/vertical_grids.md
```diff
@@ -50,20 +50,40 @@ axislegend(ax, position=:rb)
 fig
 ```
 
+Let's see how [`ExponentialFaces`](@ref) works:
```
glwagner, line 53:
```suggestion
Let's see how [`ExponentialInterfaces`](@ref) works:
```

can we use "interfaces" instead of "faces"?

navidcy: agreed; done

src/VerticalGrids.jl
```diff
@@ -23,75 +26,161 @@ function (stretching::LinearStretching)(Δz, z)
...
+function Base.show(io::IO, g::ExponentialInterfaces)
+    print(io, summary(g), '\n')
+    print(io, "├── size: ", prettysummary(g.size), '\n')
+    print(io, "├── extent: ", prettysummary(g.extent), '\n')
+    print(io, "└── scale: ", prettysummary(g.scale), '\n')
+end
+
+function (g::ExponentialInterfaces)(k)
+    Nz, depth, scale = g.size, g.extent, g.scale
+
```
navidcy, line 169: I'm wondering whether a validation is appropriate here, e.g.,

```suggestion

(!isinteger(k) || k < 1 || k > Nz+1) && throw(ArgumentError("only integer arguments within 1 and $(Nz+1)"))

```

navidcy:
merge?

examples/mediterranean_simulation_with_ecco_restoring.jl
```diff
@@ -26,35 +26,38 @@ using Dates
...
+
+z = StretchedInterfaces(; depth = 5000,
+                        surface_layer_Δz = 2.5,
+                        stretching = PowerLawStretching(1.07),
+                        surface_layer_height = 50)
 
 Nx = 15 * Int(λ₂ - λ₁) # 1/15 th of a degree resolution
 Ny = 15 * Int(φ₂ - φ₁) # 1/15 th of a degree resolution
-Nz = length(z_faces) - 1
+Nz = length(z)
```
glwagner, line 47: Shouldn't this still be `Nz = length(z) - 1` if `z` represents the interfaces?

glwagner: Perhaps we can think of `z` as representing the coordinate more generally but it is called "Interfaces" now

experiments/arctic_simulation.jl
```diff
@@ -14,8 +14,10 @@ using CUDA
 CUDA.device!(1)
 arch = GPU()
 
-r_faces = ClimaOcean.exponential_z_faces(; Nz=30, h=10, depth=2000)
-z_faces = MutableVerticalDiscretization(r_faces)
+depth = 2000meters
+Nz = 30
+z = ExponentialInterfaces(Nz, depth; scale=depth/3)
+z = MutableVerticalDiscretization([z(k) for k in 1:Nz+1])
```
glwagner, line 20:
```suggestion
z = MutableVerticalDiscretization(z)
```

doesn't this work?

experiments/coupled_simulation/ocean_sea_ice_coupled_simulation.jl
```diff
@@ -18,15 +18,14 @@ arch = CPU()
 
 depth = 1000meters
 Nz    = 10
-h     = 3
 
-r_faces = ClimaOcean.exponential_z_faces(; Nz, h, depth)
-z_faces = MutableVerticalDiscretization(r_faces)
+z = ExponentialInterfaces(Nz, depth; scale=0.3*depth)
+z = MutableVerticalDiscretization([z(k) for k in 1:Nz+1])
```
glwagner, line 23: Let's avoid this pattern and just use `MutableVerticalDiscretization(z)` as we do normally

src/VerticalGrids.jl
```diff
@@ -23,75 +26,246 @@ function (stretching::LinearStretching)(Δz, z)
...
+                                            stretching,
+                                            rounding_digits)
+        S = typeof(stretching)
+        A = typeof(z_faces)
+        return new{S, A}(extent, top_layer_minimum_spacing, top_layer_height,
+                         constant_bottom_spacing_depth, maximum_spacing, stretching, z_faces)
+    end
+end
+
+function calculate_stretched_faces(; extent = 5000,
```
glwagner, line 60:
```suggestion
function compute_stretched_faces(; extent = 5000,
```

src/VerticalGrids.jl
```diff
@@ -23,75 +26,246 @@ function (stretching::LinearStretching)(Δz, z)
...
+                                   top_layer_minimum_spacing = 5.0,
+                                   top_layer_height = 100.0,
```
glwagner, line 62: why the `.0` here? Is Float64 important?

glwagner (approved):
I think "Grid utilities" is a better name than "Vertical grids" now. but this belongs in Oceananigans too

experiments/arctic_simulation.jl
navidcy: I didn't try, perhaps.....

experiments/coupled_simulation/ocean_sea_ice_coupled_simulation.jl
navidcy: I wasn't sure it would work; I'll try it out

examples/mediterranean_simulation_with_ecco_restoring.jl
navidcy: Is `ExponentialCoordinate` more appropriate you reckon?

navidcy:
Merging now.

https://github.com/CliMA/Oceananigans.jl/pull/4623 aims to move ExponentialCoordinate to Oceananigans. When that's merged we can drop it from ClimaOcean.

PR #579, opened 2025-07-21
Fix sign error on stability variable

simone-silvestri:
which might be the cause of SST increasing drastically.


navidcy:
> hold off on more commits until the docs build?

Yes. I'm as excited to see the docs! I only committed because I saw the docs failed and had to restart them anyways ;)

navidcy:
preview at https://clima.github.io/ClimaOceanDocumentation/previews/PR579

looks GOOD! SST don't increase much from the initial condition now!

navidcy (approved):

PR #580, opened 2025-07-22
Improve plot in one degree simulation example

glwagner:
Just a minor change. Also I didn't feel quite honest calling it a "pretty" movie. Ideally there would be six full panels. It's good enough though, so I just changed the text to be more honest.

### DIFF
--- a/examples/one_degree_simulation.jl
+++ b/examples/one_degree_simulation.jl
@@ -177,9 +177,9 @@ simulation.Δt = 20minutes
 simulation.stop_time = 365days
 run!(simulation)
 
-# ### A pretty movie
+# ### A movie
 #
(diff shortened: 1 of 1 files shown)


navidcy (approved):
Fair! :)

examples/one_degree_simulation.jl
```diff
@@ -256,7 +256,7 @@ end
 # Finally, we plot a snapshot of the surface speed, temperature, and the turbulent
 # eddy kinetic energy from the CATKE vertical mixing parameterization as well as the
 # sea ice speed and the effective sea ice thickness.
-fig = Figure(size = (1200, 1200))
```
navidcy, line 259: why this?

glwagner: This is what it looks like: 5 panels, squished in a very disturbing way

PR #585, opened 2025-07-24
Increase timestep for the one-degree global simulation

navidcy:
I managed to get a stable simulation for 2 years with 8minute initialization timestep for 20days and then switching to 30minutes.

https://github.com/user-attachments/assets/13dd3162-98e6-41b3-acda-a0b8e776029d

(We should wait for #583 first.)

### DIFF
--- a/examples/one_degree_simulation.jl
+++ b/examples/one_degree_simulation.jl
@@ -92,7 +92,7 @@ set!(seaice.model, h=ecco_sea_ice_thickness, ℵ=ecco_sea_ice_concentration)
 
 # ### Atmospheric forcing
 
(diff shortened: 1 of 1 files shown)


navidcy:
Preview at https://clima.github.io/ClimaOceanDocumentation/previews/PR585/literated/one_degree_simulation/

looks good!

navidcy:
Fix it! If you know what let’s do it

simone-silvestri:
Actually now that I see it, it looks like it is correctly coupled because 
it is `sea_ice_simulation(grid, ocean; kw...)` I was saying it because there was a bug in the omip branch where the two were not correctly coupled.

navidcy:
Ok. I’ll merge

PR #588, opened 2025-07-27
Patch required for Oceananigans >=0.97.4

### LINKED ISSUE #589: CondaPkg fails to install copernicusmarine
Copernicus Marine Store Downloading GithubAction has been failing for a while now.

I managed to track down the issue and it was due to some python package incompatibility that was coming from ReactantPythonCallExt; see https://github.com/EnzymeAD/Reactant.jl/blob/main/CondaPkg.toml

navidcy:
After https://github.com/CliMA/Oceananigans.jl/pull/4668, the`Oceananigans.BoundaryConditions.FieldBoundaryConditions` inputs changed; one needs to give instantiated location tuple.

Also closes #589


navidcy:
I think this is ready to be merged!

taimoorsohail (approved):

PR #594, opened 2025-08-01
Expose `include_rivers_and_icebergs` kwarg instead of using the default

navidcy:
Despite `include_rivers_and_icebergs = false` is the default, I think it's good idea to expose it so that it's obvious to anybody that they need to tweak this to include runoff and icebergs.

We might want to consider making `include_rivers_and_icebergs = true` the default at some point.

### DIFF
--- a/examples/near_global_ocean_simulation.jl
+++ b/examples/near_global_ocean_simulation.jl
@@ -105,7 +105,8 @@ radiation = Radiation(arch)
 # The number of snapshots that are loaded into memory is determined by
 # the `backend`. Here, we load 41 snapshots at a time into memory.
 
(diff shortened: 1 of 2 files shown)


glwagner:
> We might want to consider making include_rivers_and_icebergs = true the default at some point.

Does it change the solution? The fluxes are incredibly weak.

navidcy:
> It's worth noting that we wouild also like ot move this to a "land" component for OceanSeaIceModel. These are properly understood as fluxes form a land component, not from an atmosphere component.

That's a good point. These are not atmospheric-related fields...
I think the river and icebergs are not originally part of JRA55 reanalysis; they are only included in the JRE55-do (=drive ocean) version.

I'm merging since this PR doesn't change anything other than making the kwarg more visible in the examples.

PR #602, opened 2025-08-21
Use the Veros ocean model as the ocean component in ClimaOcean

simone-silvestri:
This is an exploratory PR to use ClimaOcean to force a veros setup. 

This exercise is also usefull to make sure that ClimaOcean does not necessarily assume that the ocean model comes from Oceananigans but it is flexible enough to handle different ocean configurations.


ext/ClimaOceanPythonCallExt/veros_ocean_simulation.jl
```diff
@@ -0,0 +1,205 @@
...
+    end
+
+    Nx = length(xc) 
+    Ny = length(yc) 
+
+    return LatitudeLongitudeGrid(size=(Nx, Ny), longitude=xf, latitude=yf, topology=(TX, Bounded, Flat), halo=(2, 2))
+end
+
+"""
+    veros_set!(ocean, v, x)
```
navidcy, line 136: why not just `set!(ocean::VerosOceanSimulation, v, x)`?
because there are 3 args now?

ext/ClimaOceanPythonCallExt/ClimaOceanPythonCallExt.jl
```diff
@@ -0,0 +1,15 @@
...
+
+using ClimaOcean
+using CondaPkg
+using PythonCall
+using Oceananigans
+using Oceananigans.DistributedComputations: @root
+
+using Dates: DateTime
+
+include("clima_ocean_copernicus.jl")
```
glwagner, line 11:
```suggestion
include("copernicus.jl")
```

src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
```diff
@@ -56,13 +60,13 @@ function compute_net_ocean_fluxes!(coupled_model)
     freshwater_flux = atmosphere_fields.Mp.data
 
     ice_concentration = sea_ice_concentration(sea_ice)
-    ocean_salinity = ocean.model.tracers.S
+    ocean_salinity = get_ocean_state(coupled_model.ocean, coupled_model.interfaces.exchanger).S
```
glwagner, line 63: its hard to read, try

```suggestion
    ocean_state = get_ocean_state(coupled_model.ocean, coupled_model.interfaces.exchanger)
    ocean_salinity = ocean_state.S
```

src/OceanSeaIceModels/InterfaceComputations/assemble_net_fluxes.jl
```diff
@@ -80,6 +84,8 @@ function compute_net_ocean_fluxes!(coupled_model)
             atmos_ocean_properties,
             ocean_properties)
 
+    fill_up_net_fluxes!(coupled_model.ocean, net_ocean_fluxes)
```
glwagner, line 87:
```suggestion
    fill_net_fluxes!(coupled_model.ocean, net_ocean_fluxes)
```

experiments/veros_forced_simulation.jl
```diff
@@ -0,0 +1,92 @@
...
+VerosModule.remove_outputs(:global_4deg)
+
+# Actually loading and instantiating the Veros setup in the variable `ocean`.
+# This setup uses by default a different time-step for tracers and momentum, 
+# so we set it to the same value (1800 seconds) for both.
+
+ocean = VerosModule.VerosOceanSimulation("global_4deg", :GlobalFourDegreeSetup)
+
+VerosModule.veros_settings_set!(ocean, "dt_tracer", 1800.0)
+VerosModule.veros_settings_set!(ocean, "dt_mom",    1800.0)
```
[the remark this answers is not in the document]
simone-silvestri: right I think we can just extend the `set!` method

simone-silvestri:
I think this is ready on the ClimaOcean side of things, I'm not sure whether to add tests or not since it might create issues with CI. 
The only thing I need to sort out before merging is this function
https://github.com/CliMA/ClimaOcean.jl/blob/ac8d821ba43989cecad78c7a4c2b310ab932c60e/ext/ClimaOceanPythonCallExt/veros_state_exchanger.jl#L86-L115
which we are sorting out on the veros side today.

simone-silvestri:
I think this is ready to merge. We probably have to tweak the extension again after https://github.com/nutrik/veros/tree/atm_coupling is merged into main, and we have space where to fill in salt and temperature flux.

examples/python_ocean_forced_simulation.jl
```diff
@@ -0,0 +1,125 @@
...
+
+# We set up a progress callback that will print the current time, iteration, and maximum velocities
+# at every 5 iterations. It also collects the surface velocity fields and the net fluxes
+# into the arrays `s`, `tx`, and `ty` for later visualization.
+
+wall_time = Ref(time_ns())
+
+s  = []
+tx = []
+ty = []
```
glwagner: why can't you use a regular JLD2Writer? I think that would be preferred, to illustrate the user interface which is important.

ext/ClimaOceanPythonCallExt/veros_state_exchanger.jl
```diff
@@ -0,0 +1,114 @@
...
+
+function fill_net_fluxes!(ocean::VerosOceanSimulation, net_ocean_fluxes)
+    nx = pyconvert(Int, ocean.setup.state.settings.nx) + 4
+    ny = pyconvert(Int, ocean.setup.state.settings.ny) + 4
+
+    ρₒ = pyconvert(eltype(ocean), ocean.setup.state.settings.rho_0)
+    taux = view(parent(net_ocean_fluxes.u), 1:nx, 1:ny, 1) .* ρₒ
+    tauy = view(parent(net_ocean_fluxes.v), 1:nx, 1:ny, 1) .* ρₒ
+
+    # TODO: Do not do this 12 thingy when we can make sure
```
glwagner: what is the "12 thingy"?

ext/ClimaOceanPythonCallExt/veros_ocean_simulation.jl
```diff
@@ -0,0 +1,202 @@
...
+==========
+- `setup::AbstractString`: The name of the Veros setup module to import (e.g., `"global_4deg"`).
+- `setup_name::Symbol`: The name of the setup class or function within the module to instantiate (e.g., `:GlobalFourDegreeSetup`).
+"""
+function VerosOceanSimulation(setup, setup_name::Symbol)
+    setups = pyimport("veros.setups." * setup)
+    setup  = @eval $setups.$setup_name()
+
+    # instantiate the setup
+    setup.setup()
```
glwagner: is `setup` the best name for this object, given that it has a method called `setup()`? Perhaps the function `setup()` should be something like `instantiate()` so it's out of our control...

ext/ClimaOceanPythonCallExt/veros_state_exchanger.jl
```diff
@@ -0,0 +1,114 @@
...
+    exchange_atmosphere_state :: AST
+    atmosphere_exchanger :: AEX
+end
+
+mutable struct ExchangeOceanState{FC, CF, CC}
+    u  :: FC
+    v  :: CF
+    T  :: CC
+    S  :: CC
+end
```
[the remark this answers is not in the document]
simone-silvestri: I think we can use a general exchanger. We need to add "ocean" fields that point to the already existing fields in case of an `OceananigansSimulation`, or more generally, in case of an ocean simulation which matches the exchanger grid.

ext/ClimaOceanPythonCallExt/veros_ocean_simulation.jl
simone-silvestri: Indeed, `setup()` is the veros method, so we cannot change it. We can change the first `setup` although I agree with you that I tend to think as a `setup` as something concrete (like a type) rather than a method.
We can maybe name it `configuration`?

ext/ClimaOceanPythonCallExt/veros_state_exchanger.jl
simone-silvestri: > Is that only a feature of this particular setup (eg other veros setups are more flexible)?

actually I am not necessarily sure about veros' API. It seems that there are preconfigured setups that can be used by defining custom initial condition functions, boundary conditions and so on... so that part looks a bit more oceananigans-like. On the other hand, there seems to be options which require namelist inputs, a bit more a-la-fortran per se. The lines I am referring at are these ones: https://github.com/team-ocean/veros/blob/a6b526f0e648f9fd12bc7d385d9aed7ca46afb78/veros/setups/global_4deg/global_4deg.py#L171-L194
So it seems that these `taux` and `tauy` objects, belong to the setup rather than the actual veros class
https://github.com/team-ocean/veros/blob/a6b526f0e648f9fd12bc7d385d9aed7ca46afb78/veros/setups/global_4deg/global_4deg.py#L92-L99

simone-silvestri:
This should be ready to go. The only problem here is that the output writers do not work yet with and `OceanSeaIceModel`. We can deal with it once we refactor the name.

src/OceanSeaIceModels/InterfaceComputations/sea_ice_ocean_fluxes.jl
```diff
@@ -24,6 +24,7 @@ function compute_sea_ice_ocean_fluxes!(sea_ice_ocean_fluxes, ocean, sea_ice, mel
     ℵᵢ = sea_ice.model.ice_concentration
     hᵢ = sea_ice.model.ice_thickness
     Gh = sea_ice.model.ice_thermodynamics.thermodynamic_tendency
+    Δt = sea_ice.Δt
```
simone-silvestri:
```suggestion
```

simone-silvestri:
because an output at the moment requires the model to have a grid. In this case the "grid" of the specific `OceanSeaIceModel` can be the exchange grid I think.

simone-silvestri:
Finally the docs ran https://clima.github.io/ClimaOceanDocumentation/previews/PR602/literated/python_ocean_forced_simulation/ !

examples/veros_ocean_forced_simulation.jl
glwagner: I suggest the name: veros_ocean_simulation.jl

simone-silvestri:
Ok, I think the docs willl pass, so this PR is probably ready.

docs/src/developers/slab_sea_ice.jl
```diff
@@ -0,0 +1,212 @@
+# # Implementing a Slab Sea Ice Component
```
simone-silvestri: This file slipped in, it was meant for another PR. I will delete it after the docs build so I don't need to wait that the docs build once again

PR #607, opened 2025-08-29
Add Copernicus BGC DataWrangling

vtamsitt:
Added GLORYS 1/4 degree BGC dataset metadata to Copernicus.jl

A couple of things:
1) There is no inpainting for sea ice variables  because zeros are physical, not missing values'. This should also be the case for chlorophyll, phytoplankton, primary productivity I think? So should probably remove inpainting for these.
2) I split the dicts of variable names up for the data streams that have different variables (similar to ECCO.jl), but the metadata_prefix variable naming hasn't been updated to reflect this yet. It would be good to add a download_dataset function for Copernicus.jl as well and I'm not sure the file naming convention from Copernicus matches what's currently in metadata_prefix.


vtamsitt:
Oh I see download_dataset is addressed in #522

src/DataWrangling/Copernicus/Copernicus.jl
```diff
@@ -78,12 +94,39 @@ copernicus_dataset_variable_names = Dict(
...
+)
+
+copernicus_bgc_daily_dataset_variable_names = Dict(
+    :total_chlorophyll => "chl",
+    :primary_production => "nppv",
+    :nitrate => "no3",
+    :phosphate => "po4",
+    :dissolved_silicate => "si",
+    :dissolved_oxygen => "o2",
+)
```
glwagner, line 117: It looks possible to lump all the names into a single "main dictionary" for all `CopernicusMetadata` --- should we do that? happy to keep them separate if needed or if its better that way.

src/DataWrangling/Copernicus/Copernicus.jl
```diff
@@ -78,12 +94,39 @@ copernicus_dataset_variable_names = Dict(
...
+
 start_date_str(date) = string(date)
 end_date_str(date) = string(date)
 start_date_str(dates::AbstractVector) = first(dates) |> string
 end_date_str(dates::AbstractVector) = last(dates) |> string
 
-dataset_variable_name(metadata::CopernicusMetadata) = copernicus_dataset_variable_names[metadata.name]
+#dataset_variable_name(metadata::CopernicusMetadata) = copernicus_dataset_variable_names[metadata.name]
```
navidcy, line 124: why comment; let's delete?

src/DataWrangling/Copernicus/Copernicus.jl
navidcy: hm.. I think the `available_variables` won't work with that suggestion...

navidcy (approved):

src/DataWrangling/Copernicus/Copernicus.jl
vtamsitt: I initially lumped them together and then thought it might make more sense to separate so it's not on the user to know which variables are in which data product (and because I saw there are separate dictionaries in ECCO.jl for ECCO2 and ECCO4).

glwagner: > hm.. I think the `available_variables` won't work with that suggestion...

can you clarify?

src/DataWrangling/Copernicus/Copernicus.jl
```diff
@@ -60,12 +68,20 @@ copernicusmarine_dataset_id(::CMEMSHourlyAnalysis) = "cmems_mod_glo_phy_anfc_0.0
...
+Base.size(::GLORYSBGCDaily) = (1440, 680, 75, 1)
+Base.size(::GLORYSBGCMonthly) = (1440, 680, 75, 1)
 reversed_vertical_axis(::CopernicusDataset) = true
 
-available_variables(::CopernicusDataset) = copernicus_dataset_variable_names
+available_variables(::GLORYSStatic) = copernicus_physics_dataset_variable_names
+available_variables(::GLORYSDaily) = copernicus_physics_dataset_variable_names
+available_variables(::GLORYSMonthly) = copernicus_physics_dataset_variable_names
+available_variables(::GLORYSBGCDaily) = copernicus_bgc_daily_dataset_variable_names
+available_variables(::GLORYSBGCMonthly) = copernicus_bgc_monthly_dataset_variable_names
```
navidcy, line 82: how could we have `available_variables` give out different list of variables based on the dataset type if all variables are in one dictionary?

vtamsitt: @glwagner does this look ok now?

glwagner: yes, thank you! I can eliminate the repeated code by creating a more organized type hierarchy. do you mind if I commit to your branch?

vtamsitt: Go for it!

PR #628, opened 2025-09-26
Use `AdvectiveFormulation` for the skew fluxes

simone-silvestri:
@xkykai has found out that the advective formulation is less diffusive than the diffusive formulation. Also a bug in the computation of the slope has been fixed, so I wanted to test it out and see if the results were still egregious in the one-degree simulation


navidcy:
What’s exactly is the catastrophic dispersion you are referring to?

simone-silvestri:
It was happening with the previous version of oceananigans (before 0.99.2) that did not include this PR https://github.com/CliMA/Oceananigans.jl/pull/4810 at the time of this comment
> Results are terrible! However, I think the fix from Oceananigans#main has not been used yet because we haven't released a new version. We can wait for the new version to be released, and I will test it again here.
basically a lot of noise was propagating in the solution (I think the preview has since been deleted).

simone-silvestri:
Not sure, I think @taimoorsohail was using the diffusive formulation, the advective formulation was leading to a unusable solution.

Issue #636, opened 2025-10-01
`ECCODarwin` download fails


simone-silvestri:
Apparently the tests cannot find the `:PO₄` key
see https://buildkite.com/clima/climaocean-ci/builds/4852#01999ca3-f1c2-4c9e-826a-67fa32cdc911

@seamanticscience

navidcy:
related: #639

seamanticscience:
I believe the `ECCO2DarwinMonthly` tests work, but the interpolation from LLC270 "binary" grid to the 0.25° lat-lon "native" grid takes a relatively long time, and the CI test just stops before it is finished. I guess there's no reason not to upload the coefficients that I already calculated (since it's the same every time) and allow `MeshArrays` to download it. A similar thing is already available for the LLC90 grid to 0.5° lat-lon grid, but the interpolation here is much faster anyway. Plan?

seamanticscience:
@glwagner coefficients -> interpolation weights? Here's the relevant code:
https://github.com/CliMA/ClimaOcean.jl/blob/8def2b6b2c7c802d452c2dbfd35859eacfc869ce/src/DataWrangling/ECCO/ECCO_darwin.jl#L132-L152

If `interp_file` does not exist, the weights are computed; otherwise, the weights are loaded from the file, which is fine for regular use where Julia artifacts are preserved. I'm guessing the CI tests start afresh each time and therefore have to recompute everything from scratch. This works fine for LLC90 (`ECCO4DarwinMonthly`) but is obviously too much of a heavy-lift for LLC270/`ECCO2DarwinMonthly` testing (CPU metadata tests for ECCO2DarwinMonthly take about an hour). 

However, `MeshArrays` has a neat trick, where if the filename is precisely `interp_coeffs_halfdeg.jld2` then it will download a set of precalculated LLC90 -> 0.5° lat lon (i.e. `binary_data_grid` to `native_grid`) weights from Zenodo. I am proposing adding the same trick for an LLC270 -> 0.25° lat lon weight file.

PR #637, opened 2025-10-01
Drop ECCODarwin tests

simone-silvestri:
temporarily bypasses #636


seamanticscience:
I'm not sure what's up with `ECCO2DarwinMonthly`, but at least `ECCO4Darwin` is working correctly!

PR #670, opened 2025-11-06
Build examples on julia 1.12

simone-silvestri:
Let's see if this works...


simone-silvestri:
Looks like this works. I guess we can officially move to julia 1.12 also on ClimaOcean

navidcy (approved):

PR #677, opened 2025-11-14
implement a nonhydrostatic LES example

glwagner:
Forced by a prescribed atmosphere state representing a clear night sky.


simone-silvestri (approved):

glwagner:
@simone-silvestri what do you think about adding this to the documentation?

PR #679, opened 2025-11-16
Use RK3 time stepper for near-global 1/4 degree example

navidcy:


simone-silvestri:
yeah `diffusivity_fields` became `closure_fields`. We can merge the speedyweather ext that has this fix.

simone-silvestri:
> I pushed it up to 20min and it looks just fine!

Maybe after the timestepping PR, we can move to 30 minutes with zstar :)

PR #680, opened 2025-11-17
Remove files after ourselves in `test_downloading.jl` and remove julia 1.12 Reactant tests

simone-silvestri:
this should solve the issue of not having enough space on the github action device and the problem with Reactant, which does not seem ready for 1.12


simone-silvestri:
The problem of the invalid multi-year JRA55 link resurfaced...

PR #681, opened 2025-11-18
Precompute buoyancy gradients to speed up example execution

simone-silvestri:
Added buoyancy gradients materialization to improve computation speed.


glwagner (approved):

navidcy:
The docs with this feature seem to get out of memory.... I've run the docs multiple times.

cc @simone-silvestri

PR #684, opened 2025-11-20
Try increasing time step for near global example

navidcy:
Comment out simulation files in the to_be_literated array and pages section.


navidcy:
The example ran! I'll merge.

PR #688, opened 2025-12-01
Add DateTime clock timestepping test for OceanSeaIceModel

navidcy:
This PR adds a timestepping test for the OceanSeaIceModel. It also fixes a bunch of import-related errors that have to do with ExplicitImport changes in Oceananigans (I hope I did that correctly; cc @giordano).

This PR depends on https://github.com/JuliaRegistries/General/pull/143185 because it needs Oceananigans v0.102.1 that includes https://github.com/CliMA/Oceananigans.jl/pull/4885

### DIFF
--- a/src/OceanSeaIceModels/ocean_sea_ice_model.jl
+++ b/src/OceanSeaIceModels/ocean_sea_ice_model.jl
@@ -55,6 +55,29 @@ prognostic_fields(cm::OSIM)         = nothing
 fields(::OSIM)                      = NamedTuple()
 default_clock(TT)                   = Oceananigans.TimeSteppers.Clock{TT}(0, 0, 1)
 
+"""Return a clock from a component if available."""
+function component_clock(component)
+    if component === nothing
+        return nothing
+    elseif hasproperty(component, :clock)
+        return component.clock
+    elseif hasproperty(component, :model) && hasproperty(component.model, :clock)
+        return component.model.clock
+    else
+        return nothing
+    end
+end
+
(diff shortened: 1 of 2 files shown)


simone-silvestri (approved):

simone-silvestri:
apparently that started happening as we switched to julia 1.12

navidcy:
Seems that the error is originating at:

https://github.com/CliMA/Oceananigans.jl/blob/396bf450578cc2c7fbcaeca6b14c026e4a4d5da1/src/OutputReaders/field_time_series_indexing.jl#L115

?

src/OceanSeaIceModels/ocean_sea_ice_model.jl
```diff
@@ -55,6 +55,29 @@ prognostic_fields(cm::OSIM)         = nothing
...
 default_clock(TT)                   = Oceananigans.TimeSteppers.Clock{TT}(0, 0, 1)
 
+"""Return a clock from a component if available."""
+function component_clock(component)
+    if component === nothing
+        return nothing
+    elseif hasproperty(component, :clock)
+        return component.clock
+    elseif hasproperty(component, :model) && hasproperty(component.model, :clock)
+        return component.model.clock
```
glwagner: I think a better design is to use a function liek `get_clock` (or a better name)

glwagner: but then you need a PR to Oceananigans to define this for `Simulation`

PR #694, opened 2025-12-02
Add tests for GPU distributed DatasetRestoring

simone-silvestri:
I am having some issues with `DatasetRestoring` on multiple GPUs. The `update_model_field_time_series!` function crashes (deterministically) with CUDA illegal memory access connected to Oceananigans' `cpu_interpolating_time_indices`.

I am trying to debug this as it is halting the OMIP progress, however, in the meantime I am adding a dataset restoring GPU test for multi-GPU here to see if I can reproduce the error.

I think the tests need a bit of an overhaul.

### DIFF
--- a/test/runtests_setup.jl
+++ b/test/runtests_setup.jl
@@ -35,34 +35,34 @@ test_datasets = (ECCO2Monthly(),
                  ECCO4Monthly(), 
                  ECCO2DarwinMonthly(),
                  ECCO4DarwinMonthly(),
-                 EN4Monthly(),
-                )
+                 EN4Monthly())
 
 test_names = Dict(
-    ECCO2Monthly() => (:temperature, :salinity),
-    ECCO2Daily() => (:temperature, :salinity),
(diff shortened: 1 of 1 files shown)


simone-silvestri:
Actually, the error I was finding, which was a CUDA Illegal memory access, is just connected to the fact that we had a
```julia
ERROR: DomainError with -1.0:
sqrt was called with a negative real argument but will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
Stacktrace:
 [1] throw_complex_domainerror(f::Symbol, x::Float64)
   @ Base.Math ./math.jl:33
 [2] sqrt
   @ ./math.jl:686 [inlined]
 [3] sqrt(x::Int64)
   @ Base.Math ./math.jl:1578
 [4] top-level scope
   @ REPL[1]:1
```
error. 
Apparently, this error is not shown on the GPU, but it corrupts the GPU memory which then eventually spits out the `CUDA Illegal memory access`.

PR #697, opened 2025-12-05
Add ocean horizontal viscosity to 1 deg ocean simulation example

glwagner:
@jlk9 @heimbach


navidcy:
I've never seen that warning before. Did we change something recently related to ScalarBiharmonicDiffusivity?

glwagner:
> Here are the results with 64 x 64 x 40 and up to 6hours (averaged every 1h).
> 
> bomex_profiles_1hrs

wrong PR :-D

navidcy:
It's because the JLD2Writer is trying to save the function 

https://github.com/CliMA/ClimaOcean.jl/blob/bfae2f72a20e351afbeb167bfa74dd39013743cf/examples/one_degree_simulation.jl#L57

simone-silvestri:
for that specific simulation I was dumping individual files because it was helping me identify where numerical errors were popping up. I can open an issue and look at a solving it this week.

glwagner:
> it was helping me identify where numerical errors were popping up

please open an issue for the feature that you need, which will help you identify numerical errors, so that we can all use the same code...

simone-silvestri:
> Maybe we should also automatically add something like this to coarse setups inside ocean_simulation?

make sense, so make `default_ocean_closure` depend on the grid?

PR #699, opened 2025-12-09
(0.9.0) Modify interface to generalize the coupling

simone-silvestri:
At the moment, the ocean is predominant in ClimaOcean. However, this PR is an attempt to generalize the code to allow different ocean simulations as well as different atmosphere.

The main objective of this PR is to remove component-specific methods from `InterfaceComputations` and bring them into appropriate `OceanSimulations`, `SeaIceSimulations` and `AtmosphereSimulations` modules.

In addition another objective is to generalize the exchanger which means adding an `interpolate_sea_ice_state!` and `interpolate_ocean_state!` functions.

I am a bit unsure about the `compute_net_*_fluxes!` methods. Should they live in the `InterfaceComputations` module or in the components modules? I feel like the latter option would be cleaner but potentially lead to code duplication.

### DIFF
--- a/src/ClimaOcean.jl
+++ b/src/ClimaOcean.jl
@@ -43,6 +43,7 @@ export
     ocean_simulation,
     sea_ice_simulation,
     atmosphere_simulation,
+    sea_ice_dynamics,
     initialize!
 
 using Oceananigans


glwagner:
> This might be ready to merge when tests pass. The last thing I am a bit undecided about is the position of `FreezingLimitedOceanTemperature` which probably should be in `SeaIces` instead of `OceanSeaIceModels`.

yes it must be, is it not now?

src/OceanSeaIceModels/components.jl
```diff
@@ -15,6 +15,7 @@ ocean_surface_velocities(ocean) = ZeroField(), ZeroField()
 
 sea_ice_thickness(::Nothing) = ZeroField()
 sea_ice_concentration(::Nothing) = ZeroField()
+function default_sea_ice end
```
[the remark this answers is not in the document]
glwagner: on second thought --- don't bother

I think we will refactor this. Once we have a more general `EarthSystemModel`, we then need to create special subtypes with alternative constructors.

Now I am thinking, probably these specialty constructors will be added in a module that comes _after_ the component modules Oceans, Atmospheres, SeaIces. 

For example, with a PrescribedOcean, we wouldn't use a FreezingLimitedTemperature sea ice component.

docs/src/developers/slab_ocean.jl
```diff
@@ -0,0 +1,182 @@
...
+#
+# First, we define a struct to hold the slab ocean state and its constructor. We also define a summary and show method for the slab ocean type.
+
+using Oceananigans, Base
+
+struct SlabOcean{G, C, T, S, F}
+    grid :: G
+    clock :: C
+    temperature :: T   
+    salinity :: S
```
glwagner: totally throw away comment, but for an example we should be as SIMPLE as possible, so maybe salinity can be ignored?

glwagner (approved):
epic

Project.toml
navidcy: is this a breaking change? do we need to bump minor release?

PR #703, opened 2025-12-12
Remove duplicate Oceananigans compat in docs

navidcy:
Given that Oceananigans has a compat imposed by ClimaOcean I don't see why we need to explicitly redefine it. When we do, it leads to inconsistencies like, eg., v0.8.8 where ClimaOcean requires 0.102 while docs 0.101 and thus the docs never manage to built.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaOcean"
 uuid = "0376089a-ecfe-4b0e-a64f-9c555d74d754"
 license = "MIT"
-version = "0.8.8"
+version = "0.8.9"
 authors = ["Climate Modeling Alliance and contributors"]
 
 [deps]


giordano (approved):
I think this is ok, but a problem is that the `docs/` (and `test/`) environment is resolved separately from the top-level one.

You could add the [`[sources]` section](https://pkgdocs.julialang.org/v1/toml-files/#The-%5Bsources%5D-section) to the two downstream environments (see, e.g., https://github.com/NumericalEarth/Breeze.jl/blob/74abb436145ae435d29376496b7fc9c6283c9cc4/docs/Project.toml#L15-L16) and/or the [`[workspaces]` section](https://pkgdocs.julialang.org/v1/toml-files/#The-[workspace]-section) to the top-level one (see, e.g., https://github.com/NumericalEarth/Breeze.jl/pull/208), which effectively unifies all the environments in the workspace for the purpose of resolution.  You can have both sections, they don't conflict.

glwagner (approved):

navidcy:
> You could add the [`[sources]` section](https://pkgdocs.julialang.org/v1/toml-files/#The-%5Bsources%5D-section) to the two downstream environments (see, e.g., https://github.com/NumericalEarth/Breeze.jl/blob/74abb436145ae435d29376496b7fc9c6283c9cc4/docs/Project.toml#L15-L16) and/or the [`[workspaces]` section](https://pkgdocs.julialang.org/v1/toml-files/#The-%5Bworkspace%5D-section) to the top-level one (see, e.g., [NumericalEarth/Breeze.jl#208](https://github.com/NumericalEarth/Breeze.jl/pull/208)), which effectively unifies all the environments in the workspace for the purpose of resolution. You can have both sections, they don't conflict.

I'll merge this and consider this for later!

PR #708, opened 2025-12-18
Fix major basins behavior

simone-silvestri:
The choice of major basins is performed on main, on an extended grid that is used to compute the connectivity in case of a Periodic direction. It might be, therefore, that large basins such as the Caspian sea are repeated within the extended grid, and are removed from the extended grid without changing the interior of the bottom height which is what we actually want to mask out. 

In this PR we still compute the connectivity (the labels) using an extended grid, but we rank the basins only on the labels in the interior of the grid, correcting the behavior of `major_basins`. 
For example the output of this script:
```julia
grid = TripolarGrid(size = (360, 180, 1), z = (-5000, 0))
bottom_fields = [Array(interior(regrid_bathymetry(grid; minimum_depth = 10,  major_basins = i), :, :, 1)) for i in 1:6]
diffs = [bottom_fields[i] .- bottom_fields[i+1] for i in 1:5]

fig = Figure(size = (800, 1400), fontsize = 14)

for i in 1:5
    ax = Axis(fig[i, 1], title = "major_basins = $(i+1) − $i", aspect = DataAspect())
    heatmap!(ax, diffs[i]; colorrange = (-0.001, 0.001))
    hidedecorations!(ax)
end
```

on main:
bathymetry_main

on this PR:
bathymetry_PR

cc @taimoorsohail

### DIFF
--- a/src/DataWrangling/metadata.jl
+++ b/src/DataWrangling/metadata.jl
@@ -65,12 +65,12 @@ Keyword Arguments
            For a single date, use [`Metadatum`](@ref).
 
 - `start_date`: If `dates = nothing`, we can prescribe the first date of metadata as a date
-                (`Dates.AbstractDateTime` or `CFTime.AbstractCFDateTime`). `start_date` should lie
-                within the date range of the dataset. Default: nothing.
+                (`Dates.AbstractDateTime` or `CFTime.AbstractCFDateTime`). If outside the 
+                date range of the dataset, the first allowable date is chosen. Default: nothing.
 
 - `end_date`: If `dates = nothing`, we can prescribe the last date of metadata as a date
-              (`Dates.AbstractDateTime` or `CFTime.AbstractCFDateTime`). `end_date` should lie
-              within the date range of the dataset. Default: nothing.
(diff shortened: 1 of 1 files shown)


src/Bathymetry.jl
```diff
@@ -324,24 +328,38 @@ function remove_minor_basins!(zb, keep_major_basins)
     connectivity = ImageMorphology.strel(water)
     labels = ImageMorphology.label_components(connectivity)
 
-    total_elements = zeros(maximum(labels))
-    label_elements = zeros(maximum(labels))
+    nlabels = maximum(labels)
+
+    # Rank labels by the number of elements they occupy within the *core* region.
+    # If `core_size === nothing`, the core is the whole array.
```
simone-silvestri:
```suggestion
```

glwagner:
Did you know that [the Caspian sea is 1000m deep and the remnant of an ancient ocean](https://www.naturalhistorymag.com/features/112161/fate-of-the-caspian-sea)? Maybe you should include it in your bathymetry.

simone-silvestri:
This should be ready to merge @taimoorsohail

taimoorsohail (approved):

PR #709, opened 2025-12-23
Build docs on more GPUs

simone-silvestri:
tartarus is quite unused lately. Since we are builing new examples that run on the GPU and for the moment the docs are broken because of an out of memory issue, I thought it might be a good idea to distribute the runs over 2 GPUs.
This might be helpful since we do not always run the examples.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -34,9 +34,9 @@ Distributed.addprocs(2)
 end
 
 Distributed.pmap(1:length(to_be_literated)) do n
-    device = Distributed.myid()-1
+    device = Distributed.myid()
     @info "switching to device $(device)"
-    CUDA.device!(device) # Set the correct GPU, the used GPUs will be number 1 and 2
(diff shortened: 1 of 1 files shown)


glwagner (approved):

.buildkite/examples_build.yml
```diff
@@ -6,7 +6,7 @@ env:
   JULIA_PKG_SERVER_REGISTRY_PREFERENCE: eager
   JULIA_NUM_PRECOMPILE_TASKS: 8
   JULIA_NUM_THREADS: 8
-  CUDA_VISIBLE_DEVICES: "2" # Device for building ClimaOcean docs
+  CUDA_VISIBLE_DEVICES: "0,1,2,3" # Devices for building ClimaOcean docs (at the moment only 0, 1 and 2 are used)
```
glwagner: Don't we need 1 or 2 for benchmarking?

simone-silvestri: oh right!

PR #710, opened 2025-12-24
Add support for ERA5 Metadata

glwagner:
This PR adds provisional support for ERA5 through Metadata.

It depends on a new package, CopernicusClimateDataStore.jl which is not yet registered.

It also changes DataWrangling/Copernicus to DataWrangling/GLORYS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,8 @@ jobs:
           CDS_API_KEY: ${{ secrets.CDS_API_KEY }}
         run: |
           mkdir -p ~/.config/era5cli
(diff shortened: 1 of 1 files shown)


glwagner:
@jtbuch ths PR intends to develop Metadata support for ERA5!

It's an extension to another package: https://github.com/NumericalEarth/CopernicusClimateDataStore.jl

which wraps around `era5cli` (and I suppose, possibly other cli's from CopernicusClimateDataStore).

I just registered CopernicusClimateDataStore but it will take 3 days to go through.

glwagner:
Possibly https://github.com/NumericalEarth/CopernicusClimateDataStore.jl is redundant

PR #711, opened 2025-12-24
Restrict Oceananigans to v0.102

navidcy:
In anticipation of breaking change in 0.103; see https://github.com/CliMA/Oceananigans.jl/pull/5068#issuecomment-3690476775


glwagner (approved):
do we want to update for 0.103?

PR #712, opened 2025-12-24
Update for Oceananigans v0.103

glwagner:
This PR updates ClimaOcean.ocean_simulation so that it doesn't add closure tracers, as expected by Oceanaingans v0.103.

@navidcy let me know if I did the compat right!

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ Note that though ClimaOcean is currently focused on hydrostatic modeling with `O
 If you use ClimaOcean for your research, teaching, or fun 🤩, everyone in our community will be grateful
 if you give credit by citing the corresponding Zenodo record, e.g.,
 
-> Wagner, G. L. et al. (2025). CliMA/ClimaOcean.jl: v0.8.7 (v0.8.7). Zenodo. https://doi.org/10.5281/zenodo.7677442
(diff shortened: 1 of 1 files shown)


Project.toml
```diff
@@ -60,7 +60,7 @@ KernelAbstractions = "0.9"
 MPI = "0.20"
 MeshArrays = "0.3, 0.4"
 NCDatasets = "0.12, 0.13, 0.14"
-Oceananigans = "0.102, 0.103, 0.104"
+Oceananigans = "0.103"
```
navidcy: Great!

navidcy:
I wanted to see that the docs build OK first! That's why I pushed something to trigger them.

navidcy:
@giordano do you have any clue why often we'd see these memory errors on the ClimaOcean docs build?

https://buildkite.com/clima/climaocean-examples/builds/5129#019b59e9-55fc-4922-8299-d81eb4cee2ca/L2256

Seems like sometimes the docs get out of memory error. But each example does fit in memory... Does the memory not clear up when an example finishes? Or could it be just as innocent thing as someone just randomly somebody using the Tartarus GPU that the docs are built on?

PR #723, opened 2026-01-12
Use correct default for atmosphere - sea ice `SimilarityTheoryFluxes`

simone-silvestri:


simone-silvestri:
Tests don't pass. However this does not really change anything in the failing tests, I think the tests will be fixed by #725 and #724. tests related to this PR are in the buildkite CI. So yeah, let's wait for those to pass, then you can merge, I will continue working on the PRs that fix the tests

PR #727, opened 2026-01-14
Test difference between distributed and serial outputs for a prescribed atmosphere `OceanSeaIceModel`

taimoorsohail:
This PR adds a test for the prescribed-atmosphere-forced OSIM to ensure the serial and distributed runs have exactly the same output.


taimoorsohail:
Tests are failing with 
```
ERROR: The following 1 direct dependency failed to precompile:
ReactantCUDAExt --code-coverage=none --color=yes --check-bounds=yes --warn-overwrite=yes --depwarn=yes --inline=yes --startup-file=no --track-allocation=none
```

simone-silvestri (approved):
Good addition if it passes!

PR #730, opened 2026-01-18
Update syntax for Oceananigans v0.104

navidcy:


navidcy:
This requires Oceananigans v0.104.1 and I'm leaning on waiting for the ClimaSeaIce release that is compatible with v0.104 before proceeding.

simone-silvestri:
I have added a PR with stategy (2), but probably stategy (1) might be best, because there is a difference between a just a numerical constant and a constant across the whole grid?

PR #736, opened 2026-01-22
Implement checkpointer for the `OceanSeaIceModel`

simone-silvestri:
This PR implements the checkpointer for an `OceanSeaIceModel` following the new syntax from Oceananigans 0.104,
since I need to add new tests, I also thought to refactor the test pipeline to use a matrix syntax like oceananigans so we do not have to duplicate the code every time.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -35,14 +35,26 @@ steps:
 
   - wait
 
-  #####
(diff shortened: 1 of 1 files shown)


navidcy:
(or is this a different PR?)

navidcy (approved):

PR #748, opened 2026-02-09
Fix bug in gustiness velocity

### LINKED ISSUE #747: Unexpected behaviour of `interface_fluxes` in non-neutral conditions
### Discussed in https://github.com/CliMA/ClimaOcean.jl/discussions/741

<div type='discussions-op-text'>

<sup>Originally posted by **mmr0** February  3, 2026</sup>
Hi,

I have been reading the awesome docs on [interface fluxes](https://clima.github.io/ClimaOceanDocumentation/stable/interface_fluxes/), but am perplexed by the bottom panel of this figure figure: 
Screenshot 2026-02-03 at 2 10 31 pm

As far as I understand the example, the atmosphere specific humidity is set to be equal to the surface saturation humidity, so there should be no vapor fluxes. So I would expect the sign of the air-sea temperature difference ΔT to set the sign of the buoyancy flux. Also as per my understanding and also the text:

> The relationship between the relative air-sea state and turbulent fluxes is modified by the presence of buoyancy fluxes -- "destabilizing" fluxes, which stimulate convection, tend to increase turbulent exchange, while stabilizing fluxes suppress turbulence and turbulent exchange. 

So I'm confused as to why a large positive ΔT of 20K corresponding to a *stabilising* buoyancy flux would lead to an increased drag coefficient?

Thanks in advance for any comments/clarifications!</div>

simone-silvestri:
these was an insidious bug in the gustiness velocity which increased with stabilizing buoyancy fluxes.
This PR fixes it. 
See discussion https://github.com/CliMA/ClimaOcean.jl/discussions/741 for reference
However, for stability reasons, we should never allow a zero velocity difference between the ocean and the atmosphere, so, in accordance with the COARE algorithm, this PR also introduces a minimum bulk velocity which limits the ocean-atmosphere velocity difference.

Thanks @mmr0 for catching this bug
closes #747

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,42 @@ jobs:
           files: lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
 
+  ecco_atmosphere:
+    name: ECCO Atmosphere - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1.12'
+        os:
+          - ubuntu-latest
+        arch:
+          - x64
+        include:
+          - os: macOS-latest
+            arch: aarch64
+            version: '1.12'
+    steps:
+      - uses: actions/checkout@v6
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
+      - uses: julia-actions/cache@v2
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: julia-actions/julia-runtest@v1
+        env:
+          TEST_GROUP: "ecco_atmosphere"
+          ECCO_USERNAME: ${{ secrets.ECCO_USERNAME }} # To download ECCO data from the podaac website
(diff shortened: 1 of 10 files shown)


simone-silvestri:
The new plots look like this
image

glwagner:
I missed it at first, so adding this for documentation: the fix to the gustiness is to use

```julia
    Jᵇ_unstable = max(zero(Jᵇ), Jᵇ)
    Uᴳ = β * cbrt(Jᵇ_unstable * h_bℓ)
```

in other words, the gustiness should only be computed for unstable boundary layers. Before this PR, we could find a "non-zero gustiness" for negative J^b, effectively increasing fluxes at low wind speeds (the bug that @mmr0 found)

As for the second change --- how is that related to this PR? I understand it may be needed in general, but it doesn't seem related.

simone-silvestri:
Ok, let's keep this brewing a bit. The problem is the roughness length which divides by ustar so a 0 velocity difference will not work. What if we keep the minimum velocity low (O(1e-3)). I think that would not impact the results whatsover (also 0.1 is probably already quite low given the reduction in Cd with velocity) but for sure it will guard us from unwanted NaNs.

simone-silvestri:
Shall we have a lower value? This is what is used in COARE, but their algorithm is probably more for diagnosing fluxes rather than time-stepping.

glwagner:
pretty interesting. in that case yeah, i'd say 1e-2 is a good idea --- even a little high, but probably those moments of extreme calm are pretty ephemeral.

PR #778, opened 2026-04-07
Support ITime and stop time in coupled ocean and sea ice models

daverumph:
Add support for passing in a Dates.DateTime-based clock, as well as model stop time, for compatibility with coupling to other models.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaOcean"
 uuid = "0376089a-ecfe-4b0e-a64f-9c555d74d754"
 license = "MIT"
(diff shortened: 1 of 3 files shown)


juliasloan25 (commented):
Thanks Dave! Could you also please bump the version to 0.9.5 in the [Project.toml here](https://github.com/CliMA/ClimaOcean.jl/blob/6e05d99ad49fb5d0e9e3859c7785cdecc4edf664/Project.toml#L4), so that we can tag a patch release and use this in ClimaCoupler?

juliasloan25 (approved):

PR #784, opened 2026-04-24
re-enable MPI tests

juliasloan25:
MPI library issues are fixed on central now, so it's been added back into climacommon, and we can remove the soft fail from MPI runs in buildkite.

Reverts soft-fail change from #770


juliasloan25:
JRA55 tests are failing because they try to download some data that isn't currently available. I think it's still a good idea to merge this PR so we can resume testing with MPI

nefrathenrici (approved):
Thanks!

